PR #2, opened 2015-04-21
Apply the source to contribution source as well

guanhuan:


colemanw:
The reason this didn't make it in before is that I'm just not sure about this change.
Maybe I could go along with it if you can explain the rationale and use-case.
My hesitation is that 'new_contact_source' is a field on the _contact_ object, and so it doesn't completely make sense to apply it to contributions, participants, etc. I can't think of another example where we mix-up entities and fields in that way.
Also, since there can be multiple contacts on the form, it's not intuitive to me that the contribution record should get it's "source" populated from contact 1. If I had to choose any contact of course 1 makes sense, but this doesn't seem like it would be intuitive to users or administrators.

guanhuan:
Hi @colemanw ,

Yes, make sense. I will probably spend some time to move this field to the form settings beside the source label to allow exposing. Do you think that will be ideal?

Best,
Guanhuan

colemanw:
Yes, if we want to expose contribution source field, we should do so in the contribution object.

guanhuan:
Are you suggesting to create a separate source field in contribution field? As my initial thoughts is to add a checkbox beside the "Source Label" field in "Additional Options" tab and remove the source field from Contact tab since webform civicrm seems using "Source Label" as both the source for contact creation and contribution. The only drawback will be that people who are using this field will get affected if they update the module.

However, I just realise that the source field in contact tab might not work respectively for more than one contact. So probably a good idea to change it before too many people start using it?

colemanw:
I believe the source fields are currently working properly. Reading through the code, this is the current behavior:
- You can globally set the source label for all contacts/participants/contributions/etc. in "Additional Options"
- You can override this default per contact by enabling the "Source" field for one or more contacts.
- Doing so will also override the source label for participants and memberships created for that contact.

Hmm, based on the 3rd point, maybe it does make sense to do so for contributions as well?

What was your proposal with the checkbox? What would it do?

guanhuan:
No, you are right. The source field works for more than one contacts respectively.

I was just thinking to expose the source label field as a global source field using a checkbox but as the contact source fields work as expected then it probably makes sense to add a contribution source field separately.

colemanw:
Agreed.

PR #6, opened 2015-05-18
Event new changes

guanhuan:
Following new features:
Add option for loading events from URL;
Add option for disabling unregister user from unselected events;

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -686,11 +686,6 @@ class wf_crm_admin_form {
       '#type' => 'checkbox',
       '#title' => t('Allow events to be autoloaded from URL'),
       '#default_value' => (bool) wf_crm_aval($this->data, 'reg_options:allow_url_load'),
-      '#description' => t('Example for "Register all":') .
-      '<br /><code>' . url("node/{$this->node->nid}", array('absolute' => TRUE, 'query' => array('event1' => ''))) . '{event1.event_id},{event2.event_id}</code></p>'.
-      t('Example for "Register separately":') .
-      '<br /><code>' . url("node/{$this->node->nid}", array('absolute' => TRUE, 'query' => array('c1event1' => ''))) . 
-      '{event1.event_id},{event2.event_id}&amp;c2event1={event3.event_id}</code></p>',
     );
     $this->help($this->form['participant']['reg_options']['block_form'], 'reg_options_block_form');
     $this->help($this->form['participant']['reg_options']['disable_unregister'], 'reg_options_disable_unregister');
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -686,8 +686,11 @@ class wf_crm_admin_form {
       '#type' => 'checkbox',
       '#title' => t('Allow events to be autoloaded from URL'),
       '#default_value' => (bool) wf_crm_aval($this->data, 'reg_options:allow_url_load'),
-      '#description' => t('Example: send the following link from CiviMail:') .
-      '<br /><code>' . url("node/{$this->node->nid}", array('absolute' => TRUE, 'query' => array('eid1' => ''))) . '{event1.event_id},{event2.event_id}</code></p>',
+      '#description' => t('Example for "Register all":') .
+      '<br /><code>' . url("node/{$this->node->nid}", array('absolute' => TRUE, 'query' => array('eid1' => ''))) . '{event1.event_id},{event2.event_id}</code></p>'.
+      t('Example for "Register separately":') .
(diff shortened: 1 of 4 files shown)


colemanw:
This looks great thanks for submitting it.

One nitpick that I don't feel _too_ strongly about is that I'm not in love with notations like "eid" because it's not terribly obvious what "e" stands for and if you have two entities that both start with the letter E then you've got real problems. "event" is only 2 characters longer than "eid"... is that so bad?

FYI the url notation for autofilling activities is "activity1" etc.

includes/wf_crm_webform_preprocess.inc
```diff
@@ -322,6 +326,38 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
...
+          $eids[$id[0]] = $eid;
+        }
+        if ($this->data['participant_reg_type'] == 'all') {
+          $urlParam = "eid$e";
+        }
+        else {
+          $urlParam = "c$c"."eid$e";
+        }
+        foreach (explode(',', wf_crm_aval($_GET, $urlParam)) as $key => $value) {
+          if (isset($eids[$value])){
```
colemanw: should check if `is_numeric`

includes/wf_crm_webform_preprocess.inc
```diff
@@ -322,6 +326,38 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
...
+            $event_ids[] = $eids[$value];
```
colemanw: cast to an (int) for safety. This is a raw url param.

includes/wf_crm_webform_preprocess.inc
```diff
@@ -322,6 +326,38 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
   }
 
   /**
+   * Load event data for the url
+   * @param int $c
+   */
+  private function loadURLEvents($c) {
+    $n = $this->data['participant_reg_type'] == 'separate' ? $c : 1;
+    if ($p = wf_crm_aval($this->data, "participant:$n:participant")) {
```
colemanw: assignment in conditional will get flagged by some code sniffers

guanhuan:
Hi Coleman,

Surely it is fine. I have changed the param name according to you suggestion. :)

colemanw:
One more question... what's the reason for the "allow_url_load" option? I can't see how it presents a security risk or usability issue to have this feature always enabled.

guanhuan:
A few examples (correct me if I am wrong ;) ):
1. If I want to set up a form which transfers event ticket from contact 1 to contact 2, meaning that unregister contact 1 from event and register contact 2 to the event without fees involved;
2. If I want to conditionally give contacts tickets for selected events for free.

colemanw:
Wait, putting event id in the url bypasses fees? Why?

guanhuan:
No, it does not. Sorry if I confused you.

If my form is set up this way:
1. without participant fee enabled;
2. with "Event(s)" field hidden with default values loaded;
As per scenario I mentioned above then I would not want the users to load any event they want from URL.

E.g. Ticket transfer case, in order to achieve ticket transfer: 
- Event(s) field for contact 1 will be hidden with "CiviCon" item unchecked by default;
- Event(s) field for contact 2 will be hidden with "CiviCon" item checked by default;

But if the 'Event auto load from URL' cannot be disabled, then I will be able to check "CiviCon" for both contact 1 and contact 2 from URL. Voila, Ticket duplicated for free.

includes/wf_crm_admin_form.inc
```diff
@@ -677,7 +677,24 @@ class wf_crm_admin_form {
...
+    );
+    $this->form['participant']['reg_options']['allow_url_load'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Allow events to be autoloaded from URL'),
+      '#default_value' => (bool) wf_crm_aval($this->data, 'reg_options:allow_url_load'),
+      '#description' => t('Example for "Register all":') .
+      '<br /><code>' . url("node/{$this->node->nid}", array('absolute' => TRUE, 'query' => array('event1' => ''))) . '{event1.event_id},{event2.event_id}</code></p>'.
+      t('Example for "Register separately":') .
+      '<br /><code>' . url("node/{$this->node->nid}", array('absolute' => TRUE, 'query' => array('c1event1' => ''))) . 
+      '{event1.event_id},{event2.event_id}&amp;c2event1={event3.event_id}</code></p>',
```
colemanw: I think this help text ought to be in the help popup instead, to de-clutter the form.

colemanw:
Ok sounds reasonable. See my one last comment on style and then we can merge.

guanhuan:
Thanks. The help text is now moved.

PR #7, opened 2015-05-28
added link to download civicrm files using webform

sunilpawar:


includes/wf_crm_webform_base.inc
```diff
@@ -527,10 +528,22 @@ abstract class wf_crm_webform_base {
...
+    $entity_id = '';
+    if ($entity && strpos($fieldName, 'custom_') === 0 ) {
+      switch ( $entity ) {
+      case 'contact':
+        $entity_id = $_GET['cid1'];
+        break;
+      case 'activity':
+        $entity_id = $_GET['activity1'];
+        break;
+      }
```
colemanw: $_GET is not a reliable source for ids and I don't think you can count on the number always being 1. Try replacing this switch statement with something like

```
$entity_id = $this->ent[$entity][$n]['id']
```

Although you'll need to somehow get the value for `$n`

colemanw:
Hey @sunilpawar any thoughts on that feedback?

sunilpawar:
@colemanw can you take look ?

includes/wf_crm_webform_base.inc
```diff
@@ -527,10 +529,15 @@ abstract class wf_crm_webform_base {
       );
     }
     $file = wf_crm_apivalues('file', 'get', $val);
+    $entity_id = '';
+    if ($entity && $n && ( strpos($fieldName, 'custom_') === 0 || strpos($fieldName, 'file_') === 0)) {
+      $entity_id = $this->ent[$entity][$n]['id'];
```
colemanw, line 534: This mostly looks right except that `$this->ent[$entity][$n]['id']` might not be set, in the case of e.g. entering a new contact.

sunilpawar, line 534: do we really need this functionality for entering a new contact? this is only for showing download link.

colemanw, line 534: Probably not, but my point is that your code will generate a php undefined index warning if the variable is not set.

colemanw:
Did some more testing, additional fixes at 06c5225.

PR #9, opened 2015-06-25
Tax and invoice integration

guanhuan:
Added complete tax and invoice integration according to current CiviCRM tax integration progress.

There are a few minor issues which need to be fixed in CiviCRM core before being fixed in Webform CiviCRM to keep consistency.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -47,15 +47,13 @@ abstract class wf_crm_webform_base {
         
       case 'tax_rate':
         $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
-        if (isset($taxSettings['invoicing'])) {
-          if ($taxSettings['invoicing']) {
+        if (is_array($taxSettings) && !empty($taxSettings['invoicing'])) {
             if ($this->contribution_page) {
               // tax integration
               $taxRates = CRM_Core_PseudoConstant::getTaxRates();
               $this->_tax_rate = $taxRates[$this->_contribution_page['financial_type_id']];
             }
             return $this->_tax_rate;
-          }
         }
         return;
       
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -30,10 +30,6 @@ abstract class wf_crm_webform_base {
    * @return mixed
    */
   function __get($name) {
-    $contribution_page_id = wf_crm_aval($this->data, 'contribution:1:contribution:1:contribution_page_id');
-    if ($contribution_page_id && !$this->_contribution_page) {
-      $this->_contribution_page = wf_civicrm_api('contribution_page', 'getsingle', array('id' => $contribution_page_id));
-    }
     switch ($name) {
       case 'payment_processor':
         $payment_processor_id = wf_crm_aval($this->data, 'contribution:1:contribution:1:payment_processor_id');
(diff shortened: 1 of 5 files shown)


includes/wf_crm_webform_base.inc
```diff
@@ -39,11 +43,21 @@ abstract class wf_crm_webform_base {
...
-        if ($contribution_page_id && !$this->_contribution_page) {
-          $this->_contribution_page = wf_civicrm_api('contribution_page', 'getsingle', array('id' => $contribution_page_id));
-        }
         return $this->_contribution_page;
+        
+      case 'tax_rate':
+        $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
+        if(isset($taxSettings['invoicing'])){
+          if($taxSettings['invoicing']){
+            if ($this->_contribution_page) {
```
colemanw: This seems racy. Trying to access `$this->tax_rate` before `$this->contribution_page` would result in NULL?

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1408,7 +1408,14 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     // Calculate totals
     $this->totalContribution = 0;
     foreach ($this->line_items as &$item) {
-      $this->totalContribution += $item['line_total'] = $item['unit_price'] * (int) $item['qty'];
+      // tax integration
+      if (!is_null($this->tax_rate)) {
+        $item['line_total'] = $item['unit_price'] * (int) $item['qty'];
+        $item['tax_amount'] = ($this->tax_rate / 100) * $item['line_total'];
+        $this->totalContribution += (1 + $this->tax_rate / 100) * $item['unit_price'] * (int) $item['qty'];
+      }else{
```
colemanw: see drupal coding standards re spacing and linebreaks here

includes/wf_crm_webform_base.inc
colemanw: Probably the solution is to change this line to `if ($this->contribution_page) {` (omitting the underscore to allow the magic method to perform its magic)

colemanw: Oh I see, your workaround was to move the code that fetched contribution page out of its condition.
Assuming it doesn't cause any kind of weird recursion with a magic method being called from within another magic method, I think my suggestion might be cleaner.

guanhuan:
Thanks for the feedback. I have applied the changes. Please let me know if there is any problem.

includes/wf_crm_webform_base.inc
```diff
@@ -44,6 +44,20 @@ abstract class wf_crm_webform_base {
           $this->_contribution_page = wf_civicrm_api('contribution_page', 'getsingle', array('id' => $contribution_page_id));
         }
         return $this->_contribution_page;
+        
+      case 'tax_rate':
+        $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
+        if(isset($taxSettings['invoicing'])){
+          if($taxSettings['invoicing']){
```
colemanw: Fyi drupal coding standards say there ought to be a space between the `if` and the opening `(`.
Since webform_civicrm v4 supports CiviCRM versions 4.4 and up, this setting may not exist, therefore the variable `$taxSettings` may be NULL or an empty string, so it might not be safe to treat it as an array (php 5.3 is especially quirky about treating non-arrays as arrays).
I would suggest the above two if statements be combined into a single if statement as follows:

```
if (is_array($taxSettings) && !empty($taxSettings['invoicing'])) {
```

js/webform_civicrm_payment.js
```diff
@@ -71,8 +71,20 @@ cj(function($) {
...
-      $('td+td', $lineItem).html(CRM.formatMoney(amount));
-      $lineItem.data('amount', amount);
+      var hasVAT = 0;
+      var taxPara = 1;
+      var classes = $lineItem.attr('class').split(' ');
+      for (var i = 0; i < classes.length; i++) {
+    	var match = /^vat-/.exec(classes[i]);
+      if(match !== null){
+    		hasVAT = classes[i].substr(classes[i].indexOf("-") + 1);
+    	}
```
colemanw: This seems overly complicated. Any reason you must use css classes and regex instead of just attaching the data directly? Similar to `data-amount` you could add `data-vat` above.

includes/wf_crm_webform_base.inc
```diff
@@ -44,6 +44,20 @@ abstract class wf_crm_webform_base {
           $this->_contribution_page = wf_civicrm_api('contribution_page', 'getsingle', array('id' => $contribution_page_id));
         }
         return $this->_contribution_page;
+        
+      case 'tax_rate':
+        $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
+        if (isset($taxSettings['invoicing'])) {
+          if ($taxSettings['invoicing']) {
```
colemanw: Since webform_civicrm v4 supports CiviCRM versions 4.4 and up, this setting may not exist, therefore the variable `$taxSettings` may be NULL or an empty string, so it might not be safe to treat it as an array (php 5.3 is especially quirky about treating non-arrays as arrays).
I would suggest the above two if statements be combined into a single if statement as follows:

```
if (is_array($taxSettings) && !empty($taxSettings['invoicing'])) {
```

guanhuan:
Hi Coleman,

Thanks for the further comments. Those are now all applied.

Best

Guanhuan

PR #10, opened 2015-07-02
#2516924 - Create Membership Renewal Activity

colemanw:


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1102,6 +1102,22 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $params['skipStatusCal'] = !empty($params['status_id']);
 
       $result = wf_civicrm_api('membership', 'create', $params);
+      
+      //Issue #2516924 If existing membership create renewal activity
+      foreach ($result['values'] as $memId => $membership) {
+        if ($params['membership_type_id'] == $memId) {
```
colemanw, line 1108: It looks like you're comparing apples to oranges here. Why would it matter if the membership type matched the membership id?

colemanw:
@relldoesphp can you take a look?

relldoesphp:
Sorry I meant to check the id not the membership type id, it was to check if the membership id created matched the membership id passed in

colemanw:
Ok. Actually I don't think we really need to `foreach` the results of the api call since there can only possibly be 1 record (or 0 if something went wrong). Just need to check `if (!empty($result['id']) && !empty($params['id']))`

colemanw:
Ok fixed with 3222a31

PR #11, opened 2015-07-27
Add Contribution Source field

guanhuan:
As agreed, add a source field for contributions.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -988,6 +988,10 @@ function wf_crm_get_fields($var = 'fields') {
         'value' => 0,
         'weight' => 9997,
       );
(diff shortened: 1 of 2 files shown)


includes/utils.inc
```diff
@@ -988,6 +988,11 @@ function wf_crm_get_fields($var = 'fields') {
         'value' => 0,
         'weight' => 9997,
       );
+      $fields['contribution_source'] = array(
+        'name' => t('Contribution Source'),
+        'type' => 'textfield',
+        'weight' => 9998,
```
colemanw: don't worry about weight, that's not important for this field. You can just remove it.

PR #13, opened 2015-08-17
PG-99 issue# 2. Handling of pay later option while renewing membership

nishant-bhorodia:
This PR handles the below mentioned issue:

**Issue:**
When a person who holds an active membership come to renew his membership via webform civicrm, he chose pay later on the payment method. This changes his current membership to be pending. Fix according to CiviCRM core behaviour.


colemanw:
Fix looks good, thanks.

nishant-bhorodia:
Thanks @colemanw

PR #16, opened 2015-09-09
Add token to the IPN redirect URL in case the user is not authenticated

robinmitra:
The user would see access denied message otherwise.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1705,10 +1705,9 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $url = url('node/' . $this->node->nid, array('absolute' => TRUE));
     }
     elseif ($url == '<confirmation>') {
-      global $user;
       $query = array('sid' => $this->submission->sid);
       // Add token if user is not authenticated, inline with 'webform_client_form_submit()'
-      if ((int) $user->uid === 0) {
+      if (user_is_anonymous()) {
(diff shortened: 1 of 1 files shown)


guanhuan:
Hi @colemanw ,

Currently when people complete an offsite payment and gets back to the confirmation page, they will see an "access denied" message. We have noticed that for normal onsite submission of the form, an extra token is added in the URL. This PR is to add that back to offsite payment.

Best

Guanhuan

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1705,7 +1705,13 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $url = url('node/' . $this->node->nid, array('absolute' => TRUE));
     }
     elseif ($url == '<confirmation>') {
-      $url = url("node/{$this->node->nid}/done", array('query' => array('sid' => $this->submission->sid), 'absolute' => TRUE));
+      global $user;
+      $query = array('sid' => $this->submission->sid);
+      // Add token if user is not authenticated, inline with 'webform_client_form_submit()'
+      if ((int) $user->uid === 0) {
```
colemanw: This line could be simplified as `if (user_is_anonymous())` and the line above (`global $user;`) could be removed.

colemanw:
Thanks @guanhuan for the fix.

colemanw:
@robinmitra see one coding suggestion above.

robinmitra:
Thanks for your comment @colemanw - have updated the PR accordingly.

PR #18, opened 2015-10-06
Fix sending receipt proceed after payment processing 

chanunc:
Fix sending receipt proceed after payment processing and should not send out if payment failed.


colemanw:
Thanks for the PR. Two questions:
- Where is the check to ensure that there even is a contribution being processed before attempting to send a receipt for it?
- Can we move the receipt block of code into its own function to reduce cyclomatic complexity?

chanunc:
refer to new PR.
https://github.com/colemanw/webform_civicrm/pull/20

PR #19, opened 2015-10-12
Fix numeric name breaks custom fields creation

guanhuan:
If a numeric name or a name starts with numbers is entered into autocomplete contact field in creation mode, when custom fields are updated/filled, a constraint violation is produced.
error on submit


colemanw:
Can you give steps to reproduce this? I don't quite follow the description.

colemanw:
Nevermind I was able to reproduce it.

colemanw:
Thanks for the PR. I decided to fix it at a deeper level. See b806136badf7587e2afb731d64270f4ae1e08f15

PR #20, opened 2015-10-13
Fix sending receipt proceed after payment processing

chanunc:
Fix sending receipt proceed after payment processing and should not send out if payment failed.
Edit following suggestion in 
https://github.com/colemanw/webform_civicrm/pull/18

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -240,6 +240,26 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     if (empty($this->submission->is_draft) && !empty($this->ent['contribution'][1]['id']) && $this->contributionIsIncomplete && !$this->contributionIsPayLater) {
       $this->submitIPNPayment();
     }
+      // Send receipt 
+    if (empty($this->submission->is_draft) 
+        && !empty($this->ent['contribution'][1]['id']) 
+        && !empty($this->contribution_page['is_email_receipt']) 
+        && (!$this->contributionIsIncomplete || $this->contributionIsPayLater) ) {
+      $this->sendReceipt();
+    }
+  }
+
+  /**
+   * Send receipt 
+   */
+  private function sendReceipt(){
+    // tax integration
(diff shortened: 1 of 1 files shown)


guanhuan:
Thanks for updating this @chanunc .

guanhuan:
Hi @colemanw,

Could you take a look and let us know if there is anything else needs to be changed in this PR? Thanks.

Best

Guanhuan

includes/wf_crm_webform_postprocess.inc
```diff
@@ -240,6 +240,26 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+   * Send receipt 
+   */
+  private function sendReceipt(){
+    // tax integration
+    if (!is_null($this->tax_rate)) {
+      $template = CRM_Core_Smarty::singleton();
+      $template->assign('dataArray', array( "{$this->tax_rate}" => $this->tax_rate/100 ));
+    }
+    $contribute_id = $this->ent['contribution'][1]['id'];
+    wf_civicrm_api('contribution', 'sendconfirmation', array('id' => $contribute_id) + $this->contribution_page);
```
colemanw, line 262: Ok reading the code, this will send a reciept only if:
- it is not an ipn payment
- the contribution record was created
- the contribution page is set to do so
- the contribution is complete OR it is pay later

I'm unsure about that last condition - does sending a receipt for a pay-later transaction make sense? Maybe it does, just wanted to double-check with you and make sure it's what you meant.

guanhuan, line 262: Sending an invoice email(same template with different conditions) with PDF invoice is generally needed for organisations who uses invoices. However, as I mentioned in the email, it should send out the is_pay_later version rather than receipt version. Hope this does not confuse you.

colemanw, line 262: Ok so assuming that is a separate issue, is this ready to merge?

guanhuan:
Yep, this now sends email to users except the IPN comes back say incomplete. So good to merge.

PR #22, opened 2015-10-19
Fix pay later is not properly set by the confirmation api

guanhuan:
https://www.drupal.org/node/2593357
The fix is to set "is_pay_later" to True for confirmation email if the contribution is pending after all. 

The sendconfirmation api call does not always set this variable. Also "is_pay_later" variable in message template is getting a bit confusing after the invoice is launched. This should be potentially set to be "Pending" in order to make it more accurate. Probably something to be fixed in the core later.


colemanw:
Manually merged in c4712d3

PR #23, opened 2015-10-21
Paypal express button fixes

colemanw:


colemanw:
Reading the code again, this seems superfluous. The PP Express button redirects correctly either way.

PR #24, opened 2015-11-06
Phone, Email issue

nishant-bhorodia:
Hello @colemanw 

This patch sorts the following issue:

Phone number and Email fields have associated "location_type_id". Webform is always picking up the ones that are set to primary in civi (primary phone, primary email). It ignores the location type chosen inside civicrm tab of the webform (home, work, etc.).

So for example, if admin choose "Home Phone"  field on webform_civicrm and in civi "work Phone" is marked as primary for a contact then the webform phone field will pick up the "work phone" even when the location type chosen is "Home Phone"

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -198,22 +198,53 @@ abstract class wf_crm_webform_base {
    * @param string $ent
    * @param array $values
    * @return array $reorderedArray
-   * @todo Handle user-select case
    */
   protected function reorderByLocationType($c, $ent, $values = array()){
     $reorderedArray = array();
-    // For every entity setting per contact
-    foreach ($this->settings['data']['contact'][$c][$ent] as $setting) {
+
+    if(isset($this->settings['data']['contact'][$c][$ent])){
+      // First pass
+      $reorderedArray = $this->matchLocationTypes($c, $ent, $values);
+
+      // Second pass
+      $reorderedArray = $this->handleRemainingValues($reorderedArray, $values);
+
+      return $reorderedArray;
+    } else {
+      return $values;
+    }
+  }
+
+  /**
+   * Organize values according to location types
+   *
+   * @param integer $c
+   * @param string $ent
+   * @param array $values
+   * @return array $reorderedArray
+   */
+  protected function matchLocationTypes($c, $ent, &$values){
+    // create temporary settings array to include 'user-select' fields
+    // on the right place in array
+    $settingsArray = $this->add_user_select_field_placeholder($ent, $this->settings['data']['contact'][$c]);
+
+    // Go through the array and match up locations by type
+    // Put placeholder 'user-select' where location_type_id == -1 for second pass
+    foreach ($settingsArray[$ent] as $setting) {
       $valueFound = false;
       foreach($values as $key => $value){
         if((in_array($ent, array('address', 'email')) && $value['location_type_id'] == $setting['location_type_id'])
            ||
           ($value['location_type_id'] == $setting['location_type_id'] && $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])     
           ){
-          $reorderedArray[$key] = $value;
-          $valueFound = true;
-          unset($values[$key]);
-          break;
+            $reorderedArray[$key] = $value;
+            $valueFound = true;
+            unset($values[$key]);
+            break;
+        } else if($setting['location_type_id'] == -1) { // for 'user-select' fields
+            $valueFound = true;
+            $reorderedArray[] = 'user-select';
+            break;
         }
       }
 
@@ -224,6 +255,42 @@ abstract class wf_crm_webform_base {
       }
     }
     return $reorderedArray;
+  }
+
+  /**
+   * Put remaining values in 'user-select' fields
+   *
+   * @param array $reorderedArray
+   * @param array $values
+   * @return array $reorderedArray
+   */
+  protected function handleRemainingValues($reorderedArray, &$values){
+    // Put leftover values in fields marked as 'user-select'
+    foreach($reorderedArray as $key => $value){
+      if($reorderedArray[$key] == 'user-select'){
+        $reorderedArray[$key] = !empty($values) ? array_shift($values) : '';
+      }
+    }
+    return $reorderedArray;
+  }
+
+  /**
+   * Add location_type_id = -1 for user-select fields for identification later
+   *
+   * @param string $ent
+   * @param array $settings
+   * @return array $settings
+   */
+  protected function add_user_select_field_placeholder($ent, $settings = array()){
+    if($settings['number_of_'.$ent] > count($settings[$ent])){
+      for($i = 1; $i <= $settings['number_of_'.$ent]; $i++){
+        if(!array_key_exists($i, $settings[$ent])){
+          $settings[$ent][$i]['location_type_id'] = -1;
+        }
+      }
+      ksort($settings[$ent]);
+    }
+    return $settings;
   }
 
   /**
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -85,6 +85,10 @@ abstract class wf_crm_webform_base {
           $params['options']['sort'] = 'is_primary DESC';
         }
         $result = wf_civicrm_api($ent, 'get', $params);
+        // Handle location field sorting
+        if(in_array($ent, wf_crm_location_fields())){
+          $result['values'] = $this->reorderByLocationType($ent, $result['values']);
+        }
         if (!empty($result['values'])) {
           // Index array from 1 instead of 0
           $result = array_merge(array(0), array_values($result['values']));
@@ -183,6 +187,36 @@ abstract class wf_crm_webform_base {
       }
     }
     return $info;
+  }
+
+  /**
+   * Reorder returned results according to settings chosen in wf_civicrm backend
+   *
+   * @param string $ent
+   * @param array $values
+   * @return array $reorderedArray
+   * @todo Handle user-select case
+   */
+  protected function reorderByLocationType($ent, $values = array()){
+    $reorderedArray = array();
+    // For every location field setting
+    foreach ($this->settings['data']['contact'] as $setting) {
+      // For every Entity
+      foreach($setting[$ent] as $e){
+        // For every result row
+        foreach($values as $key => $value){
+          if((in_array($ent, array('address', 'email')) && $value['location_type_id'] == $e['location_type_id'])
+            ||
+            ($value['location_type_id'] == $e['location_type_id'] && $value[$ent.'_type_id'] == $e[$ent.'_type_id'])     
+            ){
+              $reorderedArray[$key] = $value;
+              unset($values[$key]);
+              break;
+          }
+        }
+      }
+    }
+    return $reorderedArray;
   }
 
   /**
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -229,7 +229,7 @@ abstract class wf_crm_webform_base {
     $settingsArray = $this->add_user_select_field_placeholder($ent, $this->settings['data']['contact'][$c]);
     $userSelectIndex = 0;
     // Go through the array and match up locations by type
-    // Put placeholder 'user-select' where location_type_id == -1 for second pass
+    // Put placeholder 'user-select' where location_type_id is empty for second pass
     foreach ($settingsArray[$ent] as $setting) {
       $valueFound = false;
       foreach($values as $key => $value){
@@ -241,7 +241,7 @@ abstract class wf_crm_webform_base {
             $valueFound = true;
             unset($values[$key]);
             break;
-        } else if($setting['location_type_id'] == -1) { // for 'user-select' fields
+        } else if(empty($setting['location_type_id'])) { // for 'user-select' fields
             $valueFound = true;
             $reorderedArray['us'.$userSelectIndex] = 'user-select';
             $userSelectIndex ++;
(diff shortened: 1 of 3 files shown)


includes/wf_crm_webform_preprocess.inc
```diff
@@ -529,7 +529,21 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
           // If the user has already entered a value for this field, don't change it
           if (isset($this->info[$ent][$c][$table][$n][$name])
             && !(isset($component['cid']) && isset($submitted[$component['cid']]))) {
-            $val = $this->info[$ent][$c][$table][$n][$name];
+            // pick phone number and email according to selecte location type
+            if (($table == 'phone' && $name == 'phone')
+              || ($table == 'email' && $name == 'email')) {
+              $location_type_id = $this->node->webform_civicrm['data']['contact'][$c][$name][$n]['location_type_id'];
```
colemanw: you can't assume this variable is set - location_type_id is optional (could be "user select")

includes/wf_crm_webform_preprocess.inc
```diff
@@ -529,7 +529,21 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
           // If the user has already entered a value for this field, don't change it
           if (isset($this->info[$ent][$c][$table][$n][$name])
             && !(isset($component['cid']) && isset($submitted[$component['cid']]))) {
-            $val = $this->info[$ent][$c][$table][$n][$name];
+            // pick phone number and email according to selecte location type
+            if (($table == 'phone' && $name == 'phone')
+              || ($table == 'email' && $name == 'email')) {
```
colemanw: why is it only applying to a single field in the entity? Should cover the entire entity I think (this approach seems like it would get the right phone number but the wrong phone extension or phone type, for example)

colemanw:
Seems like a generally good step forward, although it makes a couple assumptions that are not always going to be correct. I wonder if there's a better place to perform this logic, perhaps at the layer where these records are being retrieved. (not saying for sure, just wondering).

What about addresses?

includes/wf_crm_webform_preprocess.inc
nishant-bhorodia: Yes you are right @colemanw . I will improve the code so that it applies to all fields of these entities. Thanks!!

includes/wf_crm_webform_preprocess.inc
nishant-bhorodia: I will run tests for this scenario and modify code accordingly

nishant-bhorodia:
Hello @colemanw. Thank you for the comments.
I was trying to sort the scenarios that I encountered but I will have to make the solution more generalised.

Where do you think this logic makes more sense to implement ?

colemanw:
I think the part of `wf_crm_webform_base::loadContact` which handles location fields should be split apart into a separate function. This function should
1. Load the records (as it currently does)
2. Look at the location types specified on the back-end settings
3. Reorder the retrieved records to match

nishant-bhorodia:
Hello @colemanw 
Please have a look at my latest commit. Do you think this is in the right direction ?
Thanks

includes/wf_crm_webform_base.inc
```diff
@@ -186,6 +190,36 @@ abstract class wf_crm_webform_base {
...
+   *
+   * @param string $ent
+   * @param array $values
+   * @return array $reorderedArray
+   * @todo Handle user-select case
+   */
+  protected function reorderByLocationType($ent, $values = array()){
+    $reorderedArray = array();
+    // For every location field setting
+    foreach ($this->settings['data']['contact'] as $setting) {
```
colemanw: I think looping through every contact on the webform is wrong - should just focus on the one relevant contact.

includes/wf_crm_webform_base.inc
```diff
@@ -186,6 +190,36 @@ abstract class wf_crm_webform_base {
...
+      // For every Entity
+      foreach($setting[$ent] as $e){
+        // For every result row
+        foreach($values as $key => $value){
```
colemanw: This loop seems to be redundant with the one above it - both `$values` and `$settings[$ent]` contain the same data - no need to loop through both.

includes/wf_crm_webform_base.inc
```diff
@@ -186,6 +190,36 @@ abstract class wf_crm_webform_base {
...
+          if((in_array($ent, array('address', 'email')) && $value['location_type_id'] == $e['location_type_id'])
+            ||
+            ($value['location_type_id'] == $e['location_type_id'] && $value[$ent.'_type_id'] == $e[$ent.'_type_id'])     
+            ){
+              $reorderedArray[$key] = $value;
+              unset($values[$key]);
+              break;
```
colemanw: it appears the effect of this will be to take an array with potentially multiple values and return an array with at most one value. This would cause the contact's second or third phone or address to be "lost"

includes/wf_crm_webform_base.inc
```diff
@@ -85,6 +85,10 @@ abstract class wf_crm_webform_base {
           $params['options']['sort'] = 'is_primary DESC';
         }
         $result = wf_civicrm_api($ent, 'get', $params);
+        // Handle location field sorting
+        if(in_array($ent, wf_crm_location_fields())){
+          $result['values'] = $this->reorderByLocationType($ent, $result['values']);
```
colemanw: This looks like the right change here. However (see comments below) I think the function isn't yet doing what we want it to do.

includes/wf_crm_webform_base.inc
nishant-bhorodia: Agreed. Will refactor the code and submit it back

includes/wf_crm_webform_base.inc
nishant-bhorodia: Hello @colemanw 
I am not sure what you mean by that as `$setting` contains the settings chosen by the user on webform_civicrm tab and `$values`  contain the data fetched based on those settings but in incorrect order.
Or do you mean something else ?

nishant-bhorodia:
Hello @colemanw 
Thank you for the going through the commit and suggesting changes.

This new commit has the following features:
1. Handles one contact at a time
2. Reorder the array without losing any values (Except address and email. Please see the question below)

_Question:_
I noticed that regardless of the type you select for address, email, in the fetched array it returns all the values for that contact. So for example if chose "work email" or "work address" in webform_civicrm, it still will return all the values (work, home, billing, etc.) in the $result['values']. In which case the function created keeps only the chosen value ("work email", "work address" for example) and remove others.
Is it ok or do I need to keep those values as well ?

Thanks in advance.

colemanw:
It sounds like you might be making the assumption that only one address or email will be on the webform.
Need to test with multiple.

nishant-bhorodia:
Hello @colemanw 

Thank you. I ran the tests with all the location fields both single and multiple on the same form and found out that no data is lost and fields are filled with the correct data.
I also tried changing the type of fields in webform_civicrm backend and the frontend fields change accordingly.

While doing that I found one other issue. Below is the scenario:

if you choose 3 fields ( for example: home phone, main phone, other phone in webform civicrm) and "main phone" is not set in civi then only two fields are fetched (in this case, home phone and other phone fields are fetched). In this scenario "main phone" webform field gets populated by "other phone" value.

This commit fixes that as well.

Please let me know if you'd like me to run tests for any specific scenarios if I am missing any.

Many Thanks.

Regards,
Nishant Bhorodia

colemanw:
Great, thanks for the additional testing and tweaking. I'll check it out soon.

nishant-bhorodia:
Thanks @colemanw  !!

nishant-bhorodia:
Hello @colemanw 

Hope you are well.
Sorry but is there anything I can do to make the merge possible a little faster please ? As some of the urgent client issues are depending on the fix.
So if there are any tests you'd like me to do, please suggest.

Many Thanks !!

includes/wf_crm_webform_base.inc
```diff
@@ -186,6 +192,41 @@ abstract class wf_crm_webform_base {
...
+   */
+  protected function reorderByLocationType($c, $ent, $values = array()){
+    $reorderedArray = array();
+    // For every entity setting per contact
+    foreach ($this->settings['data']['contact'][$c][$ent] as $setting) {
+      $valueFound = false;
+      foreach($values as $key => $value){
+        if((in_array($ent, array('address', 'email')) && $value['location_type_id'] == $setting['location_type_id'])
+           ||
+          ($value['location_type_id'] == $setting['location_type_id'] && $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])
```
colemanw, line 238: This condition seems to be assuming that `$setting['location_type_id']` is always set, but if the location type is "user select" then this will not be the case :/

colemanw:
@nishant-bhorodia have done another code review and found one issue (see inline comment).

I think the ideal behavior would be to make 2 passes:
- First go through the array and match up locations by type
- Then go through the leftovers and dump any unmatched locations into "user select" fields (where location doesn't matter)

nishant-bhorodia:
Thanks @colemanw !!

The latest commit takes care of user-select fields.
Also did some code refactoring.

Tested for user-select fields and also tested for all previous cases.

Many Thanks!!

colemanw:
Yes, that's what I meant. I guess it's simpler to just check if it's empty and am a little worried it would cause a fatal error if we actually sent `'location_type_id' => -1` into the CiviCRM api.

nishant-bhorodia:
Thank you @colemanw , makes perfect sense :)

Changed the implementation now to check for `empty()` rather than `-1`.

Regards!!

colemanw:
Thanks @nishant-bhorodia. I want to get this tested out a bit more and it may not be instant due to the holiday season. If you are in a rush I suggest you simply patch your client's site with this change while you are waiting for it to be merged.

nishant-bhorodia:
Hello @colemanw 

Thank you for letting me know. 
I wish you a Merry Christmas and an amazing holiday season :)

Cheers !!

colemanw:
Have requested community review as well.
See https://www.drupal.org/node/1503012#comment-10682358

nishant-bhorodia:
Hello @colemanw  

A very happy new year 2016 !!
Just needed to ask if you found any issues with the code.

Many Thanks.

colemanw:
The code looks good to me but so far I haven't had any takers for testing it out.
I would really like to get another pair of eyes on it since this change is going to affect everyone who uses the module.

nishant-bhorodia:
Thanks @colemanw 

Sorry for the late reply. 
I hope someone from the community tests this fix out soon :)

Many Thanks.

nishant-bhorodia:
Hello @colemanw 

Hope you are doing well.

Just an update, Compucorp UK QAed the patch and everything is working as expected according to the QA results !!

Please can you let me know how we go forward with this?

Thanks in advance :)

colemanw:
Thanks @nishant-bhorodia - did you test with "user select"? See code comment above.

nishant-bhorodia:
Hey @colemanw. Yes, the patch has been tested for "user select" as well. Below are the test cases this patch is tested against.

**Test Cases**
All tests to be run by accessing webform directly with "cid" and also by filling up existing contact by autocomplete.

1) Test email types
1.1) Set a email type as primary in civi (ex:work email) and set another in the webform civi config (ex:home email) and check if the type set civi config gets listed in webform
1.2) Test with multiple contacts
1.3) Test with all email types
1.4) Test user select

2) Test phone types
2.1) Set a phone type as primary in civi (ex:work phone) and set another in the webform civi config (ex:home phone) and check if the type set in civi config gets listed in webform
2.2) Test with multiple contacts
2.3) Test with all phone types
2.4) Test user select

3) Test address types
3.1) Set a address type as primary in civi (ex:work addr) and set another in the webform civi config (ex:home addr) and check if the types set in civi config gets listed in webform
3.2) Test with multiple contacts
3.3) Test with all address types
3.4) Test user select

colemanw:
Thanks, that looks like a thorough list of test scenarios.

nishant-bhorodia:
Thanks @colemanw :)

colemanw:
Thank for your hard work on this.

mercurymotos:
sorry to be late the party on this one, but I'm testing this patch using 7.x-4.x and I am finding a problem with the writing of the addresses.

Webform-CiviCRM correctly displays the different address and phone types we have (Mailing, Work, Work #2) when loading the form.  However, the addresses are not saved/updated at all on submission.  The webform logs show the updated data, but its not being written to CiviCRM.  Phone numbers are being updated, but not street addresses (exception - the billing address is updated correctly by the billing module).

Further testing shows that new contacts do not have their address data written at all.  Logged-in users do not have updates pushed to CiviCRM.

I'll have to copy the site onto a local environment to trace.  It might be a while...

colemanw:
@mercurymotos you might want to try with the latest -dev release of this module since there has been one or two followup fixes since this main PR was merged.

mercurymotos:
@colemanw - thanks.

Maybe I should have been clearer - I cloned the latest 7.x.-4.x that this has been merged into (but the .info file doesn't have a version?).  But I just downloaded 7.x-4.15+21-dev and I have the same issue.

I'll start to explore this on a clean install using CiviCRM 4.6x.

mercurymotos:
@colemanw 

On my test version each address set generates the following error:

> Notice: Undefined index: address in wf_crm_webform_postprocess->reorderLocationValues() (line 2211 of /Applications/MAMP/htdocs/mmtest/sites/all/modules/webform_civicrm/includes/wf_crm_webform_postprocess.inc).

I'm having issues with the new version of PhpStorm and MAMP and I can't get the debugger to behave.  I'll try to generate a patch, but if this is an easy fix please don't wait for me...

colemanw:
I won't be able to look at this right away, so yes please do create a PR 
when you can.

nishant-bhorodia:
Hello @mercurymotos 
Thank you for testing this out. I will try to replicate this on my machine and will create a patch for it this week.
Many Thanks.

colemanw:
@nishant-bhorodia, @mercurymotos could you test out the additional fix in #43 please?

PR #25, opened 2015-11-10
Fix incorrect membership status and length when use offsite payment processor during renewal

guanhuan:


guanhuan:
Currently if an offsite payment processor is selected during renewal. After form submission but before payment completion, the membership length will get extended but the whole membership gets set to pending. After the payment is completed, the membership is considered newly activated hence it only gets one year length from today.

To calculate membership length, only the status of both membership and payment are relevant. The fix removes unnecessary conditions.

PR #30, opened 2016-01-19
Move webform_civicrm_payment.js code into Drupal behaviors.

trepmag:
Drupal behaviors is documented here . Using behaviors allow to reapply javascript processing on some element which has been reloaded after an ajax call.

Issue https://www.drupal.org/node/2652764

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -1,10 +1,10 @@
 // Webform payment processing using CiviCRM's jQuery
-cj(function($) {
+(function ($) {
   'use strict';
   var
-    setting = Drupal.settings.webform_civicrm,
-    $contributionAmount = $('[name*="[civicrm_1_contribution_1_contribution_total_amount]"]'),
-    $processorFields = $('.civicrm-enabled[name$="civicrm_1_contribution_1_contribution_payment_processor_id]"]');
+    setting,
+    $contributionAmount,
+    $processorFields;
 
   function getPaymentProcessor() {
     if (!$processorFields.length) {
@@ -38,15 +38,6 @@ cj(function($) {
       $('#billing-payment-block').html('');
     }
   }
-  $processorFields.on('change', function() {
-    setting.billingSubmission || (setting.billingSubmission = {});
-    $('input:visible, select', '#billing-payment-block').each(function() {
-      var name = $(this).attr('name');
-      name && (setting.billingSubmission[name] = $(this).val());
-    });
-    loadBillingBlock();
-  });
-  loadBillingBlock();
 
   function tally() {
     var total = 0;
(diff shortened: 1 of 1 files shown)


colemanw:
Can you please describe the bug you are trying to solve with steps to reproduce? IMO drupal behaviors are a legacy holdover from the early days of jQuery (and the drupal community at large has been slow to adopt modern javascript practices).

trepmag:
Drupal behaviors isn't about legacy holdover: https://www.lullabot.com/articles/understanding-javascript-behaviors-in-drupal... Anyway, it allow you to ajaxify things in a "unobstructive" manner.

For instance, if you ajaxify civicrm contribution webform you will find that the "Contribution Amount" total won't update on input 'change' event or any other javascript things that was applied before an ajax reload. 

You probably can reproduce this with the module webform_ajax...

colemanw:
That article does not fill me with confidence that the Drupal community is up to speed with the rest of the javascript world. The modern alternative to drupal behaviors is the much more elegant jQuery.on() method.

Anyway, what is the bug you are working to fix here?

trepmag:
Well, Drupal behaviors is neither an alternative to jQuery,on(). They are just triggered in a centralized $(document).read() callback... But, I think you've made your idea... No worry...

The fix here allow to ajaxify a civicrm webform containing a contribution without losing the "Contribution Amount" total update on input 'change' event.

colemanw:
Fair enough. I'm not opposed to this change. Since this is a drupal 7 module and this is the way to do things in drupal 7.

My one word of caution is that typically there are 2 copies of jQuery on the page, and that works fine as long as we don't mix them up. This code is wrapped in a closure that imports Civi's copy of jQuery, but Drupal.behaviors passes around drupal's copy as the `context` variable.

To be safe, might need to do a little extra massaging of the `context` variable to strip it from a jQuery object to a dom node (although we have to be careful, it might already be a dom node).

trepmag:
Here is perhaps a more concise explanation of those behaviors in misc/drupal.js: http://cgit.drupalcode.org/drupal/tree/misc/drupal.js?h=7.x#n29

I've wrapped all the code into a closure that use CiviCRM's jQuery, then Drupal will just trigger the behaviors but code will still use CiviCRM's jQuery. Of course I can't promise any regression 0;^)...

Yes normally, the context variable should be passed as a selector or a dom element only and we can only rely on developers to apply this... In general developer using Drupal.attachBehaviors() doesn't pass anything else as the purpose is to apply javascript processing at a given piece of dom. 
But if we may use $(context).get()...

colemanw:
Thanks for the link. This module already uses behaviors extensively and I do know what they are and how to use them, I just don't like them very much :)  I still think that "real" javascript events, triggers and handlers are the way of the future (or if you're not in drupal land, the present). But since this is a drupal 7 module and we are stuck on the drupal island, this seems to be as good as it's going to get.

trepmag:
If you "think that "real" javascript events, triggers and handlers are the way of the future" you should satisfied with those "behaviors"; they are nothing else than that...

Btw, I'm not an evangelist of any stuff, just use what allow me to do things...

trepmag:
Thanks for the merge and you super module!

colemanw:
Let's just agree to disagree :)

trepmag:
;-)

colemanw:
@trepmag looks like this has caused a regression - thoughts on https://www.drupal.org/node/2676752 ?

trepmag:
Unfortunately, I couldn't achieve what I need to do with webform_civicrm: un-obstructive ajaxification; then, I've opt for an alternate implementation (directly with drupal and civicrm apis).

Then, as I dont use webform_civicrm anymore and this PR was only about attempting to do un-obstructive ajaxification, restoring the previous webform_civicrm_payment.js is the way to go...

Sorry for disturbance...

PR #31, opened 2016-01-19
Use system_get_info instead of parsing the info file.

adixon:
By using system_get_info, we respect the hook_system_info calls and avoid improper errors when using the git deploy module.


colemanw:
Learn something new every day...
Thanks :)

PR #32, opened 2016-01-27
Fix IPN payments marked as paid by cheque

guanhuan:
Currently all offsite payments done through webform are marked as paid by cheque. The fix is to simply replicate the logic in CiviCRM contribution page workflow plus a more straightforward condition "if pay_later is selected".
- "is_monetary" is added just to keep consistency but might be a bit redundant. 
- The fix can be enhanced further in the future to acquire the payment instrument for different payment processors

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1672,8 +1672,8 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params['is_pay_later'] = $this->contributionIsPayLater;
     //Fix IPN payments marked as paid by cheque
     if (empty($params['payment_instrument_id'])) {
-      if ($params['payment_processor_id'] != 0 && $this->contribution_page['is_monetary'] == 1) {
-        $params['payment_instrument_id'] = 1;
+      if ($params['payment_processor_id'] && $this->contribution_page['is_monetary']) {
+        $params['payment_instrument_id'] = 'Check';
       }
     }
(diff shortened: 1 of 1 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1670,6 +1670,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params = $this->contributionParams();
     $params['contribution_status_id'] = 'Pending';
     $params['is_pay_later'] = $this->contributionIsPayLater;
+    //Fix IPN payments marked as paid by cheque
+    if (empty($params['payment_instrument_id'])) {
+      if ($params['payment_processor_id'] != 0 && $this->contribution_page['is_monetary'] == 1) {
```
colemanw: This could be simplified as 

```
if ($params['payment_processor_id'] && $this->contribution_page['is_monetary']) {
```

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1670,6 +1670,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+        $params['payment_instrument_id'] = 1;
```
colemanw: I don't think hard-coded ids are a good practice. The api explorer recommends

```
$params['payment_instrument_id'] = 'Check';
```

guanhuan:
Hi Coleman,

Thanks for the comments.

Points taken. Were being very literal about the business logics. Never write code in a hurry...

PR #36, opened 2016-03-24
d.o#2693845 - update _webform_civicrm_status() for CRM-15768

xurizaemon:
Update _webform_civicrm_status() check to account for changes in [CRM-15768](https://issues.civicrm.org/jira/browse/CRM-15768).

Also a typofix for `WEBFROM`

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -10,11 +10,10 @@
 /**
  * The versions of CiviCRM and WebForm. Min is >=.  Max is <. FALSE = no MAX
  */
-define('WEBFROM_CIVICRM_CIVICRM_VERSION_MIN', '4.4');
-define('WEBFROM_CIVICRM_CIVICRM_VERSION_MAX', FALSE);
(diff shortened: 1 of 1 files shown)


colemanw:
This may be premature. Officially, webform_civicrm supports version 4.4+ and this was only fixed in 4.7.

xurizaemon:
Sure, I noted that over in [d.o#2693845](https://www.drupal.org/node/2693845) but didn't do so here or link back.

I _think_ we're OK because `WEBFORM_CIVICRM_CIVICRM_VERSION_MAX` isn't set, but without this patch OR a matching str_replace on `$civicrm['version']`, the change over in CRM-15768 means the installed CiviCRM version (7.x-4.x.x) will always be greater than the values defined in this module (4.x).

So we're OK until CiviCRM 7.x or something? :alien: Anyway spotted this and thought I'd put it on radar.

colemanw:
Don't entirely follow, but I'll take your word for it :)

PR #37, opened 2016-04-19
Fixes for drupal issues 2661232 and 2620630

darrick:
- [2661232: Payment form broken on CiviCRM 4.7](https://www.drupal.org/node/2620630)
- [2620630: Test payment processor status not working](https://www.drupal.org/node/2661232)

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -184,7 +184,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
       );
       $page_id = $this->data['contribution'][1]['contribution'][1]['contribution_page_id'];
       if ($this->civicrm_version >= '4.7') {
-        $currency = '$'; // FIXME
+        $currency = $config->defaultCurrencySymbol; 
         $contributionCallbackQuery = array('currency' => $currency, 'snippet' => 4);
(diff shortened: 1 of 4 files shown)


includes/wf_crm_webform_preprocess.inc
```diff
@@ -183,12 +183,21 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
         'scope' => 'footer',
       );
       $page_id = $this->data['contribution'][1]['contribution'][1]['contribution_page_id'];
-      $contributionCallbackQuery = array('reset' => 1, 'id' => $page_id, 'qfKey' => $this->getQfKey(), 'snippet' => 4);
+      if ($this->civicrm_version >= '4.7') {
+        $currency = '$'; // FIXME
```
colemanw: This value could be retrieved from the contribution page. In fact I think we already have the contribution page loaded in memory at this point. Try `$this->contribution_page`

darrick: Found it in the $config variable.

colemanw: Hmm, that's probably the site-wide default rather than the currency of this contribution page. The webform should follow whatever currency the page is set to.

darrick: Okay.  Found it there.

PR #38, opened 2016-04-20
Location fields saving issue

nishant-bhorodia:
Hello @colemanw . This patch reorders the webform submitted location field values so the following issue could be sorted.

**Issue:**
When saving location fields (email, phone etc.) using webform, location type in civi gets replaced by the location type supplied by the webform. Please see the replication steps for clarification.

**Replication Steps:**
1)Make work phone primary in CiviCRM.
2)In webform, add a "home phone" field.
3)Submit the webform
4)Notice that in civicrm, Work Phone will get replaced by Home Phone

Please let me know what you think.

Many Thanks !!


nishant-bhorodia:
Thanks for the merge @colemanw !!

PR #40, opened 2016-04-28
unset/bypass the CC validation if we're doing Direct Debit (ACHEFT)

KarinG:


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1530,6 +1530,14 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $billing_fields['credit_card_number'] = FALSE;
       $billing_fields['cvv2'] = FALSE;
     }
+    if (!empty($_POST['bank_account_type'])) {
+      // unset/bypass the CC validation if we're doing Direct Debit (ACHEFT) - in that case we have a Bank Account Type
+      $billing_fields['credit_card_number'] = FALSE;
+      $billing_fields['cvv2'] = FALSE;
+      $billing_fields['credit_card_type'] = FALSE;
+      $_POST['credit_card_exp_date']['M'] = '12';
+      $_POST['credit_card_exp_date']['Y'] = '2099';
```
colemanw, line 1539: Is this just to fool the form validation?

KarinG, line 1539: Direct debit payments don't have credit_card_number, cvv2 and credit_card_type fields - so let's set them to FALSE if we are doing direct debit [if we don't we indeed get form validation errors]. CiviCRM expects credit card expiry dates in many places in core - so continue setting these but to far in the future - b/c bank accounts don't have an expiry date.

PR #41, opened 2016-05-12
Use default payment instrument when no instrument is specified

guanhuan:
Change to use core mechanism to keep the consistency while fixing the instrument issue:
ref: civicrm/CRM/Contribute/Form/Contribution.php setDefaultValues() function

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1682,7 +1682,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     //Fix IPN payments marked as paid by cheque
     if (empty($params['payment_instrument_id'])) {
       if ($params['payment_processor_id'] && $this->contribution_page['is_monetary']) {
-        $params['payment_instrument_id'] = 'Check';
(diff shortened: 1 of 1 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1682,7 +1682,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     //Fix IPN payments marked as paid by cheque
     if (empty($params['payment_instrument_id'])) {
       if ($params['payment_processor_id'] && $this->contribution_page['is_monetary']) {
-        $params['payment_instrument_id'] = 'Check';
+        $params['payment_instrument_id'] = key(CRM_Core_OptionGroup::values('payment_instrument', FALSE, FALSE, FALSE, 'AND is_default = 1'));
```
colemanw: The `key()` function takes it parameter by reference, so I'm not sure you can use it like this (at least not on all versions of php).

guanhuan:
Thanks. I have updated the line to pass by reference.
PR to core to correct this as well? https://github.com/civicrm/civicrm-core/blob/master/CRM/Contribute/Form/Contribution.php#L384

guanhuan:
Although I think there should be a better way to handle the payment instrument for IPN payments. The real payment instrument used should be provided by the payment gateway e.g. we would not know if a payment is made by debit card or credit card until the payment detail is filled in on the offsite payment gateway page.

PR #43, opened 2016-08-02
Fix Address fields not being updated

omarabuhussein:
fixing the issue reported here https://github.com/colemanw/webform_civicrm/pull/24#issuecomment-216327009  that address fields (city, country , street address ... etc)  don't get updated after submitting the form.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -1594,6 +1594,21 @@ function wf_crm_location_fields() {
 }
 
 /**
+ * These are the address fields this module supports
+ *
+ * @return array
+ */
+function wf_crm_address_fields() {
+  return array(
+    'street_address',
+    'city',
+    'state_province_id',
+    'country_id',
+    'postal_code'
+  );
+}
+
+/**
  * Returns a count of children of a webform component
  *
  * @param int $nid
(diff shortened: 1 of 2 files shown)


nishant-bhorodia:
Hello @omarabuhussein, @colemanw ,

Code looks good to me. This handles address fields as well as supplemental address too !!
It would be great if @mercurymotos could have a look at this as well.

Many Thanks!

colemanw:
I'll merge this based on your feedback @nishant-bhorodia.
@mercurymotos if you get a chance to test this out that would be helpful too.

omarabuhussein:
@colemanw 

one of our QA team at @compucorp tested it and it works .. to give you a little background on how we use civicrm_webform module , we have this form :

selection_025

where we allow users to edit their data from it , and on this form we tested the patch.

omarabuhussein:
hoooo we submitted our comments nearly at the same time :smile:  , thanks @colemanw

omarabuhussein:
@colemanw  

a quick question ... when are you expecting to release the next stable version ?  .. ( 7.x-4.15 ) is latest one which is released  at ( Feb 19 ) [ which is so long ago ].

colemanw:
@omarabuhussein thanks for the testing. I've just published a new release :)

omarabuhussein:
oh ,, cool :D

mercurymotos:
And I’ve finally shaken off enough of the summer to test as well - it works as expected!  thanks so much.

PR #44, opened 2016-09-13
Filter by relationship - POC (not fully working)

aydun:
This relates to https://www.drupal.org/node/2797341  It is not ready to merge.

@colemanw please provide your feedback on this.  I don't see a way to filter on relationships directly within the Contact.get api, hence this approach.  It works ok in my basic testing, but the biggest question currently is how to make the previously selected contact id's available to  wf_crm_get_contacts() so that the relationship filter can be correctly evaluated.  How can we do that?

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -46,6 +46,7 @@ function _webform_defaults_civicrm_contact() {
         'group' => array(),
         'tag' => array(),
         'check_permissions' => 1,
+        'relationship' => array(),
       ),
     ),
   );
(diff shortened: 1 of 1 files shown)


colemanw:
I think you'll have to pass the id of contact 1 into that function. Note that sometimes the id is unknown, and sometimes it changes dynamically, so if this is an ajax request we'll need to modify the clientside js to grab that contact id as well. In the case of multistep forms I think we will need to inject a hidden form element into pages that follow the one containig cid1.

colemanw:
I'm going to close this until someone wants to pick it up.

PR #46, opened 2016-11-02
Add support for custom fields in dedupe rules

rocxa:
Reatled issue is here:
https://www.drupal.org/node/2711713


colemanw:
Thanks for the PR :)
There are a few non-functional whitespace changes here in the diff which go against Drupals's style guide. Comments are supposed to go on their own line, as are `else {` clauses.

colemanw:
Looks great, thanks :)

PR #48, opened 2016-12-02
Made get entity values compatible with civicrm 4.4

jaapjansma:
See Drupal issue: https://www.drupal.org/node/2832331

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -797,21 +797,15 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
       }
     }
     if ($items) {
-      /**
-       * https://www.drupal.org/node/2832331
-       *
-       * Update an existing case from case1 url parameter fails.
-       * Problem is that IN parameter does not work with the case.get api in CiviCRM 4.4.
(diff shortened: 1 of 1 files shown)


includes/wf_crm_webform_preprocess.inc
```diff
@@ -797,7 +797,26 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
...
+       *
+       * Update an existing case from case1 url parameter fails.
+       * Problem is that IN parameter does not work with the case.get api in CiviCRM 4.4.
+       * So we check whether the version of CiviCRM is 4.4 and if so we will
+       * do an API call on each item.
+       */
+      $civiVersion = wf_crm_apivalues('Domain', 'get', array('return' => 'version'));
+      $civiVersion = reset($civiVersion);
+      $civiVersion = (float) substr($civiVersion['version'], 0, 3);
+      if ($civiVersion < 4.6) {
```
colemanw: No need to fetch version via api. This is already available to this object; use `$this->civicrm_version`.

Use `version_compare` not `<`

colemanw (changes requested):

jaapjansma:
I have updated the code to use version_compare. Thanks for the tip @colemanw !

colemanw (approved):

PR #50, opened 2016-12-06
Issue #2769423 Fields specified in Existing Contact:"Fields to lock" are not stored

ghost:


colemanw:
My take on this is that having hidden fields not passed through to the submission is a bug. If memory serves me right, it's actually a regression. I seem to recall it working before. So this is more of a workaround than a fix.

PR #53, opened 2016-12-20
Include entire webform submission in activity details when updating an existing activity

jaapjansma:
Fixed issue 2837945 (https://www.drupal.org/node/2837945). Also update the detail of an existing activity and give the user the selection to disable this.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1303,7 +1303,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $params['id'] = $this->ent['activity'][$n]['id'];
 
           // Update details when user has selected he wants to update the details
-          if (empty($this->data['activity'][$n]['details']['no_update_existing'])) {
+          if (!empty($this->data['activity'][$n]['details']['update_existing'])) {
             // Format details as html
             $params['details'] = nl2br(wf_crm_aval($params, 'details', ''));
             // Add webform results to details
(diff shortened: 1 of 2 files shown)


colemanw (changes requested):
This looks like some repetitive code. Can it be moved into a function for reusability instead of copy/pasted?

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1301,6 +1301,30 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         // Update mode
         else {
           $params['id'] = $this->ent['activity'][$n]['id'];
+
+          // Update details when user has selected he wants to update the details
+          if (empty($this->data['activity'][$n]['details']['no_update_existing'])) {
```
colemanw: Because you've implemented this as a negative (do not update), this is going to change the behavior of all existing installs, because this condition will be true for all of them. Doing it as a positive (update details) would keep existing forms the same.

jaapjansma: I have changed it

jaapjansma:
@colemanw I have changed the code to be positive worded. Rather than negative ('Do not update existing...' is now 'Update existing...')

colemanw:
Great. And what about moving the redundant code into a function?

jaapjansma:
I have moved the redudant code into a function.

colemanw (approved):

colemanw:
@jaapjansma this may have caused a regression. Can you please take a look at https://www.drupal.org/node/2858811

PR #54, opened 2016-12-23
For on small screens

KarinG:
On the left: big screens (400px - as it was) - on right: 100% to fit on small screens:

400max


colemanw:
I like it.

PR #56, opened 2017-01-12
Recur v1

KarinG:


colemanw:
Nice to see contributionRecur living in its own function. Neat & tidy.

KarinG:
😀 Still needs form work but these are the mechanics of adding recurring functionality;

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1689,7 +1689,17 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    * This happens during form validation and sets a form error if unsuccessful
    */
   private function submitLivePayment() {
-    $result = wf_civicrm_api('contribution', 'transact', $this->contributionParams());
+    $contributionParams = $this->contributionParams();
+
+    // Only if #installments = 1, do we process a single transaction/contribution. #installments = 0 in CiviCRM Core means open ended commitment;
+    $numInstallments = wf_crm_aval($contributionParams, 'installments', 1, $strict = TRUE);
+    if ($numInstallments === 1) {
```
colemanw: The `===` operator will fail if it's a string value "1", which it probably is coming from a form submission.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1689,7 +1689,17 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    * This happens during form validation and sets a form error if unsuccessful
    */
   private function submitLivePayment() {
-    $result = wf_civicrm_api('contribution', 'transact', $this->contributionParams());
+    $contributionParams = $this->contributionParams();
+
+    // Only if #installments = 1, do we process a single transaction/contribution. #installments = 0 in CiviCRM Core means open ended commitment;
+    $numInstallments = wf_crm_aval($contributionParams, 'installments', 1, $strict = TRUE);
```
colemanw: No need to define the variable `$strict` here.

includes/wf_crm_webform_postprocess.inc
KarinG: === is actually working properly - but will change this to == (I just verified that == is working fine for three cases: 0, 1 and >1 (e.g. 10)

includes/wf_crm_webform_postprocess.inc
KarinG: Oops - fixing that!

KarinG:
Tested a bunch of scenarios on 4.6.x w/ iATS Payment Processor Extension [an example of a processor which implements CiviCRM's core doDirectPayment - Live mode; I looked at #installments = 1; #installments = 0, #installments = 10, monthly, weekly, every 2nd month. All looking good in that all the monies were transacted properly - and the Recurring series all looked good and are properly linked to tokenized credit cards.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1693,11 +1693,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
 
     // Only if #installments = 1, do we process a single transaction/contribution. #installments = 0 in CiviCRM Core means open ended commitment;
     $numInstallments = wf_crm_aval($contributionParams, 'installments', 1, TRUE);
-    if ($numInstallments == 1) {
-      $result = wf_civicrm_api('contribution', 'transact', $contributionParams);
+    $frequencyInterval = wf_crm_aval($contributionParams, 'frequency_unit', NULL, TRUE);
```
colemanw: This may be overkill. You're setting the `wf_crm_aval` function to use strict mode here but then using the not-so-strict `!empty()` check below. It's probably safe to leave off the last 2 params here.

KarinG: Ok let me take these out and run some extra/debugs just to make sure

KarinG: Yep you're correct! Ok fixed. And quickly ran some extra tests.

PR #63, opened 2017-03-28
Civipostcode integration

nishant-bhorodia:
Hello @colemanw , 

Hope you are well.

**Requirement:** Integrate [CiviPostcode Exntension](https://github.com/veda-consulting/uk.co.vedaconsulting.module.civicrmpostcodelookup) with webform_civicrm.

**What this PR does:** This PR integrates functionality of civipostcode extension to webform civicrm by adding a webform component called "Civi Postcode". 
Administrators can change widget type of "postal code" fields to Civi Postcode to transform it from a _regular textfield_ to a _Postcode lookup field_ which makes use of the resources provided by  [CiviPostcode Exntension](https://github.com/veda-consulting/uk.co.vedaconsulting.module.civicrmpostcodelookup).

**Test Results:**
1. Check if postcode can be looked up in the Postal code field - Passed
2. Check if both capital and small letters can be used in the postal code field - Passed
3. Check if you enter name and email, etc, the post code lookup still works - Passed
4. Check if you can select an address and it gets populated in the fields - Passed
5. Check if you select an address and then change the postcode, and select another address, the new address gets populated - Passed
6. Check if an address has a company name before the address, if that is getting populated too as in the civicrm version - Passed
7. Check if you can change the widget type back to textfield - Passed
8. Check if you add 2 civi contacts with addresses, in both cases civi postcode works - Passed
9. When you go to the preview page, and check the addresses for the contacts, you will be able to see the postcode - Passed
10. When you go to the preview (next) page and then go back to the previous one, the postcode is still there - Passed
11. Check if the lookup is working when address sharing is enabled for 2 contacts - Passed

**Demo GIF:** civipostcode-webform-integration-demo-after-changi

Many Thanks.

Kind Regards,
Nishant Bhorodia

### DIFF
--- a/includes/civipostcode_component.inc
+++ b/includes/civipostcode_component.inc
@@ -37,6 +37,7 @@ function _webform_display_civipostcode($component, $value, $format = 'html') {
 }
 
 function _webform_render_civipostcode($component, $value = NULL, $filter = TRUE) {
+
    $element['postcode'] = array(
     '#type' => 'textfield',
     '#title' => $filter ? webform_filter_xss($component['name']) : $component['name'],
@@ -46,11 +47,6 @@ function _webform_render_civipostcode($component, $value = NULL, $filter = TRUE)
     '#attributes' => array(),
     '#theme_wrappers' => array('webform_element'),
     '#webform_component' => array('type' => $component['type'] . '_civipostcode') + $component,
-    '#attached' => array(
-      'js' => array(
-        drupal_get_path('module', 'webform_civicrm') . '/js/civipostcode_component.js',
-      ),
-    )
   );
 
   return $element;
--- a/includes/civipostcode_component.inc
+++ b/includes/civipostcode_component.inc
@@ -1,0 +1,57 @@
+<?php
+
+function _webform_edit_civipostcode($component) { 
+  $form = array();
+
+  $form['#attached']['js'][] = wf_crm_tokeninput_path();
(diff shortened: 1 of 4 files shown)


colemanw:
This looks like good functionality. But IMO all of this code ought to live in a separate module, using the same hooks that this one uses.

guanhuan:
Hi @colemanw, this is very CiviCRM specific functionality. It supports Webform CiviCRM but also largely depending on it. It might make sense to make this available to others via Webform CiviCRM in this case. We can add a bit README for this if needed.

colemanw:
Yes it is specific to civi, but just as the postcode functionality is in an extension rather than civicrm core, I think this addition should be separate from webform_civicrm core. So a module named e.g. `webform_civicrm_postcode` could be bundled with the postcode extension to enable integration with webforms.

colemanw:
The main problem is packaging. Unfortunately a Civi extension cannot implement Drupal hooks (because Drupal won't look for it when invoking hooks). Bundling the module in the extension repo would work fine but would require site admins to manually move the module code into their modules directory, which makes upgrades a bit of a pain.
I think the best solution would be to make a proper module and publish it on drupal.org so it can be auto-installed and updated.

nishant-bhorodia:
Hello @colemanw , 
Sorry for the delay is replying.
Thank you for your suggestions. I see your point here. I will create a separate module for this.
Many Thanks!

PR #64, opened 2017-03-31
Fix Webform overwriting membership data

jitendrapurohit:
@colemanw Not sure of this, but I didn't found anywhere to raise an issue related to `webform_civicrm`, hence would like to discuss with a PR.

Problem : If user has an existing membership of *any* type (say 'Student'), submitting a webform of another type(General) would overwrite the existing one.

This is because the previous code would collect membership id if it finds the organization id to be same as that of the existing one. I think we should even check for membership type.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1101,20 +1101,19 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         continue;
       }
       // Search for existing membership to renew - must belong to same domain and organization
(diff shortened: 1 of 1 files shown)


colemanw:
This is by design. Signing up for a membership of a different type within the same organization is considered an "upsell" e.g. changing your membership level from Silver to Gold.

If you wish to allow separate concurrent memberships, make separate member organizations for them.

jitendrapurohit:
make sense, somehow didn't figure out the upsell reason. Thanks.

PR #66, opened 2017-03-31
Set null for multichoice custom field if value is empty

sunilpawar:


colemanw:
@sunilpawar could you provide steps to reproduce this bug?

colemanw:
Waiting for response from @sunilpawar to move this forward.

sunilpawar:
@colemanw i will get back with details, its one year old task.

PR #67, opened 2017-04-01
DRU-2856493 - reProcess the relationship to help inheritance of Membe…

KarinG:
…rship.


KarinG:
reProcessing the relationships - after the Membership has been created - helps to ensure Membership is inherited properly. I've had this in use for 10days now for one project and though and I've not seen any missed inheritances, I'll continue to keep a close eye on this - and report back if needed.

PR #68, opened 2017-04-04
Fix soft credit contribution does not get created

chanunc:
This is a fix of soft credit contribution does not get created when making payment through webform which configured with Contribution page and “Soft Credit To”.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1960,12 +1960,24 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $this->saveCustomData($this->data['contribution'][1], $id, 'Contribution', FALSE);
     // Save soft credits
     if (!empty($contribution['soft'])) {
+      // Get total amount either from contribution object or the total amount after line item calculation
+      $amount = $contribution['total_amount'] != null ? $contribution['total_amount'] : $this->totalContribution;
+      
+      // Get Default softcredit type
+      $default_soft_credit_type = wf_civicrm_api('OptionValue', 'getsingle', array(
+        'sequential' => 1,
+        'return' => "value",
+        'option_group_id' => "soft_credit_type",
+        'is_default' => 1,
+      ));
+
       foreach (array_filter($contribution['soft']) as $cid) {
         wf_civicrm_api('contribution_soft', 'create', array(
           'contact_id' => $cid,
           'contribution_id' => $id,
-          'amount' => $contribution['total_amount'],
+          'amount' => $amount,
           'currency' => $this->contribution_page['currency'],
+          'soft_credit_type_id' => $default_soft_credit_type['value'],
         ));
       }
     }
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1965,7 +1965,6 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       
       // Get Default softcredit type
       $default_soft_credit_type = wf_civicrm_api('OptionValue', 'getsingle', array(
-        'sequential' => 1,
         'return' => "value",
         'option_group_id' => "soft_credit_type",
         'is_default' => 1,
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1960,8 +1960,8 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $this->saveCustomData($this->data['contribution'][1], $id, 'Contribution', FALSE);
     // Save soft credits
     if (!empty($contribution['soft'])) {
-      // Get total amount either from contribution object or the total amount after line item calculation
-      $amount = $contribution['total_amount'] != null ? $contribution['total_amount'] : $this->totalContribution;
+      // Get total amount from total amount after line item calculation
+      $amount = $this->totalContribution;
       
       // Get Default softcredit type
       $default_soft_credit_type = wf_civicrm_api('OptionValue', 'getsingle', array(


colemanw:
By design, you should create the soft credits within the webform and not on the contribution page. Maybe we could document this better.

Unless we get some votes in favor of this feature I think we probably should not add it.

guanhuan:
@chanunc can you add in complete issue description for this please.

Hi @colemanw, 

I don't think this a new feature (correct me if I am wrong). Currently the "Soft Credit To" field doesn't do anything when line item other than the contribution amount is involved and this is for fixing this issue.

colemanw:
@chanunc maybe I'm just not clear about what problem this is trying to solve. Can you please post steps to reproduce the bug?

colemanw:
Happy to reopen this if it gets clarified.

chanunc:
Hi Coleman, 
The business logic here is to create a webform for a buyer to purchase a membership for a recipient.

To reproduce 
### 1 .Setup Contribution page for membership payment
- Financial Type: Member Dues
- Set up dummy processor
- Enable Membership section
- Select a membership type (any)

### 2. Setup Webform and enable CiviCRM processing
- Add 2 individuals, one for buyer, one for recipent -> Save settings
- Select 0 Number of Memberships for buyer and select 1 to Number of Memberships for recipient -> Save settings
- Set up Contribution page to the one on the 1st step
- Set Soft Credit To: recipient -> Save settings

### 3. Proceed the webform to purchase membership for a recipient.
- Fill buyer and recipient form
- Purchase membership with test payment
- After complete the form, open the recipient contact page to check membership and contribution.

**Expected result:** 
The recipient contact should have a membership and contribution with soft credit from buyer contact.

**Faulty result:**
The recipient contact has a membership but not contribution with soft credit.

What I have found is Webform sent empty amount when it call CiviCRM API to create soft credit, therefore I have added a logic to get the amount firstly from contribution object, if the amount is empty then get the amount from line item calculation. 
More over, I also add default soft create type as a parameter when Webform call CiviCRM API to create soft credit.

guanhuan:
@colemanw please reopen. Please let @chanunc know if it all make sense to you and feedback if any. Thanks

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1960,12 +1960,24 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $this->saveCustomData($this->data['contribution'][1], $id, 'Contribution', FALSE);
     // Save soft credits
     if (!empty($contribution['soft'])) {
+      // Get total amount either from contribution object or the total amount after line item calculation
+      $amount = $contribution['total_amount'] != null ? $contribution['total_amount'] : $this->totalContribution;
```
agileware: This does not yield correct results if there is *also* a contribution amount field - and under some circumstances, e.g. a 5¢ additional contribution, reverts to the behaviour where the soft credit is not recorded.
I think better to just use `$this->totalContribution` regardless?

ghost (changes requested):
Shouldn't conditionally use the total amount, as this might be filled with a portion of the total

agileware:
Added an issue [https://www.drupal.org/node/2882281](https://www.drupal.org/node/2882281) to the queue on drupal.org

chanunc:
@agileware Agree!
I have changed following your suggestion and tested with new installation of Drupal, Webform and CiviCRM.

guanhuan:
@colemanw can you take a look at this please? thank you

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1960,12 +1960,24 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $this->saveCustomData($this->data['contribution'][1], $id, 'Contribution', FALSE);
     // Save soft credits
     if (!empty($contribution['soft'])) {
+      // Get total amount from total amount after line item calculation
+      $amount = $this->totalContribution;
+      
+      // Get Default softcredit type
+      $default_soft_credit_type = wf_civicrm_api('OptionValue', 'getsingle', array(
+        'sequential' => 1,
```
colemanw: `sequential` doesn't make sense as a param for api.getsingle.

nishant-bhorodia:
Hello @colemanw , 
Hope you are well!
Thanks for checking it out. I have updated the PR and removed sequential. 
Many Thanks :)

KarinG:
Before: on 4.6 is that the Soft Credit does get created - but that the amount is incorrect:
image

KarinG:
Type is indeed empty: 
image

KarinG:
@chanunc - please note that you do not need to set up (on Contribution Page): "Enable Membership section" "Select a membership type (any)"

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1960,12 +1960,23 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+      // Get total amount from total amount after line item calculation
+      $amount = $this->totalContribution;
+      
+      // Get Default softcredit type
+      $default_soft_credit_type = wf_civicrm_api('OptionValue', 'getsingle', array(
+        'return' => "value",
+        'option_group_id' => "soft_credit_type",
+        'is_default' => 1,
+      ));
+
```
KarinG, line 1972: Default ("Solicited" unless configured by admins) is ok for now - though there are 11 options in CiviCRM native - key purpose of this PR is to fix the $0.00 amount;

KarinG:
To summarize: 

- I've reproduced the issue on 4.6
- I've confirmed it's fixed on 4.6 (with this PR)

image

If you can show us some screenshots that this is working on 4.7 - then this can be merged in.

colemanw:
Ok sounds like this is a net benefit. Merging.

nishant-bhorodia:
Thanks @colemanw @KarinG :)

PR #69, opened 2017-04-05
Revert "DRU-2856493 - reProcess the relationship to help inheritance of Membe…"

colemanw:
Reverts colemanw/webform_civicrm#67


colemanw:
The solution didn't fully work, so rolling it back for now.

PR #70, opened 2017-04-11
Change Tax rate to a floating point value

vincent1892:
VAT appears as 20.0000000000% no matter how many times you save it in civicrm "/civicrm/admin/financial/financialAccount?reset=1", this shows up on webforms as well. Fixed by changing the $this->tax_rate to a floating point value. This strips out the trailing zeros.

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -549,9 +549,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
         $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
 
         if ($taxSettings['tax_display_settings'] != 'Do_not_show') {
-          //strip trailing zeros from tax_rate e.g 20.00000% => 20.00%
-          $tax_rate = number_format($this->tax_rate, 2);
-          $item['label'] .= ' (' . t('includes !rate @tax', array('!rate' => $tax_rate . '%', '@tax' => $taxSettings['tax_term'])) . ')';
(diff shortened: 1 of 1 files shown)


includes/wf_crm_webform_preprocess.inc
```diff
@@ -549,7 +549,8 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
         $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
 
         if ($taxSettings['tax_display_settings'] != 'Do_not_show') {
-          $item['label'] .= ' (' . t('includes !rate @tax', array('!rate' => $this->tax_rate . '%', '@tax' => $taxSettings['tax_term'])) . ')';
+          $tax_rate = number_format($this->tax_rate, 2); //strip trailing zeros from tax_rate e.g 20.00000% => 20.00%
```
meladawy: Please add the comment in a separate line following instructions in https://www.drupal.org/node/1354

meladawy (changes requested):

vincent1892:
@meladawy  I've changed the comment style

vincent1892:
@colemanw sorry to bother but could you please review this PR, we have a couple of tickets that are blocked cause of this?

jamienovick:
@vincent1892 this **shouldn't** be fixed to a number of d.p. -> It shouldn't show any '0' after the decimal place. i.e. If its 20.00000% it should show 20%; if its 20.125000% then 20.125% that should be shown.

colemanw:
@vincent1892 can you please check Jamie's comment?

vincent1892:
@jamienovick @colemanw I've updated the PR. Webform now displays VAT without the trailing zeros.

PR #71, opened 2017-04-19
Issue #2868220 - Webform not linking to CiviCRM when other language i…

francescbassas:
…s selected

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -160,12 +160,12 @@ function webform_civicrm_node_delete($node) {
  * @param $langcode
  */
 function webform_civicrm_node_view($node, $view_mode, $langcode) {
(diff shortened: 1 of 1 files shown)


colemanw:
Thanks. Please use 2 spaces instead of tabs.

francescbassas:
Done.

colemanw:
I notice that there is no condition in here to check whether or not the node is a webform, and whether the `webform_civicrm` property exists in the source node.

francescbassas:
Thanks, I commited your corrections.

colemanw:
@francescbassas I've just pushed another change to try to make the conditions safer. Can you please test on your site and ensure it still works.

francescbassas:
Thanks @colemanw. Tested and still working.

PR #72, opened 2017-05-15
Issue #2878590 - warning message on invalid checksum links

laryn:
https://www.drupal.org/node/2878590

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -683,6 +683,10 @@ function wf_crm_contact_access($component, $filters, $cid) {
       elseif (!empty($_GET["cs$c"]) && CRM_Contact_BAO_Contact_Utils::validChecksum($cid, $_GET["cs$c"])) {
         $filters['check_permissions'] = FALSE;
       }
+      // If the checksum is present but invalid, add a message.
+      if ($filters['check_permissions'] != FALSE) {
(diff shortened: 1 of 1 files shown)


colemanw:
A couple things about this:

1. I don't think this is the right place in the code to add a warning message. This is a utility function so it probably shouldn't be setting messages.
2. The message doesn't apply to all use-cases. We probably need a per-form configuration option to enable or configure it. Similar to the "not you" message.

laryn:
@colemanw 

1. Any tips on how I can send the information back to a more appropriate place to set a message?

2. I will try to copy existing code to do this. Seems to make sense to put it in the "Additional options" section where checksums are already discussed, right?

colemanw:
@laryn 

1. Check the spot where the "not you" message is set in the code. It could perform a similar check, namely loop through all contacts on the form, see if there is a corresponding $_GET['cidx'] and set the message if that contact has not been verified. I don't think we need to check for the presence of a checksum, as this option would be more generically useful if we just check for unknown contacts with a cid in the url.
2. Sounds good.

colemanw:
@laryn any progress? Want to keep this open?

laryn:
@colemanw It's in my rainy day pile and I haven't had a chance to work on it for some time. I'll close for now and will reopen when I have something to show.

PR #73, opened 2017-05-17
ensure total and tax amount are rounded upto two decimal units

jitendrapurohit:
Having long decimals in `tax_amount` and `total_amount` is not considered as type = MONEY by civi and fails the validation in contribution create api.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1515,6 +1515,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         $this->totalContribution += $item['line_total'] = $item['unit_price'] * (int) $item['qty'];
       }
     }
+    $this->totalContribution = round($this->totalContribution, 2);
     return $this->totalContribution;
   }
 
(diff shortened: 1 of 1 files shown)


petednz:
I just tested on a form that has gst, and the outcome, on a form that registers multiple people on multiple events which are using GST, was a complete success as far as I could tell. Not seen Receipts as were sent to DB. Will test on live next

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1892,7 +1893,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     // TODO: needs review if this is changed in 4.6
     if (!is_null($this->tax_rate)) {
       $params['non_deductible_amount']  = $this->totalContribution;
-      $params['tax_amount'] = ($params['total_amount'] / ($this->tax_rate + 100)) * $this->tax_rate;
+      $params['tax_amount'] = round(($params['total_amount'] / ($this->tax_rate + 100)) * $this->tax_rate, 2);
     }
     $params['description'] = t('Webform Payment: @title', array('@title' => $this->node->title));
```
KarinG, line 1898: Yikes - this is not your line - but that's very wrong. Non-deductible amounts have absolutely nothing to do with sales taxes. 

`        $params['non_deductible_amount']  = $this->totalContribution; `

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1515,6 +1515,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         $this->totalContribution += $item['line_total'] = $item['unit_price'] * (int) $item['qty'];
       }
     }
+    $this->totalContribution = round($this->totalContribution, 2);
     return $this->totalContribution;
```
KarinG, line 1519: Rounding happens after multiplying by qty - excellent.

includes/wf_crm_webform_postprocess.inc
KarinG: That does make me question this section - and why tax_amount is calculated here [as well].

KarinG:
Just a note that I've pinged Guanhuan on that sales tax vs income tax confusion (on his original PR).

PR #74, opened 2017-06-13
Add new field to create relationship in additional options section

jitendrapurohit:
Fix for https://www.drupal.org/node/2511208

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -345,7 +345,7 @@ abstract class wf_crm_webform_base {
       foreach($values as $key => $value){
         if((in_array($ent, array('address', 'email')) && $value['location_type_id'] == $setting['location_type_id'])
            ||
-          ($value['location_type_id'] == $setting['location_type_id'] && $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])     
+          ($value['location_type_id'] == $setting['location_type_id'] && $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])
           ){
             $reorderedArray[$key] = $value;
             $valueFound = true;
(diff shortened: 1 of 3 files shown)


includes/wf_crm_webform_base.inc
```diff
@@ -417,6 +417,10 @@ abstract class wf_crm_webform_base {
    */
   protected function getRelationship($r_types, $cid1, $cid2) {
     $found = array();
+    $active_clause = '';
+    if ($this->settings['create_new_relationship']) {
```
colemanw: I think you should wrap this variable in `!empty()` to avoid a possible php notice if it's not defined.

colemanw (changes requested):

includes/wf_crm_webform_base.inc
jitendrapurohit: Updated :-)

includes/wf_crm_webform_postprocess.inc
```diff
@@ -966,6 +966,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $perm = $perm == 1 ? 2 : 1;
         }
       }
+      //initialise start_date for create action.
+      if (empty($existing) !array_key_exists('start_date', $params) {
```
colemanw: I'm having trouble reading this if statement. There's no operator between the two conditions.
Also it looks like there's a missing closing bracket. Did you try running this code? Looks like you'd get a fatal syntax error.

colemanw (changes requested):

includes/wf_crm_webform_postprocess.inc
jitendrapurohit: seems like it got messed up while pushing, was working fine while testing. I've tested the flow again and confirmed that it creates a new relationship when an expired/inactive duplicate relationship exists for a contact.

Thanks for spotting!

PR #75, opened 2017-07-04
Add setting to the additional options

jitendrapurohit:
Follow up of https://github.com/colemanw/webform_civicrm/pull/74 to add the setting on the form (Additional Options). This also modifies relationship `start_date` to start from today instead of being `NULL`.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -968,7 +968,8 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
       //initialise start_date for create action.
       if (empty($existing) && !array_key_exists('start_date', $params)) {
-        $params['start_date'] = NULL;
+        $today = date('Y-m-d');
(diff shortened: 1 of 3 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -968,7 +968,8 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
       //initialise start_date for create action.
       if (empty($existing) && !array_key_exists('start_date', $params)) {
-        $params['start_date'] = NULL;
+        $today = date('Y-m-d');
+        $params['start_date'] = $today;
```
colemanw: Pretty sure the api will accept the string 'now'

jitendrapurohit:
Have included `now` string confirming that it works. Also did a minor tweak to consider `end_date IS NULL` as active. Thanks.

jitendrapurohit:
Thanks @colemanw

PR #76, opened 2017-07-11
Daynamic custom fields improvments

omarabuhussein:
This is work done by my colleague @nishant-bhorodia  .

This patch implements the following : 

1. When you click add fields dynamically on webforms civicrm tab, a webform field set for that custom field set is created in the webform tab.
* The field set on Webform tab takes the name of the civicrm fieldset (reflected on CiviCRM tab)
* the fields on the Webform tab take the name of the civicrm fields and is added under the specific field set as configured in CiviCRM (currently it puts them under an entity field set - webform 4.18 module).

2. Even if the custom field set doesn't have any custom fields in it yet and 'Add dynamically' is clicked, you should still see the field set in:
* CiviCRM tab (_Note: you still see the field set on CiviCRM tab even if 'Add dynamically' is not clicked_)
* webforms tab (_Note: you see all the changes to a field set and its fields once 'Add Dynamically' is clicked_)

3. From CiviCRM Admin > Custom data when you:
* add
* edit e.g.: rename, add/delete options for select lists etc.
* delete
* disable 

the fields from the CiviCRM field set, this is going to be reflected on both Webform tab and CiviCRM tab.

4. When you disable/ delete all fields in a field set in CiviCRM, the field set should still appear on CiviCRM tab and webform tab, and remain empty.

5. User is able to edit all details of field set other than machine name on the webform tab

6. If you untick 'Add dynamically' - it doesn't need to remove the fields from webform and civicrm tab if the fields are still ticked on civicrm tab. User needs to remove fields individually and save (default behaviour that can stay).

7. If the field sets are on different pages, when adding new fields to a field set in CiviCRM, these should be placed in the correct field set under its respective page.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -23,7 +23,9 @@ function wf_crm_field_options($field, $context, $data) {
   if ($pieces = wf_crm_explode_key($field['form_key'])) {
     list( , $c, $ent, $n, $table, $name) = $pieces;
     // Ensure we have complete info for this field
-    $field += $fields[$table . '_' . $name];
+    if(isset($fields[$table . '_' . $name])) {
+      $field += $fields[$table . '_' . $name];
+    }
     if ($name === 'contact_sub_type') {
       list($contact_types, $sub_types) = wf_crm_get_contact_types();
       $ret = wf_crm_aval($sub_types, $data['contact'][$c]['contact'][1]['contact_type'], array());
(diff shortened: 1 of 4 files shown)


omarabuhussein:
@colemanw 

can you please review this .. regarding QA it was tested on clean civicrm by our QA and they confirmed that all the points above working fine.

colemanw:
@omarabuhussein is this for CompuCorp?

omarabuhussein:
yes @colemanw

jamienovick:
@colemanw this was actually a clean up of some of the work that you did previously I believe but there were some bugs and we didn't want to disrupt cases work. We are a little pressed to get these webform improvements out if you can give them some attention @guanhuan FYI to allocate some time to this. Thanks J

colemanw:
Tested it out and reviewed the code. Works well.

omarabuhussein:
thanks man

jamienovick:
Thanks both!

PR #77, opened 2017-07-11
Add token to show a link for created case after webform submission

omarabuhussein:
This is another PR developed by my colleague @nishant-bhorodia .

This basically add new token called **[submission:case_link]**  that you can added on the webform submission Confirmation message to show a link for the case created by webform_civicrm  module (in case the webform creates a case).

to be honest I am a bit skeptical if this should be merged into webform_civicrm module but I thought that maybe we can add more tokens too later like (for activities , contacts .. etc) so users can use them  out of the box and without the need to do any customization  . What do you think @colemanw ? 

----------------------------------------------------------------------------------------------------------

# **Update 1 :**

Following the discussion below ,  I've implemented a support for case links token if there was more than one case and also added support for contact and activity link tokens .

**Usage :** 

If the webform is configured to create (2 contacts , 2 activities and 3 cases) , then we can use any of the following tokens :

_**Hence that these tokens will only work after submitting the webform only.**_

For the two contacts : 
- *[submission:contact-link:1]*
- *[submission:contact-link:2]*

For the two activities: 
- *[submission:activity-link:1]*
- *[submission:activity-link:2]*

For the three cases: 
- *[submission:case-link:1]*
- *[submission:case-link:2]*
- *[submission:case-link:3]*

so basically add the number of the entity (case, activity or contact) to the end of token. so if your webform is configured to created 11 cases , then you can show the link for case number 5 by using this token : 

- *[submission:case-link:5]*

**Usage in action:** 

1- Adding the tokens : (hence that these gifs is a bit outdated , token names now are following singular instead of plural form (so for example case-link:1 instead of case-link**s**:1)

2- Submitting the form  :

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -15,6 +15,8 @@ define('WEBFORM_CIVICRM_CIVICRM_VERSION_MAX', FALSE);
 
 define('WEBFORM_CIVICRM_WEBFORM_VERSION', '4.12');
 
+define('WEBFORM_CIVICRM_DEFAULT_CONTACT_ID', 1);
+
 /**
  * Implements hook_menu().
  *
@@ -637,10 +639,22 @@ function _webform_civicrm_status() {
  * Implements hook_token_info().
  */
 function webform_civicrm_token_info() {
-  // Create case link token
-  $info['tokens']['submission']['case_link'] = array(
-    'name' => t('Case Link'),
-    'description' => t('The link to case that got created after submission.'),
+  $info['tokens']['submission']['contact-links'] = array(
+    'name' => t('Webform CiviCRM Contacts Links'),
+    'description' => t('The links to Contacts that got created after submission the webform. Replace the "?" with the contact number starting from 1'),
+    'dynamic' => TRUE,
+  );
+
+  $info['tokens']['submission']['activity-links'] = array(
+    'name' => t('Webform CiviCRM Activity Links'),
+    'description' => t('The links to activities that got created after submission webform. Replace the "?" with the activity number starting from 1'),
+    'dynamic' => TRUE,
+  );
+
+  $info['tokens']['submission']['case-links'] = array(
+    'name' => t('Webform CiviCRM Case Links'),
+    'description' => t('The links to cases that got created after submission webform. Replace the "?" with the case number starting from 1'),
+    'dynamic' => TRUE,
   );
 
   return $info;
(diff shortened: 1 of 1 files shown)


colemanw:
Hmm.

:+1: For the idea of tokens

But I'm not sure about this one. For one thing, a webform can create > 1 case.

omarabuhussein:
hmmm right .. what about using numbers .. such as [submission:case_link_1], [submission:case_link_2] or something like that for each case the form can create ?

colemanw:
That could work. If we're going to do this I'd like to see some basics covered. Could you also add a link for each contact? That would make this PR feel a bit more solid.

omarabuhussein:
yup .. will work on contacts and activities at least . thanks man.

colemanw:
Very nice :)
Has anyone on your side tested this yet?

omarabuhussein:
not yet @colemanw  .. I tested it myself for now and will assign it to someone to test it soon .. I just wanted to push and update the PR description for now .. also I have a problem with webform_civicrm_tokens hook .. it seems that civi confuse between it and hook_civicrm_tokens  due to the name of this module and the hook so a warning like this appear on the top of the page after submitting any webform : 

`Warning: Missing argument 2 for webform_civicrm_tokens(), called in /vagrant/dmaster/sites/all/modules/civicrm/CRM/Utils/Hook.php on line 270 and defined in webform_civicrm_tokens() (line 666 of /vagrant/dmaster/sites/all/modules/contrib/webform_civicrm/webform_civicrm.module).`

so I am trying to think of a way to work around that without doing any change to civi core

colemanw:
Yikes. That's not good. Maybe the workaround is just to put in an early return if the params given match what would be coming from the civi hook instead of the webform hook.

omarabuhussein:
well, since the warning was about the number of arguments passed from civicrm hook invoker [not sure if invoker is a correct work ] (civicrm hook invoker only send 1 argument while drupal hook invoker sends too )  , I just added default empty value to the 2nd hook_tokens parameter ($tokens = '') and the warning gone .

I will update you once this get QAed and passed testing.

colemanw:
This looks excellent.

One nitpick. I would prefer the singular `[submission:contact-link:1]` instead of the plural `[submission:contact-links:1]` since each token is for a single link.

omarabuhussein:
sounds reasonable

colemanw:
@omarabuhussein Thanks for this. Would you also please take a minute to update the wiki page?
https://wiki.civicrm.org/confluence/display/CRMDOC/Webform+CiviCRM+Integration

omarabuhussein:
ops that was quick .. I wanted to let our QA test if first before merging it  .. but the good news is that our QA tested it not so long after your merged it and it pass testing (there was some small typos that I fixed in new PR) so all good .

Anyway thank so much @colemanw  .. I will update the wiki.

omarabuhussein:
seems that I need to request an account on confluence first .

colemanw:
@omarabuhussein after merging I found a few syntax issues that would cause problems on older versions of PHP. Fixed those along with some general code cleanup in ea7f795

omarabuhussein:
thanks @colemanw

PR #80, opened 2017-07-14
Allow thousand separator value a valid amount for money field

jitendrapurohit:
Money field uses `CRM_Utils_Type::escape('Money', $dt, FALSE)` to validate its value. This function  checks [for numeric value](https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/Type.php#L260) and fails `1,000`  contribution from the webform.

image

Either this needs fix in core to use `CRM_Utils_Rule::money()` instead of `numeric()` or an extra condition for Money datatype in `webform_civicrm.`

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2181,8 +2181,11 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $sum = 0;
           foreach ((array) $val as $k => $v) {
             //Strip any thousand separator for further calculation.
+            $separator = wf_civicrm_api('Setting', 'getvalue', array(
+              'name' => "monetaryThousandSeparator",
+            ));
             if ($field['data_type'] === 'Money') {
-              $v = preg_replace('/(?<=\d),(?=\d{3}\b)/', '', $v);
+              $v = preg_replace("/(?<=\d){$separator}(?=\d{3}\b)/", '', $v);
             }
             // Perform multiplication across grid elements
             if ($component['type'] == 'grid' && is_numeric($k)) {
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -337,6 +337,11 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
               form_error($element, t('Mismatch: "@state" is not a state/province of %country. Please enter a valid state/province abbreviation for %field.', array('@state' => $element['#value'], '%country' => $countries[$country_id], '%field' => $element['#title'])));
             }
           }
+          elseif ($dt == 'Money') {
+            if (CRM_Utils_Rule::money($element['#value']) == FALSE) {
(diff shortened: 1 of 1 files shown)


colemanw:
This PR seems to be disabling *all* currency validation.
Probably it should do a regex check like `[1-0,.]`

jitendrapurohit:
Updated the PR to use regex instead of money() function. 

Tested strings that work fine - `10`, `10.24`, `22,000`, `23,000,000`. 

Strings that fail - alphanumeric chars and strings like `$10`, `12,11`, `1,,239`, etc.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2162,6 +2164,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         elseif ($field['type'] === 'number') {
           $sum = 0;
           foreach ((array) $val as $k => $v) {
+            //Strip any thousand separator for further calculation.
+            if ($field['data_type'] === 'Money') {
+              $v = preg_replace('/(?<=\d),(?=\d{3}\b)/', '', $v);
+            }
```
colemanw: Shouldn't we be respecting the thousands separator in crm config?

colemanw: @jitendrapurohit it's

    wf_civicrm_api('Setting', 'get', array(
      'return' => array("monetaryThousandSeparator"),
    ));

jitendrapurohit: Thanks for the comments @colemanw I think I overlooked the previous comment notification. I'll check this and update the PR.

babitawedsdeepak: updated the PR with the separator mentioned in the config.

jitendrapurohit: @colemanw updated the PR with the separator stored in the config.

mattwire:
Could we use CRM_Utils_Rule::money here instead as that function is well tested in CiviCRM core? https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/Rule.php#L577

mattwire:
@jitendrapurohit @colemanw What is the status of this PR - it's 2 years old now.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2178,6 +2180,13 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         elseif ($field['type'] === 'number') {
           $sum = 0;
           foreach ((array) $val as $k => $v) {
+            //Strip any thousand separator for further calculation.
+            $separator = wf_civicrm_api('Setting', 'getvalue', array(
+              'name' => "monetaryThousandSeparator",
+            ));
+            if ($field['data_type'] === 'Money') {
+              $v = preg_replace("/(?<=\d){$separator}(?=\d{3}\b)/", '', $v);
+            }
```
colemanw, line 2189: I haven't tried it out with alternate thousands separators but I'd imagine this will also require fixing the decimal place symbol.

PR #82, opened 2017-07-27
Fix nondeduct amount.

KarinG:


KarinG:
Tested on local clone of real client site:

If is deductible? Yes:
image

If is deductible? No:
image

KarinG:
@jake-mw - could you please give this a quick look?

ghost:
@KarinG This looks correct to me.

PR #85, opened 2017-09-11
Fetch recent data in post_save for custom field

nishant-bhorodia:
Hello @colemanw , 

This PR sorts the following issue: 
Editing the field within a filedset from civicrm end does not reflect back on the webform tab. When you edit same field again then the previous edit is shown on webform tab.

Thanks.


colemanw:
@nishant-bhorodia I've pushed a change to this PR to simplify the api call. Can you please re-test and verify it works as expected?

nishant-bhorodia:
Hello @colemanw , 
Thanks for this. I tested and it works as expected :)
Regards.

PR #86, opened 2017-09-23
Call doPayment if doTransferCheckout does not exist

eileenmcnaughton:
Probably in 4.6 but definitely by 4.7 doPayment was the preferred method to call to do payments of any sort. Some processors might not fully support this, although in 4.7 & possibly 4.6 the parent class will handle that appropriately.

Least risky way is to still support doTransferPayment if it exists.


colemanw:
@KarinG what do you think?

KarinG:
Ok - looking at this now

KarinG:
Yes - this is looks good - and I confirmed it by patching a 4.6 client (live test).

PR #87, opened 2017-10-05
Issue #2914054: Re-initialize webform fields when creating new contact

omarabuhussein:
The issue is reported here : https://www.drupal.org/node/2914054

Basically, If there is a contact that is set as "existing contact" with "address" fields locked (hidden) then the statue/province field (which is an Ajax select field) will appear as a normal text input. This behavior was done here : https://github.com/colemanw/webform_civicrm/commit/8bb0173c6ccb080d98bdfab0d8b1a6b7d838f421

**Before**
after1

But if the field is not a readonly field anymore which happen when you entering a details of a new contact inside (existing contact) field then the state/province should appear normally.

I fixed this by Re-initializing it after **pub.existingSelect** is called. 

**After**
before

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -31,9 +31,6 @@ var wfCivi = (function ($, D) {
           $(':input[name$="civicrm_'+num+'_contact_1_contact_'+i+'_name]"]', '.webform-client-form-'+nid).val(names[i]);
         }
       }
-
-      D.behaviors.webform_civicrmForm.attach();
-
       return;
     }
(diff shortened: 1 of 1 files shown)


jamienovick:
@colemanw this is a HW go live blocker for the release on Monday. Can we get this reviewed asap!!

colemanw:
Calling that function twice might have unintended consequenses. Maybe there's a better way..

omarabuhussein:
@colemanw  You are correct, it was just rather a rushed solution so I just pushed new commit , hopefully this is a better approach.

colemanw:
Looks good, thanks.

PR #88, opened 2017-10-19
Call validatePaymentInstrument instead of validateCreditCard

mattwire:
validateCreditCard should not be called directly, but should be called via the overrideable base class in CRM_Core_Payment_Form::validatePaymentInstrument.

This was breaking stripe via webform as we don't validate credit card fields for stripe (a token is passed instead).

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1614,7 +1614,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
     }
     // Validate credit card number & cvv2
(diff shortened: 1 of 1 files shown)


colemanw:
@eileenmcnaughton does this look right to you?

eileenmcnaughton:
Yes it does - although most of that function is redundant once you are doing that. I think you are at the point though where you should cut a release that does not officially support older versions on CiviCRM. I can speak to the fact that is correct with 4.7. It's been over a year since I looked at any earlier versions and you risk breaking compatibility with earlier versions as you adapt to 4.7. Removing formal support gets you off the hook for that (& people can use the current version of webform for those older versions so they have not missed out)

mattwire:
Closed in favour of #90 (no changes, just branch name)

PR #89, opened 2017-10-29
Fix PHP notices when elements are not defined

mattwire:
Not all financial types have a tax rate.  The array index is not defined if the specified financial type does not.
Also, fix a typo in "_contribution_page" variable name.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -60,7 +60,7 @@ abstract class wf_crm_webform_base {
           if ($this->contribution_page) {
             // tax integration
             $taxRates = CRM_Core_PseudoConstant::getTaxRates();
-            $this->_tax_rate = $taxRates[$this->_contribution_page['financial_type_id']];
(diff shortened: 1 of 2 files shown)


includes/wf_crm_webform_base.inc
```diff
@@ -57,10 +57,10 @@ abstract class wf_crm_webform_base {
       case 'tax_rate':
         $taxSettings = CRM_Core_BAO_Setting::getItem(CRM_Core_BAO_Setting::CONTRIBUTE_PREFERENCES_NAME, 'contribution_invoice_settings');
         if (is_array($taxSettings) && !empty($taxSettings['invoicing'])) {
-          if ($this->contribution_page) {
+          if ($this->_contribution_page) {
```
colemanw: That's not a typo. See comment above the variable.

mattwire: Ok, removed that bit.  This PHP notice fix should be good though

KarinG:
The fixes for the PHP notices look fine - thanks @mattwire;

PR #92, opened 2017-11-01
Add contact labels when registering multiple participants

jitendrapurohit:
Registering more than 1 participant through webform gives a list of event on the confirmation page. This PR adds information of which participant is chosen for a specific event on confirmation as well as line items shown on view contribution page.

image

image

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -382,6 +382,14 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       else {
         continue;
       }
+      $participants = current(wf_crm_aval($this->data['contact'], "$c:contact", array()));
+      $firstName = CRM_Utils_Array::value('first_name', $participants);
+      $lastName = CRM_Utils_Array::value('last_name', $participants);
+
+      $pName = "{$firstName} {$lastName}";
+      if (empty(trim($pName))) {
+        $pName = CRM_Utils_Array::value('webform_label', $participants);
+      }
       foreach (wf_crm_aval($par, 'participant', array()) as $n => $p) {
         foreach (array_filter(wf_crm_aval($p, 'event_id', array())) as $id_and_type) {
           list($eid) = explode('-', $id_and_type);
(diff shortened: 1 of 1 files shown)


KarinG:
Hi @jitendrapurohit - Pete showed me this a few weeks back - thank you for PRing it; is your client on 4.7 or 4.6? Coleman and I decided not to branch webform_civicrm module - so we're looking to make sure we continue to cover both;

KarinG:
Doing some live testing on a 4.6 now;

KarinG:
Before:
image

Participant Registrations: 
image

Contribution Record: 
image

KarinG:
After: note: I'm getting the Fieldset labels -> not the Swimmer (which is an Existing Contact - Static - Relationship to Parent/Guardian); 

image

image

KarinG:
After: CiviCRM Participant Registration (same as Before);
Contribution: not getting the lineItems:

image

KarinG:
To summarize:

- great idea/improvement
- does not yet work for Existing Contacts though (on 4.6)
- don't get the lineItems (on 4.6)

Something related that has always bugged me is - when you view a single Participant record - it looks like there has been an overpayment:

image

jitendrapurohit:
Hi @KarinG Thanks for the testing. Almost all of our clients are on 4.7, so I made this PR while working on the latest version of CiviCRM. I'll check with 4.6 and see if I can make this PR work on both the versions.

KarinG:
Hi @jitendrapurohit - if it at least does not break anything in 4.6 [I did not see any fatal errors; perhaps you can double check/check for notices] - then I'm ok to merge this in. Our main goal is to not break anything in 4.6 [without fair warning to people]; But no need to support all new functionality in 4.6

jitendrapurohit:
@KarinG I had a look at this and have updated the PR with the existing contact fix. The line items not shown on 4.6 is the default behaviour and doesn't seem to be related to this changes.

KarinG:
Hi @jitendrapurohit - sorry for the delay on this. I've now upgraded my CiviCRM test site to 4.7 - and am applying your PR - getting the following error - which I'm pretty sure is unrelated to your PR - but it's very annoying. Can you still pay with credit cards via webform civicrm w/ 4.7.27?

image

KarinG:
I don't think new functionality needs to work on 4.6 - we just need to be cautious not to break anything - and I've confirmed this PR does not. So I've merged it.

PR #93, opened 2017-11-01
CGP-904: Implementing 'Create mode' to Contact's custom fieldsets - WIP

coldrunKacper:
Implementation of 'Create mode' select field for any Custom Group of Contact entity.

'Create mode' select field appears at each Custom Group fieldset of Contact entity (in CIVICRM tab of Webform configuration).
cgp-904

There are two available values of 'Create mode' field:

0 - "Create/ Edit" - default mode, works with original behavior.
1 - "Create only" - with this mode, when editing a webform submission, the civicrm custom values of the custom group are added (created) as new values set instead of being updated. It only affects civicrm custom values. Drupal webform's values are still updated as originally.
It allows to create infinite sets of civicrm custom group's values using webform submission.

Sample flow using new create modes:

1. We have a custom group with two original values in civicrm:
cgp-904-step1-civicrm-custom-group

2. Here is how it looks on our sample webform: (the civicrm custom values are prepopulated)
cgp-904-step2-open-webform

3. Here is our webform configuration: (including both 'create mode' values)
cgp-904-step3-webform-configuration

4. We open the webform and fill with the values below:
cgp-904-step4-save-webform

5. Here is how contact's custom values look like after saving the webform:
cgp-904-step5-civicrm-custom-group-after-webform-submit

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -324,8 +324,18 @@ class wf_crm_admin_form {
           }
         }
 
-        // Add 'Create mode' select field to custom group fieldset.
+        $isMultipleCustomGroup = FALSE;
+
         if (substr($sid, 0, 2) == 'cg') {
+          $customGroupId = (int) substr($sid, 2);
+          $customGroup = $this->getCustomGroup($customGroupId);
+          if (!empty($customGroup['is_multiple']) && (int) $customGroup['is_multiple']) {
+            $isMultipleCustomGroup = TRUE;
+          }
+        }
+
+        // Add 'Create mode' select field to multiple custom group fieldset.
+        if ($isMultipleCustomGroup) {
           $createModeKey = 'civicrm_' . $n . '_contact_' . $i . '_' . $sid . '_createmode';
 
           $createModeValue = isset($this->settings['data']['config']['create_mode'][$createModeKey]) ? $this->settings['data']['config']['create_mode'][$createModeKey] : NULL;
@@ -374,6 +384,17 @@ class wf_crm_admin_form {
         $this->help($rule_field, 'matching_rule');
       }
     }
+  }
+
+  /**
+   * Returns an array containing Custom Group properties.
+   *
+   * @param int $id
+   *
+   * @return array
+   */
+  private function getCustomGroup($id) {
+    return civicrm_api3('CustomGroup', 'getsingle', array('id' => $id));
   }
 
   /**
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -350,6 +350,7 @@ class wf_crm_admin_form {
               wf_crm_webform_base::MULTIVALUE_FIELDSET_MODE_CREATE_ONLY => t('Create Only')
             ),
             '#title' => t('Create mode'),
+            '#weight' => -1,
           );
           $this->help($multivalueFieldsetCreateMode, 'multivalue_fieldset_create_mode', t('Multivalue fieldset create mode'));
           $fieldset[$createModeKey] = $multivalueFieldsetCreateMode;
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -5,8 +5,6 @@
  * Webform CiviCRM module's admin form.
  */
 
-module_load_include('inc', 'webform_civicrm', 'includes/wf_crm_webform_base');
-
 class wf_crm_admin_form {
   private $form;
   private $form_state;
@@ -23,6 +21,7 @@ class wf_crm_admin_form {
   function __construct($form, &$form_state, $node) {
     module_load_include('inc', 'webform_civicrm', 'includes/utils');
     module_load_include('inc', 'webform_civicrm', 'includes/wf_crm_admin_help');
+    module_load_include('inc', 'webform_civicrm', 'includes/wf_crm_webform_base');
     civicrm_initialize();
     $this->form = $form;
     $this->form_state = &$form_state;
@@ -324,18 +323,8 @@ class wf_crm_admin_form {
           }
         }
 
-        $isMultipleCustomGroup = FALSE;
-
-        if (substr($sid, 0, 2) == 'cg') {
-          $customGroupId = (int) substr($sid, 2);
-          $customGroup = $this->getCustomGroup($customGroupId);
-          if (!empty($customGroup['is_multiple']) && (int) $customGroup['is_multiple']) {
-            $isMultipleCustomGroup = TRUE;
-          }
-        }
-
         // Add 'Create mode' select field to multiple custom group fieldset.
-        if ($isMultipleCustomGroup) {
+        if (substr($sid, 0, 2) == 'cg' && wf_crm_aval($set, 'max_instances') > 1) {
           $createModeKey = 'civicrm_' . $n . '_contact_' . $i . '_' . $sid . '_createmode';
 
           $createModeValue = isset($this->settings['data']['config']['create_mode'][$createModeKey]) ? $this->settings['data']['config']['create_mode'][$createModeKey] : NULL;
@@ -385,17 +374,6 @@ class wf_crm_admin_form {
         $this->help($rule_field, 'matching_rule');
       }
     }
-  }
-
-  /**
-   * Returns an array containing Custom Group properties.
-   *
-   * @param int $id
-   *
-   * @return array
-   */
-  private function getCustomGroup($id) {
-    return civicrm_api3('CustomGroup', 'getsingle', array('id' => $id));
   }
 
   /**
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1636,7 +1636,8 @@ class wf_crm_admin_form {
             $component['value'] = $val;
             webform_component_update($component);
           }
-          elseif (substr($key, -11) === '_createmode') {
+
+          if (substr($key, -11) === '_createmode') {
             // Update webform's settings with 'Create mode' value for custom group.
             $this->settings['data']['config']['create_mode'][$key] = $val;
           }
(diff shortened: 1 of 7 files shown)


guanhuan:
@coldrunKacper  could you move the "create mode" field to around where the "add dynamically" field is? screen shot 2017-11-02 at 11 49 59

The PR description seems to be focusing submission edit behaviour but the key is how the webform fields will work on the actual webform when specific create mode is selected. Can you clarify?

coldrunKacper:
@guanhuan yes that's how it works, please see updated PR description.

colemanw:
@coldrunKacper how would this work when editing a webform submission? I'm guessing it would incorrectly create the values a 2nd time unless you've added a safeguard against this.

guanhuan:
@colemanw 
I feel this is the only way it could work since the other two options would be:

1. Don't create new records when editing submission
2. Update existing records when editing submission

Option 1 is doable but doesn't seem more correct than current behaviour (create new records).

Option 2 is not correct since they are "create only" fields and even if you want to map it to existing records, I don't think it can be mapped to the ones that are created via this submission.

Also the users who submitted the form will not be able to edit the submissions themselves. If the admin wants to update the records, I think it makes more sense to do it in CiviCRM.

Above is my consideration when we were discussing this internally but please let me know if there is a approach that covers the business logic better.

coldrunKacper:
I've set css styles for 'Create mode' select to display it in the same way as 'Add dynamically' checkbox:
cgp-904-create-mode-positioning

colemanw:
@guanhuan do you have this code in production for your client(s)?

guanhuan:
@colemanw not yet, for patches likes this we always target to get author's approval first.

colemanw:
A couple more questions that came up when discussing this one with @KarinG:

- is this on CiviCRM 4.7 and/or 4.6? Have you tested it for both?
- "users who submitted the form will not be able to edit the submissions themselves" - that's handled at a permission level though - how do you prevent site admins from giving people Edit own webform submissions - permission?

guanhuan:
Hi @colemanw ,

- is this on CiviCRM 4.7 and/or 4.6? Have you tested it for both?

We can test it on the 4.7 and 4.6 and let you know.

- "users who submitted the form will not be able to edit the submissions themselves" - that's handled at a permission level though - how do you prevent site admins from giving people Edit own webform submissions - permission?

Sorry, I was mistaken. Yes, they can be given the permission to edit the submission but as per breakdown here, I think it makes the most of sense to create new records when this occurs if the fields are set to create-only.

madurak:
Tested on Civi 4.6.33 + latest webform_civicrm from git + patch 93
Tested on Civi 4.7.27 + latest webform_civicrm from git + patch 93
No issues found on 'create/edit' and 'create only' modes

includes/wf_crm_admin_help.inc
```diff
@@ -425,6 +425,14 @@ class wf_crm_admin_help {
       '</p>';
   }
 
+  public static function multivalue_fieldset_create_mode() {
+      print '<p>' .
+        t('Create/ Edit Mode (Default): Pre-populate the existing entry (if any) in order to allow user modifying it. If there isn\'t any existing entry, any value entered into the fieldset will be created as a new entry.') .
```
colemanw: The convention with `t()` is that you don\\'t escape quotes when possible. Putting this string in double quotes would avoid the need to escape the single quote.

colemanw:
When defining a custom set in CiviCRM you can specify a maximum number of records. Does this code respect that limit?

KarinG:
I was wondering if you could describe a use case for this. I like details!

guanhuan:
@colemanw @KarinG 
The use case to this is to look at the multiple submission aspect of multi-submission custom fieldset rather than just the fact that multi-submission custom fieldset can hold more than one record.

In the current Webform CiviCRM, a user can only submit a number of records for multi-submission custom fieldset and the number is decided by how many set of fields are added to the form. When the user come back and submit the same form again, the previously submitted records will have to overwritten by the new records and even if the user left a few sets empty, empty values will overwrite the existing value in the corresponding slot. There is no way to prevent that.

Assuming you are collecting the user's measurements, the user will submit a set of measurements every three month to reflect their exercise progress but not submitting 4 sets of measurements at the same time.

And in response to question to "maximum number of records", the "Create-only" mode fields will not respect that since there is no good way to reflect the limit in the form currently. If there are limits to how many records user can submit, for example 5, then use "Create/Edit" mode (current behaviour) will achieve that but still not perfect since user is not notified that their submission will overwrite the old values.

KarinG:
@guanhuan - ok so could this be used for e.g. [detailed potential use case - to try understand this better]:

Professional members of an association need to report their Education (Continuing Education hours every year when they renew their Membership. Hours are stored in a dataset that is configured to "allow multiple records"

image

 a second field in this set is the Year -> so the result multiple entries of Year/Hours:

image

image

With your PR - can this data be pulled into a webform -> 
showing 
2014 - 100h
2015 - 300h

and can the user than add:

2016 - 350h 
2017 - 200h

resubmit - and next time they go back see this data?

If so (or if not) - could you show some screenshots of your client's use case?

guanhuan:
@KarinG This is related to a CiviHR feature but the example in CiviHR is not as straight forward since this change provides a workaround for the CiviHR feature but not an actual use case for it.

The "create only mode" represents how multi submission fields work in CiviCRM core. As I mentioned:

> The use case to this is to look at the multiple submission aspect of multi-submission custom fieldset rather than just the fact that multi-submission custom fieldset can hold more than one record.

In CiviCRM core, you can add multi-submission custom fields into a profile. The profile can only include one set of the fields rather than multiple sets per webform civicrm behaviour. When user submit the profile form, a new record will always be created in the multi-submission custom fields in that contact record. The use case of allowing user to edit previous submission is very useful but at the moment exclusive to webform as far as I know.

Using an example: an organisation wants to collect the feedback from visitors every time they visited a exhibition centre (very common scenario), they decide to do this electronically via custom fields instead of handing them a pen and a form. See custom field configuration below:
screen shot 2017-11-23 at 10 52 42
screen shot 2017-11-23 at 10 52 49

Now every time I visit the centre, I will be filling out a profile form below. I get to fill out my experience for this visit.

screen shot 2017-11-23 at 10 37 25

After 3 visits, my submissions look like this.
screen shot 2017-11-23 at 10 37 38

Now without "create only mode" this is currently not possible via Webform CiviCRM but it is possible and also the only way it works in CiviCRM core.

includes/wf_crm_admin_form.inc
```diff
@@ -321,6 +323,39 @@ class wf_crm_admin_form {
...
+        if (substr($sid, 0, 2) == 'cg') {
+          $customGroupId = (int) substr($sid, 2);
+          $customGroup = $this->getCustomGroup($customGroupId);
+          if (!empty($customGroup['is_multiple']) && (int) $customGroup['is_multiple']) {
+            $isMultipleCustomGroup = TRUE;
+          }
+        }
+
+        // Add 'Create mode' select field to multiple custom group fieldset.
+        if ($isMultipleCustomGroup) {
```
colemanw: None of the above code defining `$isMultipleCustomGroup` is necessary. It would be sufficient to change this if statement to read `if (substr($sid, 0, 2) == 'cg' && wf_crm_aval($set, 'max_instances') > 1) {`

colemanw:
I think we can merge this once the nitpicks from my code review have been addressed.

includes/wf_crm_admin_form.inc
```diff
@@ -352,6 +387,17 @@ class wf_crm_admin_form {
...
+   * Returns an array containing Custom Group properties.
+   *
+   * @param int $id
+   *
+   * @return array
+   */
+  private function getCustomGroup($id) {
+    return civicrm_api3('CustomGroup', 'getsingle', array('id' => $id));
+  }
+
```
colemanw: This function is not necessary.

includes/wf_crm_admin_form.inc
```diff
@@ -5,6 +5,8 @@
  * Webform CiviCRM module's admin form.
  */
 
+module_load_include('inc', 'webform_civicrm', 'includes/wf_crm_webform_base');
```
colemanw: I don't think this is needed.

mickadoo: Seems to be required for loading of the class constants. Without it I get an error 

```
 Fatal error: Class 'wf_crm_webform_base' not found in /repos/webform_civicrm/includes/wf_crm_admin_form.inc on line 339
```

It probably fits better in the constructor though with the other `module_load_include` statements so I'll move it there.

mickadoo:
@colemanw I think @coldrunKacper has moved on to something else so I've been assigned the task of finishing this PR.

> When defining a custom set in CiviCRM you can specify a maximum number of records. Does this code respect that limit?

From what I've seen it does not. I had a look at how this could be done. I thought about disabling the fields in the `\wf_crm_webform_preprocess::fillForm` method _if_ the provided contact already has reached the submission limit for this custom group fieldset. 

However for webforms where the contact is chosen by the user this wouldn't work. The default contact might have reached the submission limit, but a different contact might not have. 

We could do it in a validation rule in `\wf_crm_webform_postprocess::validate` either, but then the user would have to wait until submitting the form to realize there was a problem. Even in validation I'm not sure how to deal with the validation failure. Should we disable the fields for the custom group that's reached its limit? If the contact ID was changed again by the user then this input could become valid. We could leave it as it is and require the user to manually remove all input from the fields, although that doesn't seem optimal either.

For me it seems like the ideal solution would be to use Javascript to check when the contact ID changes which custom group fieldsets have reached their limit, and disable the fields for them, but this might take some time (especially since I'm more of a backend dev and JS world is confusing to me)

Do you have any suggestions for the best way to deal with this?

colemanw:
Thanks for looking into that question @mickadoo. I think for now let's leave that as a TODO, since it's a thorny problem and webform admins will just have to know better than to use this feature in conjunction with a limited field.

mickadoo:
@colemanw sure. I pushed two commits addressing your review changes and another that fixes a bug with saving the creation mode for new webforms. Whenever you have time can you have a look?

colemanw:
Looks good.

PR #95, opened 2017-11-02
WIP Make getFileInfo work when custom file field is a membership field

mattwire:
I made this change some time ago whilst configuring a webform that uploads a file to a civicrm membership custom field.

It didn't work without this patch, but further testing is required before it should be merged!


KarinG:
Hi @mattwire - ok let's further test then! Can you please test this on 4.7: document/screenshot all your steps here - and then I'll test it on 4.6;

mattwire:
I'm not able to reproduce the original issue

PR #97, opened 2017-11-10
allow disabled field value submission

mountev:
This fix ensures disabled fields are included when the activity details summary is constructed.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -34,6 +34,7 @@ function _webform_defaults_civicrm_contact() {
       'hide_fields' => array(),
       'hide_method' => 'hide',
       'no_hide_blank' => FALSE,
+      'submit_disabled' => FALSE,
       'attributes' => array(),
       'private' => FALSE,
       'default' => '',
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -181,7 +181,7 @@ function _webform_edit_civicrm_contact($component) {
   $form['field_handling']['submit_disabled'] = array(
     '#type' => 'checkbox',
     '#title' => t("Submit disabled field value(s)"),
-    '#description' => t('Store disabled field values in webform submissions.'),
+    '#description' => t('Store disabled field value(s) in webform submissions.'),
     '#default_value' => $component['extra']['submit_disabled'],
     '#parents' => array('extra', 'submit_disabled'),
   );
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -187,6 +187,15 @@ function _webform_edit_civicrm_contact($component) {
     '#default_value' => $component['extra']['unique'],
     '#parents' => array('extra', 'unique'),
   );
+  if ($c == 2) {
+    $form['field_handling']['submit_disabled'] = array(
+      '#type' => 'checkbox',
+      '#title' => t("Submit disabled field value(s)"),
+      '#description' => t('Store disabled field values in webform submissions.'),
+      '#default_value' => $component['extra']['submit_disabled'],
+      '#parents' => array('extra', 'submit_disabled'),
+    );
+  }
   $form['extra']['allow_create'] = array(
     '#type' => 'hidden',
     '#value' => $allow_create,
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -178,6 +178,13 @@ function _webform_edit_civicrm_contact($component) {
     '#default_value' => $component['extra']['no_hide_blank'],
     '#parents' => array('extra', 'no_hide_blank'),
   );
+  $form['field_handling']['submit_disabled'] = array(
+    '#type' => 'checkbox',
+    '#title' => t("Submit disabled field value(s)"),
+    '#description' => t('Store disabled field values in webform submissions.'),
+    '#default_value' => $component['extra']['submit_disabled'],
+    '#parents' => array('extra', 'submit_disabled'),
+  );
   $form['validation']['unique'] = array(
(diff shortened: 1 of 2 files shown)


lcdservices:
Reviewed and tested. This works as expected:

- data for disabled fields is stored in webform submission
- data is captured and stored in the activity details field (if that option is enabled)

colemanw:
Sorry I really don't understand the significance of contact 2. A webform can have up to 50 contacts. So are you saying that 49 of them work correctly but one of them mysteriously doesn't?

deepak-srivastava:
Thanks for feedback.  Tried to dig out facts a bit more. 

It was only happening when contact2 picks up existing contact, while other contacts (say contact3) creates a new one. In this case, new contact fields are not locked and therefore submitted. So it made it look like the problem was only happening for contact2. But this issue is with all the contacts if all of them have disabled fields.

We have updated the PR to generalize the fix.

mountev:
Thanks for the feedback.  Tried to dig out facts a bit more. 

It was only happening when contact2 picks up existing contact, while other contacts (say contact3) creates a new one. In this case, new contact fields are not locked and therefore submitted. So it made it look like the problem was only with contact2. But this issue is with all the contacts if all of them have disabled fields.

We have updated the PR to generalize the fix.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2131,7 +2131,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           continue;
         }
         // Ignore values from fields hidden by existing contact component
-        if ($this->isFieldHiddenByExistingContactSettings($ent, $c, $table, $n, $name)) {
```
KarinG: Update the comment to something like "Check to see if configured to ignore hidden field values"

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2225,16 +2225,22 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+  private function isFieldHiddenByExistingContactSettings($ent, $c, $table, $n, $name, $checkSubmitDisabledSetting = FALSE) {
     if ($ent == 'contact' && isset($this->enabled["civicrm_{$c}_contact_1_contact_existing"])) {
       $component = $this->getComponent("civicrm_{$c}_contact_1_contact_existing");
       $existing_contact_val = $this->submissionValue($component['cid']);
       // Fields are hidden if value is empty (no selection) or a numeric contact id
       if (!$existing_contact_val[0] || is_numeric($existing_contact_val[0])) {
         $type = ($table == 'contact' && strpos($name, 'name')) ? 'name' : $table;
         if (in_array($type, $component['extra']['hide_fields'])) {
+          // Check if 'Submit Disabled' setting is switched on and do it for contact_2 only as the data for contact 2 doesn't get to the webform submissions
+          if ($checkSubmitDisabledSetting && $component['extra']['submit_disabled']) {
```
KarinG: Update comment to remove references to contact 2 -> to something like: "Check to see if configured to Submit disabled field value(s)"

includes/contact_component.inc
```diff
@@ -178,6 +178,13 @@ function _webform_edit_civicrm_contact($component) {
     '#default_value' => $component['extra']['no_hide_blank'],
     '#parents' => array('extra', 'no_hide_blank'),
   );
+  $form['field_handling']['submit_disabled'] = array(
+    '#type' => 'checkbox',
+    '#title' => t("Submit disabled field value(s)"),
+    '#description' => t('Store disabled field values in webform submissions.'),
+    '#default_value' => $component['extra']['submit_disabled'],
```
KarinG: Let's be consistent for description: Submit disabled field value(s) ...

mountev:
@KarinG - Thanks, have updated the comments.

colemanw:
I'm still confused about this issue. How are these fields being disabled? Can you please give details of the use case and steps to reproduce the bug so we can verify it is indeed a bug in this module?

lcdservices:
@colemanw when implementing the "Existing Contact" option in the CiviCRM webform settings for a contact, you have the option of locking fields and displaying them as disabled. This is useful if you want the user to be able to view details for a contact but not edit the details (you can optionally choose to not lock fields that are empty -- so the user can fill in missing data but not change existing data). However, when set, those disabled fields are not stored with the webform submission and if you are using the activity tool to store the webform summary in the activity details field, any disabled fields are left empty. This option now includes those disabled fields in the submission so they are in the webform data and activity data.

See the attached -- I've highlighted the two existing fields and third field that this PR adds.

image

lcdservices:
@colemanw any other questions about this PR? it's an important improvement so I want to make sure it gets merged.

includes/contact_component.inc
```diff
@@ -178,6 +178,13 @@ function _webform_edit_civicrm_contact($component) {
     '#default_value' => $component['extra']['no_hide_blank'],
     '#parents' => array('extra', 'no_hide_blank'),
   );
+  $form['field_handling']['submit_disabled'] = array(
+    '#type' => 'checkbox',
+    '#title' => t("Submit disabled field value(s)"),
+    '#description' => t('Store disabled field value(s) in webform submissions.'),
+    '#default_value' => $component['extra']['submit_disabled'],
```
colemanw, line 186: This will produce an undefined index - you need to add the default for this field to the function `_webform_defaults_civicrm_contact()`

mountev:
default added.

colemanw:
This looks good, thanks.

PR #99, opened 2017-11-30
Make webform_autocomplete work for civi custom fields

jitendrapurohit:
Fixes https://www.drupal.org/project/webform_autocomplete/issues/2796209


colemanw:
I reviewed this and found some issues

1. Access control wasn't being enforced
2. IO wasn't handled properly - field label was being stored instead of key which did not work correctly

I fixed those and made some additional improvements

1. Automatically create autocomplete components
2. Alter autocomplete component edit form to remove irrelevant options when used by CiviCRM
3. Respect max results limit
4. Use wf_crm_api wrapper instead of civicrm api & bao functions

jitendrapurohit:
Thanks for the additional changes @colemanw 👍 . I've tested the updated PR and it works fine.

PR #101, opened 2017-12-26
Fix remaining space in event select option list

jitendrapurohit:
Fixes - https://www.drupal.org/project/webform_civicrm/issues/2926383


colemanw:
Looking at the git history, this was a mistake. Thanks for fixing it.

PR #102, opened 2017-12-26
Don't reset date field if default is set for the component

jitendrapurohit:
Fixes - https://www.drupal.org/project/webform_civicrm/issues/2933003

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -10,7 +10,7 @@ var wfCivi = (function ($, D) {
    */
   var pub = {};
 
-  pub.existingSelect = function (num, nid, path, toHide, hideOrDisable, showEmpty, cid, fetch) {
+  pub.existingSelect = function (num, nid, path, toHide, hideOrDisable, showEmpty, cid, fetch, defaults) {
     if (cid.charAt(0) === '-') {
       resetFields(num, nid, true, 'show', toHide, hideOrDisable, showEmpty, 500);
       // Fill name fields with name typed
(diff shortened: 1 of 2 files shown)


colemanw:
It looks like you are skipping the field instead of restoring it to the set default. Wouldn't it be better to set it to the default value?

jitendrapurohit:
The value is already set when the webform is loaded, we just needed to avoid it to reset from the contact selection. 

Moreover, the default value of the date field is meant to be any string present in [GNU input date formats](http://www.gnu.org/software/tar/manual/html_chapter/Date-input-formats.html) like `today`, `+2 months`, `Dec 9 2004 `, etc. I don't see any modification that can be done here which could set the value corresponding to all these strings? Any pointers which can do the same? Also, it doesn't seem necessary to override "already existing" value with the "same value" on the webform?

I've tested this patch on all webforms using this use-case on our sites and also on the sample file attached to the issue link. They all work as expected.

colemanw:
I think the way to pass the default to javascript would be to run it through `strtotime()` in php before passing it to the client side.

jitendrapurohit:
updated the code to set the default value from component instead of skipping the field. Thanks @colemanw

js/webform_civicrm_forms.js
```diff
@@ -131,7 +131,21 @@ var wfCivi = (function ($, D) {
...
               if (this.id && $(this).val() != '') {
-                $(this).val('');
+                //Set default value if it is specified in component settings.
+                if (typeof defaults != "undefined" && defaults.hasOwnProperty(name)) {
+                  if ($el.hasClass('webform-component-date')) {
+                    var date = defaults[name].split('-');
+                    $el.find('select.year, input.year').val(+date[0]);
+                    $el.find('select.month').val(+date[1]);
+                    $el.find('select.day').val(+date[2]);
+                  }
```
colemanw: I don't understand how this would work. We're in a loop of all `$(':input', this).not(':radio, :checkbox, :button, :submit, :file, .form-file')` but acting on a wrapper div around other form elements.

jitendrapurohit: Moved `$el` update to little above so that it is modified in the div loop initiated in the `resetfields()` function rather than inside `$(':input', this)`. 

Confirmed the updated changes and it too works fine on my testing.

PR #104, opened 2018-01-21
issue 2938155

vakeesan26:
This changes will redirect to Pay-pal site.
Token & Payer ID being returned. 

at the moment it is being redirected to node/{nid}/done?token=xx&PayerID=yyy
to complete the payment doExpressCheckout method should be called via doPayment method .

CiviCRM having confirmation page on contribution pages. therefor when make payment button being pressed they are calling doPayment method with token & payerid 

in same way some how doPayment method should be called in webform after the SetExpressCheckout redirection.

https://www.drupal.org/project/webform_civicrm/issues/2938155

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1848,7 +1848,28 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params['webform_redirect_success'] = $this->getIpnRedirectUrl('success');
 
     if (method_exists($paymentProcessor, 'doTransferCheckout')) {
-      // doTransferCheckout is deprecated but some processors might still implement it.
+      /* Paypal Express checkout */
+	  if(wf_crm_aval($_POST, 'credit_card_number') == 'express'){
+		try {
+		  $params['button'] = 'express';
+		  $params['component'] = 'contribute';
+		  $result = $paymentProcessor->doPreApproval($params);
+		  if (empty($result)) {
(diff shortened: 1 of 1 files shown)


colemanw:
@KarinG does this look good to you?

vakeesan26:
any update on this issue...

colemanw:
I don't have a paypal site to test this on and neither does Karin. It would be good to find someone who can try it out and give feedback.

Meanwhile there are a few code style issues that need a little fixing up. Please use 2 spaces instead of tabs, and two slashes for a comment instead of slash-star, and keep the whitespace consistent with the rest of the code (one space after a control statement, one space before an open bracket, etc).

Thanks for contributing!

jdkoelsch:
Hi,

I ran into the same issue as the submitter, and this patch fixed the issue for me as well. With the updated code formatting, is it in a position to be accepted now? 

I have tested it using the the current version of the module using Paypal Pro as the payment processor. Please let me know if I can provide more info that would help get this PR accepted.

colemanw:
Ok @jdkoelsch I'll go ahead and merge this based on your testing.
But I note that this patch is adding code to a spot that is deprecated. @eileenmcnaughton might have some insight into what will be replacing it.

mattwire:
The `doTransferCheckout()` and `doDirectPayment()` methods are both deprecated in favour of implementing `doPayment()` directly.  So this PR will continue to work as long as the paypal code implements `doTransferCheckout()`.

PR #105, opened 2018-01-29
Calculate sales tax for each lineitem based on financial_type_id

mattwire:
Previously, webform_civicrm was using the contribution page financial_type_id to retrieve the tax rate from CiviCRM.  However, this may not be the same as the tax rate for the line item (if you have multiple lineitems) or if you have inserted lineitems via hooks / custom code.

This PR changes they way webform_civicrm handles tax rates for lineitems so it retrieves each tax rate from CiviCRM based on the financial_type_id for that lineitem.  If not set, it will fall back to the original method of using the contribution page financial_type_id.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -252,7 +252,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    * Send receipt
    */
   private function sendReceipt(){
-    // tax integration
+    // Sales Tax integration
     if (!is_null($this->tax_rate)) {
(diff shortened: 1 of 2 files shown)


KarinG:
Thanks for this Matt - tested this on 4.7 [and since you have a backup method -> continue to use the Financial Type as defined on the contribution page] - this should be safe for 4.6

- contribution page has "Donation" as Financial Type
- membership w/ tax has Financial Type "Taxable Item" 5% sales tax associated with it
- math looks much better now - thanks!

This is the before/after:

image

and in CiviCRM:
image

PR #109, opened 2018-02-17
(WIP) Payment Plan Work

omarabuhussein:
This PR contain all the CiviCRM Webform-related **Payment Plan** work we have done so far which is : 

## Membership Pay Later with Payment Plan in Installments

to explain this let us assume the following webform configurations: 

1. Create a webform and Enable CiviCRM Processing on the CiviCRM tab section
2. Enable First Name, Last Name and Email for the Contact 1
3. Enable one membership for Contact 1.
4. Enable Membership Fee on the membership tab
5. Select a Contribution Page that has Pay Later option on the Contribution tab
6. Select "month" in "Frequency of Installments". then Enable number of installments and Interval of Installments

Now suppose there is  now a user submitted this webform to pay for a membership that worth £120 - 1 year standard membership in 12 monthly installments using "pay later" option.

Before this work this is how it would behave : 

But this is not ideal, and this is how it is supposed to work instead : 

1.  A recurring contribution should be created with 1/12 of the membership fee per installment for 12 installments.

2. All 12 contributions should be generated upon submission at once. All contributions will be in status "Pending" until being marked as "Complete" manually.

3. The linked membership will have full one year duration and will be activated upon the completion of the first contribution.

On the contact record:
4. After the purchase is completed, one recurring contribution with 1/12 of total amount should be created. It should have 12 instalments.

5. 12 contributions with 1/12 of total amount should also be created with status of "pending". All the the contributions are a part of the "recurring" (see the "(recurring)" text on the contribution) and linked to the recurring contribution in point 4.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -898,4 +898,32 @@ abstract class wf_crm_webform_base {
     return CRM_Core_Key::get('CRM_Contribute_Controller_Contribution', TRUE);
   }
 
+  /**
+   * Determines if the target payment processor
+   * is pay later payment processor or not.
+   *
+   * @param int $paymentProcessorsID
+   *   The target payment processor ID
+   *
+   * @return bool
+   *   TRUE if the target payment processor is
+   *   pay later or if no payment processor is
+   *   selected, Return FALSE otherwise.
+   */
+  protected function isPayLaterPaymentProcessor($paymentProcessorsID) {
+    $payLaterPaymentProcessors = array();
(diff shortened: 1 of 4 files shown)


KarinG:
We're happy to work with you to make incremental steps/improvements in how webform_civicrm interacts with CiviCRM - but this is way too big [especially for a financial PR] and as such it can not be reviewed. Our guidelines: each PR must have a clear/specific before/after purpose; each PR must be tested on actual 4.6 / 4.7 sites;

PR #110, opened 2018-02-18
Also show line_item for participant if registered for a free event.

KarinG:
I've had this in production/live for CiviCamp Calgary since back in January

image


colemanw:
Does it work if *all* the selected events are free? Expected behavior would be to hide the payment processor fields in that case, and not fail validation for lack of payment.

KarinG:
Yes that works -> your registration for CiviCamp Calgary was an example of that. You had two $0 lineItems - and registered successfully for both CiviCamp day ($49 -> discounted to $0 and Sprint $0);

KarinG:
PS - poke Tim to register for CiviCamp Calgary -> and he can report back to us confirming the $0 lineItems worked just fine :-)

totten:
I was able to register for CiviCamp Calgary with a code -- both line items were processed as `$0`, and no billing blocks were required.

KarinG:
Awesome!

KarinG:
For completeness: registrations processed properly: 
image

PR #111, opened 2018-02-19
Improving"pay later" with installments prcoessing

omarabuhussein:
## Before

Currently when enabling installments in the webform contribution configurations : 

2018-02-19 22_41_33-nvidia geforce overlay dt

There will be either one of two cases : 

1- Using specific payment processor : 

In this case, webform_civicrm module will create a single '**completed**' contribution with amount that represent part of the total payment, and a recurring contribution record will be also created . If there was a membership configured, it will also be created and be in "**New**" status. Each time another installment time comes, the payment processor (if it supports recurring contribution) will trigger an IPN request to a civicrm callback URL and another contribution will be created until all the installments (the total amount) is paid. so as we can see webform_civicrm handle this case properly.

2- Using 'Pay Later' Option : 

In this case, webform_civicrm creates only a single contribution with the total amount in '**pending**' status, no recurring contribution record will be created and if there is a membership configured then it will be created in '**pending**' status. So basically webform_civicrm module does not handle it properly.

## Example

To Explain the above with an example, let us assume the following webform  configurations : 

1. Create a webform and Enable CiviCRM Processing on the CiviCRM tab section
2. Enable First Name, Last Name and Email for the Contact 1
3. Enable one membership for Contact 1.
4. Enable Membership Fee on the membership tab
5. Select a Contribution Page that has Pay Later option on the Contribution tab
6. Select "month" in "Frequency of Installments". then Enable number of installments and Interval of Installments

Now suppose there is now a user submitted this webform to pay for a membership that worth $120 for a  1 year membership in 12 monthly installments using "pay later" option, this is what will happen : 

1- A single contribution with amount  = 120$ in pending status will be created.
2- No recurring contribution will be created.
3- A membership in pending  statues will created.

before_pp1

## After

Since there is no IPN in case of "pay later" option, all the installment contributions should be created in advance in 'pending' status (instead of single one in pending contribution that contain the total amount) and a recurring contribution record should also be created and all the installment contributions should link to it. The admin then can manually change each installment contribution status to paid if it get paid by the user since as we said there is no IPN in this case to do that.

In the example above, 12 contributions should be created, each with amount = 10$, and a recurring contribution record will be created.

after_pp_1

## Notes

- There are notices appear after submitting the webform (it appear even before my changes), I did a check on them and fixed one of them which is caused by **sendconfirmation** api call by ensuring the correct payment processor get passed to it. The other notices require many modification to civicrm core to get rid of them since they happen due to wrong assumptions + code smells on civicrm side but they should not affect the webform module for now.
 

- Tested on both civi 4.7.31 and 4.6.33.

- This solution is done originally here https://github.com/compucorp/webform_civicrm-1/pull/1 and here https://github.com/compucorp/webform_civicrm-1/pull/2 ,  I just tried to improve it, fix some issues  and did some refactoring and testing.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -264,8 +264,9 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
 
     // Send confirmation for each generated Contribution.
     foreach ($this->ent['contribution'] as $contributionRow) {
-      $contributionID = $contributionRow['id'];
-      wf_civicrm_api('contribution', 'sendconfirmation', array('id' => $contributionID) + $this->contribution_page);
+      $params = array_merge($this->contribution_page, array('id' => $contributionRow['id']));
+      $params['payment_processor_id'] = $this->data['contribution'][1]['contribution'][1]['payment_processor_id'];
+      wf_civicrm_api('contribution', 'sendconfirmation', $params);
     }
   }
 
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1498,6 +1498,11 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $contributionsCount = $installmentsCount;
     }
 
+    $contributionParams = $this->contributionParams();
+    $frequencyUnit = wf_crm_aval($contributionParams, 'frequency_unit');
+
+    $currentDate = (new DateTime())->format('Y-m-d');
+
     foreach (wf_crm_aval($this->data, 'membership', array()) as $c => $memberships) {
       if (isset($this->existing_contacts[$c]) && !empty($memberships['number_of_membership'])) {
         foreach ($memberships['membership'] as $n => $lineItem) {
@@ -1515,16 +1520,24 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
             }
 
             if ($price) {
+              $singleInstallmentAmount= $this->calculateSingleInstallmentAmount($price, $contributionsCount);
+
+              $label = $this->getMembershipTypeField($type, 'name');
+              if ($contributionsCount > 1) {
+                $singleInstallmentPercentage = $this->calculateSingleInstallmentPercentage($singleInstallmentAmount, $price);
+                $label .= ' (' . $singleInstallmentPercentage . '%)';
+              }
+
               $lineItemData = array(
                 'qty' => $lineItem['num_terms'],
                 'financial_type_id' => $this->getMembershipTypeField($type, 'financial_type_id'),
-                'label' => $this->getMembershipTypeField($type, 'name'),
                 'element' => "civicrm_{$c}_membership_{$n}",
                 'entity_table' => 'civicrm_membership',
+                'unit_price' => $singleInstallmentAmount,
               );
 
               for ($i = 1; $i <= $contributionsCount; $i++) {
-                $lineItemData['unit_price'] = $this->calculateSingleInstallmentAmount($price, $contributionsCount);
+                $lineItemData['label'] = $label . CRM_Utils_Date::customFormat($this->calculateInstallmentReceiveDate($i, $frequencyUnit));
                 $this->line_items[] = $lineItemData;
               }
             }
@@ -1547,6 +1560,76 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
     }
     return round($this->totalContribution, 2);
+  }
+
+
+  /**
+   * Calculate and returns the receive date for
+   * a single installment.
+   *
+   * @param int $contributionNumber
+   * @param string $frequencyUnit
+   *
+   * @return string
+   */
+  private function calculateInstallmentReceiveDate($contributionNumber, $frequencyUnit) {
+    $currentDate = new DateTime();
+    $contributionNumber--;
+    switch ($frequencyUnit) {
+      case 'month':
+        $currentDate = $this->shiftMonth($currentDate, $contributionNumber);
+        break;
+      default:
+        $currentDate->modify('+' . $contributionNumber . ' ' . $frequencyUnit);
+        break;
+    }
+
+    return $currentDate->format('Y-m-d');
+  }
+  /**
+   * Copied from http://php.net/manual/en/datetime.modify.php#120513
+   *
+   * Returns end of months when we shift to a shorter or longer month
+   * workaround for http://php.net/manual/en/datetime.add.php#example-2489.
+   *
+   * Makes the assumption that shifting from the 28th Feb +1 month is 31st March
+   * Makes the assumption that shifting from the 28th Feb -1 month is 31st Jan
+   * Makes the assumption that shifting from the 29,30,31 Jan +1 month is 28th (or 29th) Feb
+   *
+   * @param DateTime $aDate
+   * @param int $months positive or negative
+   *
+   * @return DateTime new instance - original parameter is unchanged
+   */
+  private function shiftMonth(DateTime $aDate, $months) {
+    $dateA = clone($aDate);
+    $dateB = clone($aDate);
+    $plusMonths = clone($dateA->modify($months . ' Month'));
+    // Check whether reversing the month addition gives us the original day back.
+    if ($dateB != $dateA->modify($months * -1 . ' Month')) {
+      $result = $plusMonths->modify('last day of last month');
+    }
+    elseif ($aDate == $dateB->modify('last day of this month')) {
+      $result =  $plusMonths->modify('last day of this month');
+    }
+    else {
+      $result = $plusMonths;
+    }
+    return $result;
+  }
+
+  /**
+   * Calculates and returns the percentage value
+   * of the single installment compared
+   * to the total amount.
+   *
+   * @param float $installmentAmount
+   * @param float $totalAmount
+   *
+   * @return float
+   */
+  private function calculateSingleInstallmentPercentage($installmentAmount, $totalAmount) {
+    return round(($installmentAmount / $totalAmount) * 100, 2, PHP_ROUND_HALF_DOWN);
   }
 
   /**
@@ -1862,9 +1945,15 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     }
 
     $totalAmount = $params['total_amount'];
+    $singleInstallmentAmount = $this->calculateSingleInstallmentAmount($totalAmount, $params['installments']);
+    $params['non_deductible_amount'] = $params['total_amount'] = $singleInstallmentAmount;
     for ($i = 1; $i <= $contributionsCount; $i++) {
-      $params['non_deductible_amount'] = $params['total_amount'] = $this->calculateSingleInstallmentAmount($totalAmount, $params['installments']);
       $params['invoice_id'] = md5(uniqid(rand(), TRUE));
+
+      if (!empty($recurringContribution['values'][0]['frequency_unit'])) {
+        $params['receive_date'] = $this->calculateInstallmentReceiveDate($i, $recurringContribution['values'][0]['frequency_unit']);
+      }
+
       $contribution = wf_civicrm_api('contribution', 'create', $params);
       $this->ent['contribution'][$i]['id'] = $contribution['id'];
     }
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -261,8 +261,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $template = CRM_Core_Smarty::singleton();
       $template->assign('is_pay_later', 1);
     }
-    $contribute_id = $this->ent['contribution'][1]['id'];
-    wf_civicrm_api('contribution', 'sendconfirmation', array('id' => $contribute_id) + $this->contribution_page);
+
+    // Send confirmation for each generated Contribution.
+    foreach ($this->ent['contribution'] as $contributionRow) {
+      $contributionID = $contributionRow['id'];
+      wf_civicrm_api('contribution', 'sendconfirmation', array('id' => $contributionID) + $this->contribution_page);
+    }
   }
 
   /**
@@ -1189,7 +1193,6 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         foreach ($this->line_items as &$item) {
           if ($item['element'] == "civicrm_{$c}_membership_{$n}") {
             $item['membership_id'] = $result['id'];
-            break;
           }
         }
       }
@@ -1476,7 +1479,6 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    * Calculate line-items for this webform submission
    */
   private function tallyLineItems() {
-    // Contribution
     $fid = 'civicrm_1_contribution_1_contribution_total_amount';
     if (isset($this->enabled[$fid]) || $this->getData($fid) > 0) {
       $this->line_items[] = array(
@@ -1488,37 +1490,49 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         'entity_table' => 'civicrm_contribution',
       );
     }
-    // Membership
+
+    $contributionIsPayLater = empty($this->data['contribution'][1]['contribution'][1]['payment_processor_id']);
+    $installmentsCount = $this->getData('civicrm_1_contribution_1_contribution_installments');
+    $contributionsCount = 1;
+    if ($contributionIsPayLater  && $installmentsCount > 1) {
+      $contributionsCount = $installmentsCount;
+    }
+
     foreach (wf_crm_aval($this->data, 'membership', array()) as $c => $memberships) {
       if (isset($this->existing_contacts[$c]) && !empty($memberships['number_of_membership'])) {
-        foreach ($memberships['membership'] as $n => $item) {
-          if (!empty($item['membership_type_id'])) {
-            $type = $item['membership_type_id'];
+        foreach ($memberships['membership'] as $n => $lineItem) {
+          if (!empty($lineItem['membership_type_id'])) {
+            $type = $lineItem['membership_type_id'];
             $price = $this->getMembershipTypeField($type, 'minimum_fee');
 
             //if number of terms is set, regard membership fee field as price per term
             //if you choose to set dates manually while membership fee field is enabled, take the membership fee as total cost of this membership
-            if (isset($item['fee_amount'])) {
-              $price = $item['fee_amount'];
-              if (empty($item['num_terms'])) {
-                $item['num_terms'] = 1;
+            if (isset($lineItem['fee_amount'])) {
+              $price = $lineItem['fee_amount'];
+              if (empty($lineItem['num_terms'])) {
+                $lineItem['num_terms'] = 1;
               }
             }
 
             if ($price) {
-              $this->line_items[] = array(
-                'qty' => $item['num_terms'],
-                'unit_price' => $price,
+              $lineItemData = array(
+                'qty' => $lineItem['num_terms'],
                 'financial_type_id' => $this->getMembershipTypeField($type, 'financial_type_id'),
-                'label' => $this->getMembershipTypeField($type, 'name') . ": " . wf_crm_display_name($this->existing_contacts[$c]),
+                'label' => $this->getMembershipTypeField($type, 'name'),
                 'element' => "civicrm_{$c}_membership_{$n}",
                 'entity_table' => 'civicrm_membership',
               );
+
+              for ($i = 1; $i <= $contributionsCount; $i++) {
+                $lineItemData['unit_price'] = $this->calculateSingleInstallmentAmount($price, $contributionsCount);
+                $this->line_items[] = $lineItemData;
+              }
             }
           }
         }
       }
     }
+
     // Calculate totals
     $this->totalContribution = 0;
     foreach ($this->line_items as &$item) {
@@ -1533,6 +1547,27 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
     }
     return round($this->totalContribution, 2);
+  }
+
+  /**
+   * Calculates a single installment amount (price)
+   * if there are more than one installment.
+   *
+   * If there is only one installment then its
+   * amount will be the total amount.
+   *
+   * @param float $totalAmount
+   * @param int $installmentsCount
+   *
+   * @return float
+   */
+  private function calculateSingleInstallmentAmount($totalAmount, $installmentsCount) {
+    $amount =  $totalAmount;
+    if ($installmentsCount > 1) {
+      $amount = floor(($totalAmount / $installmentsCount) * 100) / 100;
+    }
+
+    return $amount;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


guanhuan:
Hi @colemanw,

Please let us know what we could do to accelerate the review of this PR?

Thanks!

KarinG:
Hi @guanhuan; for Financial PRs (see my note on another Compucorp PR) we really need small PRs/incremental improvements that have been tested on production sites.

omarabuhussein:
@KarinG , this PR cannot be any smaller  or get separated into incremental changes as you said in the previous one, all the work in it is related to each other, I separated the changes into separate commits though.

KarinG:
Of course you can break this up into smaller bits. And of course you can show us things work on your production sites; thousands of sites are using webform_civicrm module - we have no automated testing for it - so the bar for financial PRs is high; it has to be to prevent regressions;

omarabuhussein:
> Of course you can break this up into smaller bits.

hmm, Sometimes it does not make much sense to separate code into separate PRs just because the diff looks a bit big, but I will try to see if there is a way to do that.

> And of course you can show us things work on your production sites

I think this point is not clear Karin, I already tested it on both 4.6 and 4.7 (there are gifs above),  I am not clear what is the difference between doing that on test sites and doing it on a client production site ? It is practically the same, Right ?

Another thing, this is part of other upcoming improvements (PRs), so this alone is not something enough to be deployed to a client site without the other upcoming improvements since this PR only fix and improve only one thing. So I don't think that this is a good idea.

guanhuan:
Hi @KarinG,

While I agree that PR's can always be broken down into smaller trunks, I don't think it is reasonable to break down the PR in this case. This PR is to fix an existing issue related to using 'Pay Later' with recurring contribution which should not affect any other working functionality that currently exists in the module. 

Also as @omarabuhussein mentioned that this PR is to provide one solution, I do not think breaking it down will provide any value or achieve any independent user story. I also think that merging a partial of a solution which does not work on its own is probably not the best idea in the branch structure we are using in CiviCRM repositories because anything we merged will go directory into the master branch (in this case 7.x-4.x) which will indeed cause problems for anyone who is using the branch directly.

I have asked our QA to provide the test steps and results in the PR to facilitate the review. Hopefully that will provide enough support to the result of the changes.

Please let me know if the above sounds reasonable.

Best

Guanhuan

KarinG:
I'm sorry but I can't be evaluating / merging any Financial PRs that have not been live tested on production sites; you even admit you don't want to push this PR onto one of your client sites; this is not a draft/proposal queue; what we merge here goes into thousands of webform civicrm instances;

guanhuan:
Hi @KarinG,

While I agree that PR's can always be broken down into smaller trunks, I don't think it is reasonable to break down the PR in this case. This PR is to fix an existing issue related to using 'Pay Later' with recurring contribution which should not affect any other working functionality that currently exists in the module. 

Also as @omarabuhussein mentioned that this PR is to provide one solution, I do not think breaking it down will provide any value or achieve any independent user story. I also think that merging a partial of a solution which does not work on its own is probably not the best idea in the branch structure we are using in CiviCRM repositories because anything we merged will go directory into the master branch (in this case 7.x-4.x) which will indeed cause problems for anyone who is using the branch directly.

I have asked our QA to provide the test steps and results in the PR to facilitate the review. Hopefully that will provide enough support to the result of the changes.

Please let me know if the above sounds reasonable.

Best

Guanhuan

KarinG:
@guanhuan - I think we crossed posted. I'm expecting Financial PRs that I can test/deploy on a clone (or even live) of our own many webform civicrm projects - I would pick one that (in this case) has recurring enabled; 

PS - self review will not be adequate; would it e.g. catch 
`$params['non_deductible_amount'] = $params['total_amount'] = $singleInstallmentAmount;`

guanhuan:
@omarabuhussein please have a look into the issue mentioned in the comment above.

@KarinG Thanks for the suggestion and clarification. Yes, we can certainly change to test this on one of our client staging site so we can audit any issue in a production-equivalent environment.

We will post our test results here once the testings are complete.

madurak:
Hi @KarinG,

We have now tested this functionality on the latest version of webform_civicrm with and without this PR and there is no difference in the way it works, meaning recurring contributions are Not being created with payment processor. Used Paypal as the payment processor for this. 

**Test steps:**
1. Create a webform and enable Enable CiviCRM Processing on the CiviCRM tab section
2. Enable First Name, Last Name and Email for the Contact 1
3. Enable one membership for Contact 1
4. Enable Membership Fee on the membership tab
5. Select a Contribution Page that has PayPal option on the Contribution tab
6. Select "month" in "Frequency of Instalments". Enable number of instalments and Interval of Instalments
7. Complete a payment with Paypal and check the contribution created under the Contact 1

**Testing for PR-11 functionality:**
Tested functionality which PR 111 is introducing (Pay Later with Payment Plan in Instalments) on two different production equivalent environments and it works as expected. Test Steps are same as above with Pay Later. Here are the test results for this:
1. If a person chose to pay for a membership in 12 instalments via "pay later" option, a recurring contribution is created with 1/12 of the membership fee per instalment for 12 instalments
2. All 12 contributions are generated upon submission at once. All contributions will be in status "Pending" until being marked as "Complete" manually
3. The linked membership has full one year duration and will be activated upon the completion of the first contribution
4. On the contact record, after the purchase is completed, one recurring contribution with 1/12 of total amount is created. It has 12 instalments
5. On the contact record, 12 contributions with 1/12 of total amount is created with status of "pending". All the contributions are linked to the recurring contribution in point 4
Recurring payment is not working in the current webform_civicrm (without PR), so it doesn't seem like the PR is breaking anything from what we can see from our testings. 

**Versions tested on are:**
CiviCRM: 4.7.31
Drupal: 7.57
Webform_CiviCRM: 7.x-4.19

Recurring payment is not working in the current webform_civicrm (without PR), so it doesn't seem like the PR is breaking anything from what we can see from our testings. Also as mentioned above PR-111 functionality with pay later for payment in instalments works as expected as well.
Please let me know if there is anything else we need to check from our end. Thank you!

Regards,
Madura

KarinG:
Three points in addition to what I've already mentioned re: Financial PRs (length and production quality):

**1. You're not testing current/existing use cases properly** if you say:
> Recurring payment is not working in the current webform_civicrm (without PR), so it doesn't seem like the PR is breaking anything from what we can see from our testings.

It certainly is working - I've been using it with a number of clients since summer of 2017; sure there are things that can be improved - but the nuts and bolts are most definitively working:

image

My own payments to our swim club are coming out automatically [admins don't complete/edit anything] - monthly intervals, correct amount, onto my credit card statement; and I could have picked ACHEFT (Direct Debit) as well. Most of my clients use iATS Payments - if you say recurring contributions with payment processor credit card/IPN kind is not working - happy to review a PR specific to that. 

**2. What you're doing here (recurring pay later contributions) - are really pledges** - are they not? And every month/interval someone then converts the promised pending contribution's status to status=completed? And then also manually advances the next_scheduled_date for the contribution series? As well as manually marks the series as completed when all the promised (pending) payments have been received? Why even have a recurring series if you create all contributions up front? Recurring series generate contributions (on the next_scheduled_contribution_date) and you're already creating all of them. 

CiviCRM native does -not- have any concept of pay later recurring contributions - in fact you get an error message - telling you you can't set up a recurring contribution if you select Pay Later; 

image

**3. Some pieces of work you've done here are definitively helpful and will improve existing recurring functionality** - thank you for suggesting these.

guanhuan:
Hi @KarinG ,

As Madura mentioned, we have tested on Paypal and it does not seem to create the recurring contribution as it should with/ without this PR. However, what has not been mentioned is with "Dummy payment processor", the records are generated as expected (i.e. a recurring contribution and the first instalment contribution). Our change does not bring any impact to how online payments are processed currently. Paypal is a separate point that needs to be looked into by people who are using it with recurring contribution on Webform currently.

I have had a long call with Eileen to run all the business logic and specifications through her before we started [this project](https://civicrm.atlassian.net/wiki/spaces/PPI/pages/2260996/2.+Payment+Plan+Instalments). 

The differences between pledges and this are all the differences between pledges and payment instalments effectively. Pledges are not legally binding financial entity, they only stand for a verbal promise of contributions and they cannot be used in place of payment plans. In the example of organisations providing a year-long membership that allows people to pay for it in 12 instalments, a payment plan is essential.

Although CiviCRM contribution page does not support using 'Pay Later' with recurring contributions, Webform CiviCRM does allow it. This is common since Webform CiviCRM is not just an alternative public interface for CiviCRM but it provides a level of flexibility and possibility that CiviCRM itself can not. In fact, the way instalment and recurring contribution are calculated in Webform CiviCRM and CiviCRM Contribution page are different. Using instalment in Webform CiviCRM translates into the business we trying to fix here which is to split the total amount of the bill into number of instalments while the contribution page will simply recur the total amount of the bill.

The behaviour around this is currently not correct when 'Pay Later' is used which is what we are trying to fix here. Generating all instalment invoices upfront is common and especially important for the major offline ETF payments method such as Direct Debit. The invoices need to pre-exist so the admin can submit them to the bank to collect the relevant payments.

If you have any further question about the business logic, we can set up a call to discuss further. 

Best

Guanhuan

KarinG:
I spent a couple of hours on this over the weekend - I do understand what you're trying to do - but this is not the way to do it. We simply can not have a lame duck recurring series - that it completely disconnected from the pending contributions already generated; 

I'm 99.9% sure that Eileen would say: create a Payment Processor for Pay Later [we talked about that when we were presenting a presentation together at CiviCON Denver years ago]; 

If you want an Example - take our iATS Payments Extension - the section pertaining to: Direct Debit (we call that ACHEFT in North America); the only differences would be that:

- you would not be connecting it with an actual payment processor to submit the batch for verification by the banks [at least not at this point]; 
- and that your Payment Processor extension would not be automatically checking in daily with the bank to see if the ACHEFT have been approved [at least not at this point]

But all the code that you do want is already there, for example:

- every day a cron job can examine which recurring series with a Payment Processor of type Pay Later need to generate a recurring contribution attached to that series; 
- repeattransaction API is then used to generate that contribution; it is left in state = Pending (Pending verification)
- Recurring series is completely manageable for any possible required adjustments (number of intervals, next scheduled date, $amount) - and any future generated contributions will adhere to any schedule changes that an admin may make; 

It would be fairly trivial [especially with our Direct Debit example right there] to create such a Payment Processor; And then it would fit in and work 99% with webform_civicrm as is! 

My largest client alone processes over $70k/every month via Direct Debit [whatever we call it - it's straight out of people's bank accounts]; So I do understand this - and am trying to help you fwd here - not just for webform_civicrm but for Core as well;

If you need help creating a Payment Processor for Pay Later - let me know; 
Closing this for now until that Payment Processor for Pay Later is in place.

jamienovick:
Hi @KarinG ,

First off just wanted to say think you for taking the time to review in detail. We really appreciate the feedback. 

To start it's just helpful to me to articulate the discussion clearly just to make sure we are on the same page. What we are mainly talking about in this PR is around the appropriate methodology for managing instalment payments in CiviCRM and specifically one of 2 options:

1) "Just in time / JIT" - where the instalments are created on or just before the date of that the payment is due 
*or*
2) "Create upfront / when contract is agreed / CUF" - where the instalments are created as soon as the contract is agreed.

Currently CiviCRM online payment processors normally take the "JIT" approach, and if I understand correctly (correct me if I'm wrong) they will generally only create the payment if the payment processor calls back with details of the payment having been successfully taken (or having failed?).

In CiviCRM there is no existing core "offline" JIT process and if you use "Pay later" to create a set of instalments through webform, CiviCRM will create the first payment, but after that nothing will happen. This is blocked by a warning message on Contribution pages.

We are proposing that with the Pay later method we would move to using a "CUF" method and create the instalments up front. The instalments would be linked to a recurring contribution and we would in future PR's over the next few weeks (or as part of our membership extras extension) provide the facilities to manage and update those contributions as part of a payment plan appropriately. 

You disagree, and suggest that we should use a JIT method (as is the case with the online payment processors), or more accurately, leave the pay later method as is (whether we agree this is broken or not is a different point) and instead create a new offline payment processor to use JIT to achieve our aims.

NB. Just to be clear, Madura was perhaps not clear in stating that "instalments are broken". By this she that the "pay later" functionality with instalments is broken - or at least does not actually create any instalments (!). We do not intend to touch any existing online payment processor functionality at all for this at all.

Ok... so in terms of the JIT vs CUF

From an accounting perspective I think that there is no clear "right" answer for which of these is preferable. In both cases the accounting implications should be the same - i.e. the revenue is only recognised when the invoice is actually created based on the invoice date (in CiviCRM the contribution date... but I think at some point in future we should split out the payment date and have a separate "invoice" or tax point date if you agree... but I digress) and the receivable is only created when the money is owed - i.e. once the date of the instalment is passed (or more precisely the date of the journal entry would still be the invoice date). So I think from this perspective both methodologies are fine and really it is about what is best from a UX perspective and perhaps what is most flexible in terms of ongoing management of the "payment plan" (we are using payment plan as the provisional name improvement for "recurring contribution").

@guanhuan and I did a review of this which you can see here:

https://docs.google.com/spreadsheets/d/1JGl-fZwEcdZ_a5GyoF6QvXVnO7rMeGWGOoBUyBkMfdE/edit#gid=0

(Sorry it's all somewhat in shorthand)! 

So there appears to be 3 main advantages to the "create upfront" approach that we can see:

**1. Ability to show account "balance" of the payment plan** 

In the future we would like to show more information about the payment plan status such as the account balance. For example showing some more information on the account balance for the payment plan as per the mockup here: https://civicrm.atlassian.net/wiki/spaces/PPI/pages/2260996/2.+Payment+Plan+Instalments?preview=/2260996/13009001/Screen%20Shot%202018-01-12%20at%2017.31.47.png
Now in actual fact we don't need to have the future contributions already created to calculate things like the total amount, but this calc may be easier to do, easier to reports on and be more flexible about changes made for a specific month (see point 2 below).

**2. Ability to edit line items within an existing payment plan for future instalments**

The biggest reason to do this is to support the ability to have future changes to the recurring contributions at the line item level. With the current JIT methodology there is no way to add in additional line items to the next instalment. In fact the current data structure would expressly forbid that as the repeattransaction API will take the previous instalment contribution and copy it. We cannot change the previous instalment contribution as this would be changing the record of what actually was expected to be received. If we did want to specify some other "template" for the repeat transaction to take as it's default for the next contribution I think it would be difficult as we make invoices 1-1 with contributions, although I guess we could have some sort of "is template" flag like we have is_test etc. Anyway not very elegant.

With a CUF methodology instead what we are able to do is put in a mechanism so the user can add lines to any future contribution by simply editing the line items. We can then prompt if they want to update the future contributions as well and update the associated payment plan monthly amount with the new amount. 

This means that for our point 1 above we can have more flexibility in managing the actual amounts to be billed and also allow adding of additional line items during the course of the payment plan i.e. if someone buys a second membership (something we consider to be an "add on"). Btw we are thinking that we would like to fully support all these scenarios in the admin UI and in webform so that you could have more flexibility in how people self serve and renew memberships.

**3. Batching**

This point is possibly a bit more UK/EU specific for direct debit but if you are working in a process that requires you to batch up the payments then there is the minor point of just managing the timing of creating the payment and the date the batch is created. So if the 19th was your payment run, you would likely want the contributions created a few days in advance so that you could run your export on the day. This is minor and could probably be sorted out with some sort of setting of when in advance the payments get created but just a point to consider.

The disadvantages of CUF:

1. It's going to create a load of contributions and if your payment plan ends early then you need to deal with them. In our view this is perhaps actually a strength as in some ways the payment plan future instalments actually represent contractual obligation to be received. In a correct accounting world these would go to a bad debt contract account and not just never be created.

2. Need to do a lot more work around building all the above and it's going against the grain in Civi. Ok... but we're game for doing it as we think its the right way to go as for the reasons about, and perhaps we can then expand CiviCRM so that payment processors can chose to either be JIT or CUF, or even chose whether to be "reactive" ie. I only act if I get a callback, or "proactive" - I have a due contribution in the system and I'll call a payment processor to take that exact amount when the day comes. 

I'd welcome more of your thoughts on this, but in the meantime we are going to pursue the CUF method in our extension as a suitable way forward and we can leave to you whether you feel that we should "fix" pay later or just ignore it for now... I think if I was new to CiviCRM and wanted to have instalments working for a person signing up for membership then I would expect that functionality to work as we are suggesting but perhaps thats not important.

Best

J

ps. thanks for reading

@eileenmcnaughton @mattwire @colemanw @JMAConsulting interested in your views too if you have time and whether it is worth reopening this PR at all.

eileenmcnaughton:
So this is a long thread & I don't have great knowledge of webform. I tend to understand it as not being 100% aligned with core behaviour but I don't know the differences.

**One minor clarification** 

"they will generally only create the payment if the payment processor calls back with details of the payment having been successfully taken (or having failed?)."

Generally the approach is that they create the contribution as a pending contribution and confirm / fail it based on the processor results. This is a minor distinction and there are some cases where it does not occur but clarifying as the preferred model.

**Regarding the implementation here** 

- I WAS expecting that the offline-recurring payment processor instance would take responsibility for creating a sequence of pending contributions (with core creating the recur record & the initial contribution), probably using repeattransaction api. By the time the processor is called there should be a pending contribution to copy.
Most processors (well arguably but increasingly true) don't deal with IPNs for recurrings, but store tokens and a scheduled job runs and finds the ones that are due and processes them. There is a core job of sorts but more payment processors seem to implement their own rather than action those - see https://github.com/eileenmcnaughton/nz.co.fuzion.omnipaymultiprocessor/tree/master/api/v3/Job - which gives the flexibility of adding additional rules (e.g to create it in advance if need be, or to skip repeating the transaction and to complete one that is already created).

**Editing future transaction details**
"The biggest reason to do this is to support the ability to have future changes to the recurring contributions at the line item level." - this is something we discussed a lot in the past. Using an existing contribution as a template for future ones is inherently flawed. Pre-creating them makes them editable (esp with the line item editor extension). The plan we hatched in the past is to change to having a template of the line items attached to the contribution recur record - https://issues.civicrm.org/jira/browse/CRM-13839. It would be great to get that funded & done but I doubt it makes sense for you to pick that up

KarinG:
Hi @jamienovick 

Your team confused two concepts here: using your terminology: creating a CUF + a recurring series, a series which becomes a lame duck as for example #of installments; frequency_unit, next_sched_contribution_date are no longer meaningful as you've already generated multiple pending contributions;

Recurring contribution series in CiviCRM are designed to generate contributions - and they don't have to be of type Credit;

JIT is also used for Direct Debit. Typically a recurring job -> examines which series are due to create contributions by checking their next_sched_contribution_date [and yes you could easily make that e.g. next_sched_contribution minus 3 days so JIT minus x days]; The series then spits out/creates a Pending contribution. 

so JIT can be with our without a real payment processor that provides a thumbs up or down on an attempt to collect a Payment for the Contribution; 

Pay Later is no different from Pay Now - in fact one could argue it's always later; in cases of a CC payment processor it's seconds later [so generated contribution only sits in Pending state for seconds], in cases of a Direct Debit it's 3-5 days later, in cases of a bank EFT - it can sit in Pending state for a week - until perhaps you import a list of contributions (approvals from your bank);

Yes the way I've currently integrated the ability to create a recurring series with webform civicrm does indeed not do anything with Pay Later and Installments; It had not crossed my mind as CiviCRM does not let you do that - but if we did - I would want to make that fit within the existing Recurring Contribution Series concept: thus creating one Pending (first installment) Contribution + a Recurring Contribution Series and yes you need a Pay Later payment processor i.e. a very simple scheduled job -> to have that Recurring Contribution Series create the next Pending Contribution.

Having the series control the contributions -> currently already allows for a fair bit of flexibility: total $amount, schedule, it allows for catch-ups - pauses, adjustments; these things do happen [our house insurance payments were moved +14days after my husband lost my credit card... just for one month - then the insurance company bumped us back to our usual schedule once I had been able to give them the new number; just giving you this as an example as that's a pretty tight contract]; oh and yes issue receipt by email for every contribution as it is generated - or not. That's a very handy setting;

What can't be done yet is change the lineItems associated with a Recurring Contribution Series - and yes that would be awesome as when people have a tokenized credit card or tokenized bank account (for Direct Debit) on file - one currently has to cancel that series and set up a new one with the proper $amount/lineItems breakdown (which means re-tokenizing via the GUI 'sorry can you please give us your bank account details again' - or some switcheroo in the backend); 

ad 1. you can do account balance with a Recurring Contribution Series; Again doing this - political orgs have maximum contributions per constituent/year - so we have the projected numbers [donations + tickets + some fun math on any recurring contribution series with frequency, interval, $amount to year end] as we need to stop them before they go over [or we need to recalculate/adjust their Recurring Contribution Series so that future contributions generated are of a smaller $amount. It can be done; 

ad 2. lineItems - see above. Much better to be able to change the Recurring Contribution Series and then have all future Contribution spit out the new and correct lineItems [instead of having to update all pending future Contributions that have already been generated];

ad 3. what you refer to as Batching -> can be done with Recurring Contribution Series; see above - you can export all Pending Contributions that are generated on any day really. 

If you want to implement CUF in your extension work - absolutely; but I'd hate to see you go that way because the current recurring contribution series concepts and possibilities in CiviCRM are not well explained. And I've given that a shot here. Please do let me know if you have any questions.

For my professional orgs an recurring series with a defined number of installments at a certain frequency/interval is the payment plan. It really is.

KarinG:
PS - re: supporting changing lineItem amounts for upcoming recurring contributions. I will try and get that funded for during the CiviCamp Calgary May22-24 - make-it-happen-now Sprint.

eileenmcnaughton:
I'm not sure creating CUF is incompatible with a recurring contribution - since generally the job that acts off that schedule is part of the processor so it could take special actions etc. The issue about possibly changing schedules is a 'thing' that needs to be worked through in that scheme (although if you wanted CUF + ability to change schedule & still have a meaningful outer entity then a pledge would likely make more sense.)

@KarinG would be great if you make progress on line items for recurring contributions - although the sprint seems probably a bit too short to get it done in. I would consider a) adding a getTemplateContribution hook & b) developing the change in an extension first & then c) moving to core. Mostly because I suspect that it will be easier to get it stable first & then work through lining up a reviewer & getting someone else to review it, preferably in chunks.

KarinG:
CUF and a recuring contribution series; sure if the series changes - then a scheduled job could go back to all the pending contributions and edit them all (based on the new schedule) - but not sure why you would want to do extra work.

I’ll likely ‘put’ Alan Dixon on it; And yes we would add it to our iATS extension first. They are already sponsoring CiviCamp Calgary and at least some of their $ can go to this make-it-happen-now project

mattwire:
I think most things have already been said but to add my comments:
- I don't think we should be adding "complex" functionality to the core webform_civicrm module.  That belongs in it's own module and if we need to create events/hooks in the core module to allow that to happen then do that.
- There are restrictions / assumptions / complexities in CiviCRM core that may have a historical reason but are not relevant for many organisations.  For example, I recently got the restriction removed from core that prevented editing recurring contributions when linked to memberships.  Now, with https://github.com/civicrm/civicrm-core/pull/11556 I've cleaned up and improved the understanding of what core is actually doing, and hopefully helped pave the way for further customisation in the future (by way of hooks etc now the code is clean enough to see what is going on).
- I'd like to see changes being made to core/webform that simplifies what happens by default (eg this PR for webform_civicrm https://github.com/colemanw/webform_civicrm/pull/100 removes a whole load of custom coding to support specific payment processors and defers it all to the actual payment processor extension).  So, a simplification actually allows for additional functionality.

PR #113, opened 2018-02-26
 Adding support for overridden memebership until specific date

omarabuhussein:
Following the changes in this PR : https://github.com/civicrm/civicrm-core/pull/11622, I made here these changes to this module : 

1- relabeling the "Membership Status" field to  "Override Status". and relabeling the option "- Automatic -" to  "No".

2- A new checkbox to memebership tab is added 
called "Status Override Until Date" When checked, a date field will be added to the webform to allow admin/ user to select the date that the status override will expire. This field:

A- does not take any effect if "Override Status" field is set to "No" in the configuration or on the form
B- the selected date will  be saved into "status_override_end_date" field if a status is selected in the "Override Status" field in the configuration or on the form.

here are the test cases : 

full_test

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -1330,7 +1330,9 @@ function wf_crm_get_fields($var = 'fields') {
             'andor' => 'or',
             'action' => 'show',
             'rules' => array(
-              'contact_contact_sub_type' => $sets[$set]['sub_types'],
+              'contact_contact_sub_type' => array(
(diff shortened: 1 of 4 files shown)


colemanw:
Thanks for the PR. I've updated it with conditionals, and in the process I discovered and fixed a bug in conditional field handling.

omarabuhussein:
thanks @colemanw

PR #114, opened 2018-03-06
Add street_name, street_number, and street_unit to processed fields in utils.inc

adanielvv:
Needed for integration with postal code database.

This way it the street name, number and unit can be auto-filled when the address already is set,
and the street name and city can be autofilled (with external script) when the user fills in postal code and street number.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -1667,6 +1667,9 @@ function wf_crm_location_fields() {
 function wf_crm_address_fields() {
   return array(
     'street_address',
+    'street_name',
+    'street_number',
+    'street_unit',
     'city',
     'state_province_id',
     'country_id',
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -683,7 +683,7 @@ function wf_crm_get_fields($var = 'fields') {
       'name' => t('Email'),
       'type' => 'email',
     );
-    foreach (array('street_address' => t('Street Address'), 'supplemental_address_1' => t('Street Address # Line 2'), 'supplemental_address_2' => t('Street Address # Line 3'), 'city' => t('City')) as $key => $value) {
+    foreach (array('street_address' => t('Street Address'), 'street_name' => t('Street Name'), 'street_number' => t('Street Number'), 'street_unit' => t('Street Number Suffix'), 'supplemental_address_1' => t('Street Address # Line 2'), 'supplemental_address_2' => t('Street Address # Line 3'), 'city' => t('City')) as $key => $value) {
       $fields['address_' . $key] = array(
         'name' => $value,
         'type' => 'textfield',
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -115,11 +115,20 @@ class wf_crm_admin_form {
     $this->rebuildData();
     // Hack for nicer UX: pre-check phone, email, etc when user increments them
     if (!empty($_POST['_triggering_element_name'])) {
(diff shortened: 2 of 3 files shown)


colemanw:
Hmm, these fields are not usually entered manually. Why do you need them exposed to the form?

adanielvv:
I want the users to fill in the postal_code, street_number, and street_unit.
In The Netherlands that combination is unique, so the street_name and city can be filled from a database (the users needs to see it, but not manually change it).

KarinG:
What is street_unit? (long time ago) I used to live on the Jasmijnhof 5, 2631 JB in Nootdorp; I like what you’re trying to do - let’s standardize it best we can.

adanielvv:
@KarinG It is the Street Number Suffix. (toevoeging huisnummer)

adanielvv:
This is how I am using it right now:
https://github.com/DanielvV/civicrm_postcodenl_autocomplete/

KarinG:
Hey - can you point me to a live site for me to try this out?

adanielvv:
Yes @KarinG 
To test directly (can't click on 'next'): https://mijn.schreeuwomleven.nl/node/3
To see all steps: https://schreeuwomleven.nl/materiaal/

KarinG:
Very cool - @Coleman - just typing in 2631 JB and 5 -> it knows I'm on street: Jasmijnhof and in village: Nootdorp [and these two fields are disabled] - that's awesome for address standardization.

image

Will want to try this on an existing live site of our own; poke me tomorrow if I've not gotten back on that yet.

colemanw:
My concern is that the standardization here was done with custom code, however for a normal user they will now be presented with a confusing array of options when enabling an address, some of which are normally used for internal processing and not usually exposed to a form:

screenshot from 2018-03-06 11 18 48

I don't think it's clear from this list which ones to select for simply entering a normal address.

adanielvv:
For me it wasn't clear _without_ those separate fields for street name, number and suffix.

The definition of a simple normal address differs per country.
For a Dutch simple normal address you don't need State/Province nor District/Country.
And I had never heard about a Postal Code Suffix until now.
But the street name, number and optional suffix are separate entities that you need in order to enter a full address. Without it I can't use address autocompletion.
Then you have the Postal Code and the City.
Thats it for a Dutch address.

colemanw:
Ok. Maybe some help text is warranted.

adanielvv:
**Proposal help text**

_Street Address_
Single field for whole address line: Street Name, Number and Number Suffix. Don't use together with the separate fields.

_Street Name_
Use together with Street Number and Street Number Suffix as an alternative for Street Address.

_Street Number_
Use together with Street Name and Street Number Suffix as an alternative for Street Address.

_Street Number Suffix_
Use together with Street Name and Street Number as an alternative for Street Address.

adanielvv:
@colemanw What do you think?
I added the help text to my branch.

colemanw:
@DanielvV I've pushed up a change to also set defaults (I know this is US-centric, sorry, we could internationalize it at some point) and now I think there's less chance of confusion.

I think there's one more thing to sort out. Can you look at the function `wf_crm_address_fields()` and determine if it needs to be updated based on how it's used?

adanielvv:
Thanks!
US defaults: np

I added the fields to `wf_crm_address_fields()`

colemanw:
I'm going to merge this and do a followup PR to see if we can get that function `wf_crm_address_fields()` to make more sense. It's still missing some fields.

PR #115, opened 2018-03-06
Add token [submission:activity-id:?]

adanielvv:
Makes it possible to create a custom url with an activity id in it.


colemanw:
This looks great. Happy to merge this. One very tiny nitpick, we usually don't put in so many blank lines. I would take out the empty lines on L838, 841 & 844.

adanielvv:
Like this?

PR #116, opened 2018-03-06
Add tokens [submission:contact-id:?] and [submission:case-id:?]

adanielvv:
Handy if you need it for a custom url


colemanw:
Code looks good and feature is useful. Thanks for the PR :)

PR #117, opened 2018-03-18
Clean up processActivities function

mattwire:
With a webform_civicrm configured as:
1. Case1 (from URL case1=)
1. Activity1 (File on Case1)
1. Activity2 (File on Case1)

I was having trouble with the first activity not being filed on the case.  In theory this PR makes no changes to the functionality of the code, however it appears to fix the local issue and both activities correctly get filed on the case.

In any case, this is a nice code cleanup which makes the code easier to read.


colemanw:
Makes good sense to me.

PR #119, opened 2018-03-20
Issue #2948459: address fields don't update...

urlisse:
This fixes the issue of address fields that won't update with CiviCRM 4.7.28. Maybe some other changes are still required regarding this module's usage of null values. 

However, it seems that core is set to also accept `'null'` strings soon along with `NULL` values... 

* https://www.drupal.org/project/webform_civicrm/issues/2948459
* https://issues.civicrm.org/jira/browse/CRM-21474
* https://github.com/civicrm/civicrm-core/pull/11324

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -808,7 +808,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
               continue;
             }
             // Reset calculated values when updating an address
-            $params['master_id'] = $params['geo_code_1'] = $params['geo_code_2'] = 'null';
+            $params['master_id'] = $params['geo_code_1'] = $params['geo_code_2'] = NULL;
           }
           $params['contact_id'] = $cid;
           if (!empty($existing[$i])) {


colemanw:
I'm confused about how this regression happened. But this patch may cause breakages for older versions of CiviCRM.

KarinG:
@urlisse - can you please show me how to reproduce this bug? I've got a live test with some address PR edits going on our own 4.6 and 4.7 - and can add this in - if you can tell me exactly what to look for;

adanielvv:
This should be fixed in the CiviCRM side in version 4.7.30

colemanw:
@DanielvV thanks - can you link to the civicrm core issue that fixes this?

adanielvv:
https://github.com/civicrm/civicrm-core/pull/11324

adanielvv:
Sorry I caused the issue...
https://github.com/civicrm/civicrm-core/pull/11019

colemanw:
Thanks for following up @DanielvV

urlisse:
@colemanw Steps to reproduce under CiviCRM 4.7.28 + webform_civicrm 7.x-4.18 (or higher), in case that is still relevant:

1. Create a new webform with any basic contact fields (ex: first_name, last_name) and a few address fields (ex: street_address and postal_code) 
2. Acces the webform of a contact who already has an address
3. Change any address field and submit the form

The address entity won't update. I can't exactly remember if a second address entity was created during the process.

A colleague of mine forwarded who experienced the same issue forwarded me the error message below, but I don't have much context about it at the moment.

```
The CiviCRM "address create" API returned the error: "A fatal error was triggered" when called by function "preSave" on line 178 of wf_crm_webform_postprocess.inc with parameters: "Array  
               (  
                   [location_type_id] => 1  
                   [country_id] => 1039  
                   [geo_code_2] => null  
                   [geo_code_1] => null  
                   [master_id] => null  
                   [contact_id] => 17  
                   [is_primary] => 1  
                   [check_permissions] =>  
                   [version] => 3  
               )  
               "  
```

> But this patch may cause breakages for older versions of CiviCRM.

Are you planning to support the few "broken" versions of CiviCRM?

KarinG:
on 4.7.27 - this is NOT an issue;

image

KarinG:
Applying this PR to this specific 4.7.27 -> does not break it - which is a good thing.

image

jaapjansma:
I can confirm that wih CiviCRM 4.7.29 and webform_civicrm 7.x-4.20 this issue still happens. I have a site where this happens where an anonymous visitor fills in a form with address fields and his address is not created in civicrm.

The solution I found was to fix webform_civicrm see: https://www.drupal.org/project/webform_civicrm/issues/2948459#comment-12617026

colemanw:
@jaapjansma can you please submit a PR for that fix so we can review?

jaapjansma:
@colemanw my PR is here: https://github.com/colemanw/webform_civicrm/pull/137

PR #120, opened 2018-03-20
Fix possible bug in address handling

colemanw:
When I was recently reading the code it appeared that there was an oversight. A function named `wf_crm_address_fields()` was returning some but not all address fields, it omitted e.g. postal code suffix as well as all custom fields. But according to the discussion in #43 address fields will not save properly without it. I'm not exactly sure how to reproduce the error but want to flag this up with a possible fix.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -1292,6 +1292,7 @@ function wf_crm_get_fields($var = 'fields') {
     while ($dao->fetch()) {
       if (isset($custom_types[$dao->html_type])) {
         $set = 'cg' . $dao->custom_group_id;
+        // Place these custom fields directly into their entity
         if ($dao->entity_type == 'address' || $dao->entity_type == 'relationship' || $dao->entity_type == 'membership') {
           $set = $dao->entity_type;
         }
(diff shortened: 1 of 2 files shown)


colemanw:
Ping @omarabuhussein @nishant-bhorodia @mercurymotos @urlisse do any of you know more about this issue and have steps to reproduce the bug in #43 to see if this is a more complete fix?

mercurymotos:
Coleman:

AFAIK the patch works and I haven’t had any issues since.

Eliot

colemanw:
The problem is that only some address fields were fixed. So if you had a form using e.g. the postal code suffix fields it may not have worked.

omarabuhussein:
This was a long time ago , It seems that I  just added support for most important address fields and ignored the rest, and you here just get the fields list from the API to ensure that all address fields get handled and saved. So  I think this is a good change, will try to test it during this or next week.

colemanw:
@omarabuhussein thanks. I found it particularly confusing that the 2nd and 3rd address lines were being handled in a different way than the first. Was there any reason for that?

omarabuhussein:
ah I see, I don't really remember much the context for doing this but I think without doing that, the code won't be able to save more than one supplemental address.

omarabuhussein:
@colemanw 

just tested this, and it works fine for both normal and custom fields, though I discovered while testing that even the previous code was saving postal code suffix just fine .

colemanw:
I'm going to merge this as it cleans up the code and doesn't seem to cause any problems.

eyeonall:
Colman, this is definitely a better solution as it only adds the fields that were submitted. Thanks. The only thing I would ask is how to handle fields like the phone extension which was not getting updated when submitted.  If I go in and change the phone number, and that number no longer has an extension, the extension field is empty so it does not get changed. I have gone back and forth on how to deal with it and its driving me nuts....

Also, I was looking at the functions in there for adding custom fields. Is there anyway to use the API to update the fields? The reason I ask is that, for whatever reason, the pre/post hooks don't get fired off when doing the BAO update as it is now which makes it difficult to execute actions and create civi rules based on those fields. It looks like those routines were written when the API didn't cover custom fields very well whereas now it does. I will look at it myself and see if I can help if you believe that this is a feasible idea.

Thanks for everything,
A.J.

PR #121, opened 2018-03-21
DRU-2947955 - billing address always ends up is_primary if no matchin…

KarinG:
…g contact found regardless of how the webform is configured.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -745,6 +745,9 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    * @param int $c
    */
   private function saveContactLocation($contact, $cid, $c) {
+    // Check which location_type_id is to be set as is_primary=1;
+    $is_primary_address_location_type = $contact['address'][1]['location_type_id'];
+    $is_primary_email_location_type = $contact['email'][1]['location_type_id'];
     foreach (wf_crm_location_fields() as $location) {
       if (!empty($contact[$location])) {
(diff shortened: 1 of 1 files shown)


KarinG:
Originally reported by Dave: https://www.drupal.org/project/webform_civicrm/issues/2947955

To reproduce: fill out a webform with an Address 1 (Primary) as well as a Payment (which injects the Billing address block); that can be a simple Contribution or a Membership; logout! And fill out the form as a completely new contact (you will want to be unmatched); pay with fake monies -> the billing address becomes the is_primary instead of the location_type_id specified in the webform config.

Live tested on two production 4.7.27 sites - with this PR: Address 1 (primary) does indeed end up as is_primary - in this example: work address;

image

I found out that during this Billing address gets saved first and b/c it's the first address for this contact CiviCRM API will make it is_primary=1 even if you specify you want is_primary=0

I spent a fair bit of time in `self::reorderLocationValues` - then decided that rather than re-ordering the addresses to help ensure the last one to be saved as is_primary actually is the real is_primary - better to create a safe logic around that;

davejenx:
I have tested this, as described at https://www.drupal.org/project/webform_civicrm/issues/2947955 . I agree with Karin's assessment at https://www.drupal.org/project/webform_civicrm/issues/2947955#comment-12537336 and reckon this PR is definitely better in than not. I can see arguments either way about the desired behaviour when an existing contact has a primary address of a different type from the first/primary address on the webform. For the other use cases described, in my view this PR brings a clear improvement.

PR #122, opened 2018-03-21
API recommends passing in Payment Processor Id (avoids mis-guesses).

KarinG:
Thank you to @omarabuhussein for pointing this out.
PS - Also includes some PHPStorm formatting edits;

Live in production on a 4.7.27 - is properly dispatching receipts for Memberships purchased via a webform civicrm;

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -239,11 +239,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       webform_submission_send_mail($this->node, $this->submission);
       $this->submitIPNPayment();
     }
-      // Send receipt
+    // Send receipt
     if (empty($this->submission->is_draft)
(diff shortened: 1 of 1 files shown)


omarabuhussein:
np @KarinG , do u want me to test  it?

KarinG:
If you could please confirm @omarabuhussein - that would be great - then we can merge this in.

omarabuhussein:
sure, will try to do that before end of this week.

omarabuhussein:
just tested this on latest civi 4.7,  it works well for both using payment processor (I tested the default dummay processor) and for pay later, and if the contibution page is configured to use more than one payment processor then the errrors used to appear before are not appearing anymore.

Also Since this was part of my Previous PR, I think we can also assume that its also already tested using an actual payment processor (paypal) since our QA tested it so I think it is fine to merge this.

@KarinG

KarinG:
Thank you for confirming this.

PR #124, opened 2018-03-23
No auto fill custom data

vakeesan26:
https://www.drupal.org/project/webform_civicrm/issues/2955665

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -168,8 +168,7 @@ abstract class wf_crm_webform_base {
     foreach ($contact as $k => $v) {
       if (substr($k, 0, 12) == 'number_of_cg' && !empty($v)) {
         if (!in_array(substr($k, 10), $exclude)) {
-          $info += $this->getCustomData($cid);
-          break;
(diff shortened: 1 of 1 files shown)


colemanw:
Can you please list the steps to reproduce this bug?

vakeesan26:
In existing contact field setting we have a setting called "Skip auto filling of"
In this setting i have chosen only one custom field group.
so when i view the webform i expected those custom group fields as empty.
but it is not working as [expected.]
[in the setting if i choose all custom groups then when i view the webform all custom fields were empty]

[ Auto Fill ](https://www.drupal.org/files/issues/2018-03-28/auto_fill.JPG)

ssngbala:
can you please merge this patch ?

colemanw:
@ssngbala if you have reviewed and tested this patch, please describe how you reproduced the bug and your results before/after this patch.

vakeesan26:
$this->getCustomData($cid); loading all the custom data regardless $exclude setting.

colemanw:
I checked this over and determined that using `array_filter()` would be more efficient than calling the `getCustomData` function repeatedly. Fixed via 2e0d2b9

PR #126, opened 2018-03-28
try getfields for address fields rather than hardcoding them and add …

eyeonall:
The hard coded fields in wf_crm_address_fields does not account for address custom fields. If we change that function to try and use the api getfields method first, list of all address subfields can be obtained, including the supplemental fields.  If the getfields call fails, then revert to the default hardcoded fields. Whatever the fields turn out to be, using drupal_static ensures that the api call to getfields is only called once per request.

When updating a phone with an extension, if we clear out, or change the extension, it is not submitted for update. The module does a good job with adding the correct field parameter to the submitted values so changing the loop to add all of the submitted values allows for other location entity custom fields and the phone extension to be updated.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -1665,16 +1665,34 @@ function wf_crm_location_fields() {
  * @return array
  */
 function wf_crm_address_fields() {
-  return array(
-    'street_address',
-    'street_name',
-    'street_number',
-    'street_unit',
-    'city',
-    'state_province_id',
-    'country_id',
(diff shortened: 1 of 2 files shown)


colemanw:
Hey AJ, @eyeonall thanks for submitting this PR.
I recently did a similar change in #120 which may or may not have solved your problem. Could you please retest with the latest -dev of this module, and see if you are still seeing any problems with address fields?

PR #127, opened 2018-03-30
correct number format before passing to API

sunilpawar:
https://www.drupal.org/project/webform_civicrm/issues/2953811


colemanw:
I'm hesitant to add hacks to work around core bugs, as it becomes complicated to manage when some versions of core contain the bug and others do not. I think a PR against core to fix the api would be better.

PR #128, opened 2018-04-03
Handle Recurring contribution for is Pay Later with installments correctly

omarabuhussein:
This PR is an alternative to what was discussed here : 

https://github.com/colemanw/webform_civicrm/pull/111

This new PR basicly do two related things : 

1- It make deferred payment (paylater or IPN payment) behave similar to live payments in case there is more than one installment by creating a recurring contribution and a single  pending contribution that equals the amount of one installment.

2- It make sure that if the deferred payment ispaid by payment processor that uses the core **Payment_Manual** class, then it will be considered as paylater and the contribution **is_pay_later** flag will be set to TRUE.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -238,7 +238,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     if (empty($this->submission->is_draft) && !empty($this->ent['contribution'][1]['id']) && $this->contributionIsIncomplete && !$this->contributionIsPayLater) {
       webform_submission_send_mail($this->node, $this->submission);
       $this->submitIPNPayment();
-  }
+    }
     // Send receipt
     if (empty($this->submission->is_draft)
       && !empty($this->ent['contribution'][1]['id'])
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -238,7 +238,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     if (empty($this->submission->is_draft) && !empty($this->ent['contribution'][1]['id']) && $this->contributionIsIncomplete && !$this->contributionIsPayLater) {
       webform_submission_send_mail($this->node, $this->submission);
       $this->submitIPNPayment();
-    }
+  }
     // Send receipt
     if (empty($this->submission->is_draft)
       && !empty($this->ent['contribution'][1]['id'])
@@ -1763,7 +1763,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   /**
    * Generate recurring contribution and transact the first one
    */
-  private function contributionRecur($contributionParams) {
+  private function contributionRecur($contributionParams, $paymentType = 'live') {
     $installments = $contributionParams['installments'];
     if ($installments != 0) {
       // DRU-2862396 - we must ensure to only present 2 decimals to CiviCRM:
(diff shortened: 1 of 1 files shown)


guanhuan:
@madurak please test this on staging site tomorrow.

guanhuan:
Hi @KarinG,

We have tested the PR and below are the results per payment processor tested:

1. Offline payment processor (using Payment_Manual class)
   a. Pay in instalment
        - a recurring contribution is created
        - one instalment contribution with amount=total amount/no of instalment is created with status Pending (Pay Later)
   b. Single contribution
        - one contribution with total amount is created with status Pending (Pay Later)
2. Pay later
   a. Pay in instalment
        - a recurring contribution is created
        - one instalment contribution with amount=total amount/no of instalment is created with status Pending (Pay Later)
   b. Single contribution
        - one contribution with total amount is created with status Pending (Pay Later)
3. Dummy
   a. Pay in instalment
        - a recurring contribution is created
        - one instalment contribution with amount=total amount/no of instalment is created with status Completed
   b. Single contribution
        - one contribution with total amount is created with status Completed
4. Paypal
   a. Pay in instalment
        - a recurring contribution is created
        - one instalment contribution with amount=total amount/no of instalment is created with status Pending and **does not** get marked as Completed when IPN shows successful
   b. Single contribution
        - one contribution with total amount is created with status Pending and is marked as Completed when IPN shows successful

All results are positive expect payment in instalment with Paypal. Since payment in instalment via webform never worked with IPN payment processor before, I think this is probably caused by the IPN handling class of Paypal processor is not ready to handle it.

Overall I think this looks good from our side. Please review and let us know if you have any comment.

guanhuan:
Alternatively, we can exclude IPN payments from the fix so it stays untouched if necessary. Our scope is to make offline payments working with recurring contribution.

@colemanw @eileenmcnaughton Could you help review this too?

KarinG:
@guanhuan - yes focus on offline payments only; what is the difference between 1. and 2.? Isn't offline = pay later?

guanhuan:
@KarinG Cool we will update the PR to exclude IPN payments.

Regarding your question, it doesn't have a huge difference expect offline payment processor are payment processors that use Payment_Manual class while Pay Later option does not supply a payment processor.

eileenmcnaughton:
sigh - I wish we could rip that whole is_pay_later field out. I don't like hard-coding it in on a specific processor class. I had hoped we could just ignore it for the manual processor.

I guess other options if people are wedded to that flag are
1) setting in  hook on the plugin being used with the manual processor.
2) having a function on the processor isPayLater to call (from webform )
3) having a more generic function on the processor to getAdditionalContributionFields to merge in.

Perhaps it's not so bad since the Manual Processor is a bit 'special' in core (being processor 0) & I just have to get over it.

omarabuhussein:
> Alternatively, we can exclude IPN payments from the fix so it stays untouched if necessary.

done here https://github.com/colemanw/webform_civicrm/pull/128/commits/f4cba2fb8e33d92293111906f4f79bbba262f574

KarinG:
Ok, I've pushed two small commits; I was tempted to also changed the 'deferred' terminology to 'pay later' as we really don't need more contribution status terminology - but it's ok;

What's important is that this now works - exactly the same for Pending (Incomplete Transaction) and Pending (Pay Later).

For posterity - my test/screens
[using 3 installments in this form example]:

Toggle Payment Processor to: Pay Later
image

Toggle Payment Processor to: ACHEFT
image

Recent Contributions: ACHEFT one is Pending (Incomplete Transaction) and Pay Later one Pending (Pay Later)
image

Recurring Series:
image

PR #131, opened 2018-04-20
Prevent a failure on PHP<5.5 with reset and make this check easier to…

mattwire:
… read

Follow on from #118.  

No functional change, the check is doing the same thing, it just works on PHP<5.5 and is easier to read.

@colemanw


colemanw:
Thanks for submitting that @mattwire

PR #136, opened 2018-05-01
Allow token to be a default value for contribution amount

jitendrapurohit:
Overview
Allow token string to be a default value for contribution amount input field.

Before
Error when token string is passed to the default value of contribution amount.

image

After
Fixed. 

Technical Details
Validation function is modified to not error for token string using `token_replace` function.

Tested string value

  - `abc` - errors
 - `[current-page:query:civicrm_1_contribution_1_contribution_total_amount]` - works fine.


KarinG:
Excellent - all tested/reviewed on two continents; thank you @jitendrapurohit and @fuzionnz

PR #138, opened 2018-05-23
Fix notice error on custom group create/update

jitendrapurohit:
Overview
Notice fix after custom group is created/edited.

Before
Following notice error is displayed when a custom group in civicrm is edited/updated.

image

After
Fixed.


KarinG:
Thank you @jitendrapurohit

PR #139, opened 2018-05-27
DRU-2966760 - Fix Notice

KarinG:
Overview
https://www.drupal.org/project/webform_civicrm/issues/2966760 - Fix Notice

Before
Notice: Undefined index: address in wf_crm_webform_postprocess->saveContactLocation() (line 751 of /sites/all/modules/webform_civicrm/includes/wf_crm_webform_postprocess.inc).

After
No more Notice

Technical Details
Easiest way to address this.


colemanw:
@KarinG looks to me like that change would have caused a different notice because the variables `$is_primary_address_location_type` and `$is_primary_email_location_type` were referenced below but conditionally defined.

KarinG:
Not sure why it didn’t give me another notice - but thank you for rewriting this.

PR #140, opened 2018-05-29
Calculate line_items when using installments.

KarinG:
Before
Currently when enabling the functionality to pay the amount due in installments -> the first Contribution is processed properly, the Recurring Series is set up properly - but the line_items of first Contribution are not correct (they are still the line_items for the total obligation rather than for one single installment); 

After
Fixed - the view Contribution looks good at line_item level:
image

Comments
The obligation shows on the webform_civicrm checkout page:
image

An HTML markup element can be used to generate any text like:
image

Note that submission:values:installment_amount is a webform_calculator field - this token does not currently work yet - but since this allows any form builder to select the language appropriate for their Membership/Event Fee/Donation form - tokens are the way forward here.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1780,12 +1780,13 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $contributionRecurAmount = floor(($contributionParams['total_amount'] / $installments) * 100) / 100;
       $contributionFirstAmount = $contributionParams['total_amount'] - $contributionRecurAmount * ($installments - 1);
       $salesTaxFirstAmount = $contributionParams['tax_amount'] - floor(($contributionParams['tax_amount'] / $installments) * 100) / 100 * ($installments - 1);
-    }
-    // Calculate the line_items for the contributionFirst:
-    foreach ($this->line_items as $key => $k) {
-      $this->line_items[$key]['unit_price'] = $k['unit_price'] / $installments;
-      $this->line_items[$key]['line_total'] = $k['line_total'] / $installments;
-      $this->line_items[$key]['tax_amount'] = $k['tax_amount'] / $installments;
+
+      // Calculate the line_items for the contributionFirst:
+      foreach ($this->line_items as $key => $k) {
+        $this->line_items[$key]['unit_price'] = $k['unit_price'] / $installments;
+        $this->line_items[$key]['line_total'] = $k['line_total'] / $installments;
+        $this->line_items[$key]['tax_amount'] = $k['tax_amount'] / $installments;
+      }
     }
 
     // Create Params for Creating the Recurring Contribution Series and Create it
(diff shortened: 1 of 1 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1779,6 +1779,13 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       // DRU-2862396 - we must ensure to only present 2 decimals to CiviCRM:
       $contributionRecurAmount = floor(($contributionParams['total_amount'] / $installments) * 100) / 100;
       $contributionFirstAmount = $contributionParams['total_amount'] - $contributionRecurAmount * ($installments - 1);
+      $salesTaxFirstAmount = $contributionParams['tax_amount'] - floor(($contributionParams['tax_amount'] / $installments) * 100) / 100 * ($installments - 1);
+    }
+    // Calculate the line_items for the contributionFirst:
+    foreach ($this->line_items as $key => $k) {
+      $this->line_items[$key]['unit_price'] = $k['unit_price'] / $installments;
+      $this->line_items[$key]['line_total'] = $k['line_total'] / $installments;
+      $this->line_items[$key]['tax_amount'] = $k['tax_amount'] / $installments;
```
colemanw: Hmm, reading the code it looks like on line 1778 `$installments` might be zero. But this foreach loop runs unconditionally, which could give us a divide by zero error?

KarinG: You’re correct - will move that into the check for = 0 above;

colemanw: The existing code seems a little odd because I would think the test should be `if ($installments > 1)` instead of `if ($installments != 0)`. Is there logic elsewhere to ensure we don't create a "recurring" contribution with only 1 occurrence?

KarinG: Yes that already happens before -> line 1754:
if ($numInstallments != 1 && !empty($frequencyInterval)) {
    $result = $this->contributionRecur($contributionParams);

KarinG: 0 is a special case that means don't do installments - but start a recurring series - open ended commitment;

KarinG:
Conclusion: allowing 0 keeps the door open for an option to create open-ended series; That's not working yet though - but we could make that happen in future.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1803,6 +1811,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
 
     // Run the Transaction - and Create the Contribution Record - relay Recurring Series information in addition to the already existing Params [and re-key where needed]; at times two keys are required
     $contributionParams['total_amount'] = $contributionFirstAmount;
+    $contributionParams['tax_amount'] = $salesTaxFirstAmount;
```
mattwire, line 1814: The problem with this code at the moment is that you can easily enter installments=0 via the UI but that causes all the "amount" variables to be unset and throws a fatal exception in CiviCRM because of a DB error - ie. you can't call ContributionRecur/Contribution.create without an amount set.

KarinG: I’m only addressing the math i.e. lineItem calculation in this PR; 

handling ===0 is next;

mattwire:
@KarinG I was going to review this properly but I see it's now been merged :-)

KarinG:
I asked Coleman to review it as I wanted to get this in before looking at your 141 :-)

colemanw:
@mattwire another set of eyes on it would still be very welcome :)

PR #141, opened 2018-06-11
If installments=0 we create an open-ended recurring contribution but we still need an amount

mattwire:
Overview
If we configure the webform_civicrm with installments=0, frequency=weekly (or any other frequency) then we should create an open-ended recurring contribution.

Before
Create recurring contribution/contribution fails because amount is not set.  You have to specify `installments=0` to create an open-ended recur.

After
Create recurring contribution/contribution succeeds because amount is set.  You don't have to specify `installments` at all to create an open-ended recur.

Technical Notes
1. The code makes assumptions that tax_amount will be set - it is optional (ie. you may not have tax and invoicing enabled in Civi).  The current webform_civicrm code will throw PHP notices if it is not set.
2. On the CiviCRM side, "installments" is very much an optional parameter and is treated differently by different processors - it doesn't actually do what it should in many cases.  webform_civicrm is doing something sensible - however, it is assuming that it will always be defined, which is not what CiviCRM expects.  This PR updates the webform_civicrm handling to allow either `installments = 0` or `!isset(installments)` to mean an open-ended recurring contribution if frequency unit/frequency interval is set.  This mimics the behaviour in CiviCRM.
3. This PR makes sure that variables are always defined before being used (`$salesTaxFirstAmount` and 
`$contributionFirstAmount` were not being set if `installments = 0` previously leading to undefined behaviour - an API exception).
4. I'm not able to test the changes in `createDeferredPayment()` as I don't know how to get there.  However, they are inline with changes in the other functions.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1748,8 +1748,8 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   private function submitLivePayment() {
     $contributionParams = $this->contributionParams();
 
-    // Only if #installments = 1, do we process a single transaction/contribution. #installments = 0 in CiviCRM Core means open ended commitment;
-    $numInstallments = wf_crm_aval($contributionParams, 'installments', 1, TRUE);
+    // Only if #installments = 1, do we process a single transaction/contribution. #installments = 0 (or not set) in CiviCRM Core means open ended commitment;
+    $numInstallments = wf_crm_aval($contributionParams, 'installments', NULL, TRUE);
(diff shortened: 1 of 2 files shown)


KarinG:
Thank you - can you please have a look at/confirm https://github.com/colemanw/webform_civicrm/pull/140 - let’s get that in first before we extend to ===0

KarinG:
Hi @mattwire - thank you for this; I ran different scenarios: Membership + Tax & Contribution -> in two monthly installments, with installments =1 and with =0 and weekly and everything adds up nicely right down to lineItems and taxes. Also thank you also for the tidying up bits;

PR #142, opened 2018-06-15
Add filter on is_public event field

aydun:
Overview
Allow users to filter the list of events by whether the event is marked as 'Public'

Before
Events are shown regardless of the 'Public' setting, so non-public events are still displayed.

After
On the Event Registration tab there is a new dropdown for 'Show Public Events' that allows the selection of 'Private and Public', 'Private' or 'Public'.  The default is 'Private and Public' to match the current behaviour.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -206,7 +206,7 @@ function wf_crm_get_events($reg_options, $context) {
   if ($event_types) {
     $sql .= ' AND event_type_id IN ( ' . implode(", ", $event_types) . ' ) ';
   }
-  if (is_numeric($reg_options['show_public_events'])) {
+  if (is_numeric(wf_crm_aval($reg_options, 'show_public_events'))) {
     $sql .= ' AND is_public = ' . $reg_options['show_public_events'];
   }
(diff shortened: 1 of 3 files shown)


includes/utils.inc
```diff
@@ -205,6 +205,9 @@ function wf_crm_get_events($reg_options, $context) {
   if (is_numeric($reg_options['event_type'])) {
     $sql .= ' AND event_type_id = ' . $reg_options['event_type'];
   }
+  if (is_numeric($reg_options['show_public_events'])) {
+    $sql .= ' AND is_public = ' . $reg_options['show_public_events'];
+  }
```
colemanw: For backward compatibility with existing forms in which this variable won't be set, I think the conditional should be

    if (is_numeric(wf_crm_aval($reg_options, 'show_public_events'))) {

colemanw (changes requested):

includes/utils.inc
aydun: Updated as suggested

colemanw (approved):

PR #143, opened 2018-06-15
Allow event selection by multiple types

aydun:
Overview
---------------------------------------
Allow events to be filtered by selecting multiple event types

Before
On the Event Registration tab there is a dropdown for 'Show Events of Type' that allows the selection of a single type.

After
The dropdown is renamed 'Show Events of Type(s)' and allows the selection of multiple types.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -202,8 +202,11 @@ function wf_crm_get_events($reg_options, $context) {
     $date_future = date('Y-m-d H:i:s', strtotime($date_future));
     $sql .= " AND (end_date <= '$date_future' OR end_date IS NULL)";
   }
-  if (is_numeric($reg_options['event_type'])) {
-    $sql .= ' AND event_type_id = ' . $reg_options['event_type'];
+  if (is_array($reg_options['event_type'])) {
+    $event_types = array_filter($reg_options['event_type'], "is_numeric");
+    if ($event_types) {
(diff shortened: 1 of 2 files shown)


includes/utils.inc
```diff
@@ -202,8 +202,11 @@ function wf_crm_get_events($reg_options, $context) {
     $date_future = date('Y-m-d H:i:s', strtotime($date_future));
     $sql .= " AND (end_date <= '$date_future' OR end_date IS NULL)";
   }
-  if (is_numeric($reg_options['event_type'])) {
-    $sql .= ' AND event_type_id = ' . $reg_options['event_type'];
+  if (is_array($reg_options['event_type'])) {
```
colemanw: Since there are existing webforms which did not save this data as an array, we need to fall-back. I suggest casting it as an array then continuing with your new code.

colemanw (changes requested):

includes/utils.inc
aydun: Ok, changed

colemanw (approved):

petednz:
thx guys - were just starting to work on this when we found this PR. Guess our version wasn't on latest (though is a newish site) - oh maybe this hasn't made it in a Release yet.

KarinG:
@colemanw did a release just weeks ago so this should be in.

PR #146, opened 2018-07-08
Check entity type is set on webform settings before compare

beto-aveiga:
Overview
Webforms with phones and selectable type are not retrieving data from contacts on webform load. The user sees as if no information was previously saved and once the user saves data (if any) will be overwritten.

Before
1. Initial data for contact
image

Same Initial data. Showing type and location for phone.
image

2. Phone field configured with type selectable by the user. 
image

3. Retrieving phone data without checking existence of type index on webform settings causes data to be seen as different from the webform configuration, thus not showing any phone related data at all from the contact.
image

After

1. Repeat steps 1 to 3 from "Before".

2. Information is now retrieved correctly, producing expected results on save.
image

Technical Details
- Attaching exported webform with node export:
[webform-phone-with-selectable-type.txt](https://github.com/colemanw/webform_civicrm/files/2174145/webform-phone-with-selectable-type.txt)

- Site created for testing with civibuild using this command
`civibuild create civi12 --type drupal-demo --civi-ver 4.7.27 --url http://civi12 --web-root /home/beto/buildkit/build/civi12`

- CiviCRM currently uses civicrm_webform 7.20, however this module was replaced with the last version found here https://github.com/colemanw/webform_civicrm

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -352,9 +352,11 @@ abstract class wf_crm_webform_base {
       $valueFound = false;
       foreach($values as $key => $value){
         if ((in_array($ent, array('address', 'email')) && $value['location_type_id'] == $setting['location_type_id'])
-           ||
-          ($value['location_type_id'] == $setting['location_type_id'] && $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])
-          ) {
+          || (
+             $value['location_type_id'] == $setting['location_type_id'] &&
+             (!isset($setting[$ent.'_type_id']) || $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])
+             )
+        ) {
             $reorderedArray[$key] = $value;
             $valueFound = true;
             unset($values[$key]);


beto-aveiga:
@KarinG Please don't test/merge it yet. I have to fix a minor thing in the condition to avoid notices.

beto-aveiga:
@KarinG You can continue with the testing. I corrected the last commit, the condition had a small issue. Now it is good.

beto-aveiga:
@KarinG @colemanw I merged locally https://github.com/colemanw/webform_civicrm/pull/146 and https://github.com/colemanw/webform_civicrm/pull/148 into the master of this repo. I made several tests and everything worked just fine.
I tested with multiple emails, multiple addresses, and multiple phones. Also tested with multiple fields with location types selectable by the user, and the results were the the expected results.
If there is something else to do please tell me, I need the both mentioned branches get merged as soon as possible to incorporate those changes in a patched version of webform_civicrm, for CiviHR and probably CiviCRM.
Thanks!

colemanw:
I'm comfortable merging this because it's a simple addition of `isset()` which is generally a good practice.

PR #149, opened 2018-07-19
Autofill contact from relationships

colemanw:
Overview
Allow existing contact to be autofilled from any relationship, not just to contact 1.

screenshot from 2018-07-19 14-21-46

Before
The default value could only be set "by relationship" to contact 1.

After
The default value can be set "by relationship" to any preceeding contact on the webform (e.g. contact 3 can be set from contacts 1 or 2 but not 4 or 5).

Comments
This is for *prepopulation* only. Selecting a different contact via autocomplete after the form is already loaded won't do anything.

------
https://www.drupal.org/project/webform_civicrm/issues/2797345

### DIFF
--- a/css/webform_civicrm_admin.css
+++ b/css/webform_civicrm_admin.css
@@ -190,3 +190,13 @@ form .civi-icon.grant {
 form .civi-icon.case {
   background-position:-203px 0;
 }
+
+/* Contact component form */
+#webform-component-edit-form div.form-item-extra-default,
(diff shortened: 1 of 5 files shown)


KarinG:
Cool!

stesi561:
This is great. We've got this working to allow us to populate a default value for a contact which is related to cid1 via an intermediate contact - i.e. - cid1 = Contact1 and cid3 = Contact C.

Contact A <= relationship => Contact B
Contact B <= relationship => Contact C

PR #151, opened 2018-07-30
Add defaults to resetFields call when creating new contact

AsylumSeekersCentre:
…in webform_civicrm_forms.js, solving the bug of missing CiviCRM defaults.

Overview
When creating a new contact in a webform, defaults for Civi fields are not preserved, while defaults for non-Civi fields remain. This change preserves the defaults for Civi fields as well.

Before
This is our Test Defaults webform:
webform_civicrm_defaults_01select
Here is what happens when selecting an existing Contact (all defaults are preserved, fields with saved data are updated):
webform_civicrm_defaults_02existing
Here is what happens when creating a new contact (Civi defaults are not preserved):
webform_civicrm_defaults_03create

After
Now all defaults are preserved when creating a new Contact:
webform_civicrm_defaults_04createwithdefaults

Technical Details
The _defaults_ parameter of the _resetFields_ function seems to be relatively new. It looks like it can be used when creating a new Contact, and adding it here doesn't appear to break anything else (but maybe there's a use case I'm not considering).

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -12,7 +12,7 @@ var wfCivi = (function ($, D) {
 
   pub.existingSelect = function (num, nid, path, toHide, hideOrDisable, showEmpty, cid, fetch, defaults) {
     if (cid.charAt(0) === '-') {
-      resetFields(num, nid, true, 'show', toHide, hideOrDisable, showEmpty, 500);
+      resetFields(num, nid, true, 'show', toHide, hideOrDisable, showEmpty, 500, defaults);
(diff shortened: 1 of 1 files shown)


colemanw:
@jitendrapurohit what do you think about this?

jitendrapurohit:
@colemanw Confirm that this fixes the issue described above. This is a good change OR an extension of https://github.com/colemanw/webform_civicrm/pull/102 which also fixes the reset of defaults on the new contact creation.

jitendrapurohit (approved):

PR #152, opened 2018-07-30
Add description to the params sent to the processor

jitendrapurohit:
Overview
Add `description` to the param sent to the payment processor.

Before
Exception thrown by GoCardless processor for missing required params.

After
Fixed.

Technical Details
Either this can be fixed in the GoCordless extension or here, but I see `description` param is set in different functions in the same file and also in core civicrm. It was missing in `submitIPNPayment()` function which is now added in this PR.

Details - https://chat.civicrm.org/civicrm/pl/d11tnxs4wfyhjkbibqy8czowac


colemanw:
Seems harmless. I wish the payment processor extensions were more standardized though.

PR #154, opened 2018-08-08
Fix Case ID in tokens

mattwire:
Overview
Case ID does not get set because there is a typo.

Before
Case ID is not set in tokens (eg [submission:case-id:1])

After
Case ID is set in tokens (eg [submission:case-id:1])

Technical Details
Typo `$caseId` != `$caseID`


colemanw:
Agree there was no need for the 2 separate lines.

PR #155, opened 2018-08-10
Save credit card type and last 4 digits

eruraindil:
Overview
Save additional parameters about credit cards (type and last 4 digits) when processing a credit card after version 4.7.17 (this was implemented in https://issues.civicrm.org/jira/browse/CRM-20158)

I don't know if checking for credit card numbers is required, but better safe than sorry (I have never tried to use webform_civicrm with direct debit, so I don't know what it would do.)

Before
image

After
image

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2018,6 +2018,23 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       $params['payment_processor'] = wf_crm_aval($params, 'payment_processor_id');
     }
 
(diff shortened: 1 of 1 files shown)


KarinG:
Thanks Matthew - @mattwire: any concerns re: Stripe for this?

mattwire:
This looks fine.  It's just adding additional parameters to pass if they are available and everything is wrapped in conditionals.
My only comment would be that I think the whole code block in this PR should be wrapped in it's own function `contributionParamsCreditCard()` or similar as that will help maintainability in the future.

colemanw:
Sounds good. Thanks @eruraindil - congrats on your first PR to this project. Looking forward to many more in the future :)

PR #156, opened 2018-08-24
Add orig_contact_id to params in processCases if it's missing

AsylumSeekersCentre:
Add orig_contact_id to params in processCases function if it's missing and there are multiple Contacts on the Case

Overview
When there are multiple Contacts on one Case, updating custom fields through a webform fails. This change allows these custom field updates to go through.

Before
Submitting a webform which updates the custom fields of a Case (where the Case has multiple Contacts) would appear to work, but the custom fields were not updated, and this error would appear in the Drupal log:
`The CiviCRM "case create" API returned the error: "Case is linked with more than one contact id. Provide the required params orig_contact_id to be replaced" when called by function "postSave" on line 222 of wf_crm_webform_postprocess.inc with parameters: "Array ( [medium_id] => 1 [subject] => Subject 0 [client_id] => 12528 [id] => 58 [check_permissions] => [version] => 3 ) "`

After
Submitting the webform updates the Case custom fields correctly, even when there are multiple Contacts on the Case.

Technical Details
This code is quite specific to the parameters available in our use case, and I don't know how well it will generalise. It uses the content of the client_id param (if available) to populate orig_contact_id (if that param is empty). If client_id is not available, it does not populate orig_contact_id and the error above is likely to occur again (I don't know if it's actually possible for client_id to be empty in this situation).

Comments
I attempted to discuss this issue here:
[https://www.drupal.org/project/webform_civicrm/issues/2933134](https://www.drupal.org/project/webform_civicrm/issues/2933134)
In doing so and editing my post, I seem to have triggered some sort of spam prevention, and now I get a 403 error when I try to view that page. I don't know whether my post went through at all, so I'm posting the proposed change here.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1283,6 +1283,14 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           // These params aren't allowed in update mode
           unset($params['creator_id'], $params['case_type_id']);
         }
+        // Add orig_contact_id if it's missing and there are multiple
+        // Contacts on the Case. Use client_id as orig_contact_id if possible.
+        if (empty($params['orig_contact_id']) and !empty($params['id'])) {
+          $caseContactIds = CRM_Case_BAO_Case::retrieveContactIdsByCaseId($params['id']);
+          if (count($caseContactIds) > 1 and !empty($params['client_id'])) {
+            $params['orig_contact_id'] = $params['client_id'];
+          }
+        }
         // Allow "automatic" status to pass-thru
         if (empty($params['status_id'])) {
           unset($params['status_id']);


colemanw:
@AsylumSeekersCentre based on the error message, it seems the api thinks we are changing the case clients, so I wonder if *removing* the `client_id` param would also fix the problem?

AsylumSeekersCentre:
Apologies for the very long delay in answering.

I think this pull request should be closed without merging, but I've added some information from my latest testing of the issue. The problem still occurs, but seems to be solved by a much smaller change.

I have just upgraded our copy of webform_civicrm on a test system, and confirmed that I can reproduce the issue described above. However, the patch I previously suggested does not solve the problem, and instead the error message changes to `API returned the error: "Invalid case contact id (orig_contact_id)"`. Another change suggested by TwoMice in the discussion below does solve the issue for the use case I have just tested:

https://www.drupal.org/project/webform_civicrm/issues/2933134

That change alone (unsetting contact_id and client_id in the else block) allows our webform to operate as expected.

Edit: This seems consistent with your suggestion to remove client_id rather than add orig_contact_id.

### Reproducible issue description and mitigation

Drupal version: 7.72
CiviCRM version: 5.27.5
Webform version: 7.x-4.23
webform_civicrm version: 7.x-5.1

Details of the test:

We have multiple clients per case enabled, and a Case Type called "Client File".

We have a custom fieldset applied to Client Files with a number of fields, including "Medicare" and "Medicare expiry date".

I have created a webform which can update these Medicare fields on existing Client Files. The Case ID and Contact ID are passed to the webform like so:

`?cid2=34231&case1=20342`

If the Case has only one Client, the webform functions correctly.

With two Clients on the Case, the webform fails to update the Medicare fields, and this error appears in the log (as before):

`The CiviCRM "case create" API returned the error: "Case is linked with more than one contact id. Provide the required params orig_contact_id to be replaced" when called by function "postSave" on line 219 of wf_crm_webform_postprocess.inc with parameters: "Array ( [client_id] => Array ( [0] => 34231 ) [medium_id] => 9 [id] => 20342 [check_permissions] => [version] => 3 ) "`

With the patch I previously suggested, it fails with this error message:

`The CiviCRM "case create" API returned the error: "Invalid case contact id (orig_contact_id)" when called by function "postSave" on line 219 of wf_crm_webform_postprocess.inc with parameters: "Array ( [client_id] => Array ( [0] => 34231 ) [medium_id] => 9 [id] => 20342 [orig_contact_id] => Array ( [0] => 34231 ) [check_permissions] => [version] => 3 ) "`

With the change suggested by TwoMice (old patch file [here](https://www.drupal.org/files/issues/update_case.patch)), the fields are updated correctly and no errors appear in the log. I'm hoping to apply this to our production server on Monday, so I should find out in the following week whether there are any edge cases where that change is not enough.

colemanw:
Thanks @AsylumSeekersCentre - could you please put up a PR for your recommended fix?

PR #158, opened 2018-08-27
Allow filtering existing contacts by relationship

colemanw:
Overview
Adds a new filter option:

screenshot from 2018-08-27 14-42-59

Notes
-----
This filter should work with just about every webform configuration, including autocomplete lists, and multipage forms. Demo:

1. Create a new webform with 3 contacts.
2. Add an "Existing contact" field for all 3.
3. Edit the Contact 2 "Existing contact" field: set widget to "Select" and set a relationship filter to Contact 1.
4. Edit the Contact 2 "Existing contact" field: set widget to "Autocomplete" and set a relationship filter to Contact 2.
5. Try the form. You'll see that contact 2 has a dropdown populated by relations to the current logged in user. Changing that selection will update the autocomplete results for contact 3 in real-time.

Limitations
------
The one thing it won't do is update a `Select` list on-the-fly. E.g. if the Contact 3 widget was `Select` in the above scenario it wouldn't update the Contact 3 list in realtime. Workaround is to put a page break between contacts 2 and 3, or use autocomplete widget.

### DIFF
--- a/css/webform_civicrm_admin.css
+++ b/css/webform_civicrm_admin.css
@@ -193,10 +193,13 @@ form .civi-icon.case {
 
 /* Contact component form */
 #webform-component-edit-form div.form-item-extra-default,
-#webform-component-edit-form div.form-item-extra-default-relationship-to {
+#webform-component-edit-form div.form-item-extra-default-relationship-to,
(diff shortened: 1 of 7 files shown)


shitijg:
This worked - happy for it to be merged! Thanks!

PR #159, opened 2018-08-28
Contribution line item support.

KarinG:
Overview
Adding line item support

Before
You could only add one Contribution Amount

After
You can now add up to 5 additional line items:
image

Technical Details
First check-in for testing/review. Still requires some additional clean/tidy up.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -1060,6 +1060,10 @@ function wf_crm_get_fields($var = 'fields') {
       // KG - start
       // LineItems
       $sets['lineItems'] = array('entity_type' => 'line_item', 'label' => t('Line Items'));
+      $fields['contribution_line_total'] = array(
+          'name' => t('Line Item Amount'),
+          'set' => 'lineItems',
+        ) + $moneyDefaults;
       $fields['contribution_financial_type_id'] = array(
         'name' => t('Financial Type'),
         'type' => 'select',
@@ -1069,10 +1073,6 @@ function wf_crm_get_fields($var = 'fields') {
         'default' => 1,
         'set' => 'lineItems',
       );
-      $fields['contribution_line_total'] = array(
-          'name' => t('Line item Amount'),
-          'set' => 'lineItems',
-        ) + $moneyDefaults;
       // KG - end
       $sets['contributionRecur'] = array('entity_type' => 'contribution', 'label' => t('Recurring Contribution'));
       $fields['contribution_frequency_unit'] = array(
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -18,7 +18,6 @@
  * @return array
  */
 function wf_crm_field_options($field, $context, $data) {
-  // KG
   $ret = array();
   $fields = wf_crm_get_fields();
   if ($pieces = wf_crm_explode_key($field['form_key'])) {
@@ -517,7 +516,6 @@ function wf_crm_get_fields($var = 'fields') {
       'im' => array('entity_type' => 'contact', 'label' => t('Instant Message'), 'max_instances' => 9, 'custom_fields' => 'combined'),
       'activity' => array('entity_type' => 'activity', 'label' => t('Activity'), 'max_instances' => 30,  'attachments' => TRUE),
       'relationship' => array('entity_type' => 'contact', 'label' => t('Relationship'), 'help_text' => TRUE, 'custom_fields' => 'combined'),
-      // KG
       'lineItems' => array('entity_type' => 'line_item', 'label' => t('Line Items')),
     );
     $conditional_sets = array(
@@ -1060,9 +1058,7 @@ function wf_crm_get_fields($var = 'fields') {
         'name' => t('Contribution Source'),
         'type' => 'textfield',
       );
-      // KG - start
       // LineItems
-      //$sets['lineItems'] = array('entity_type' => 'line_item', 'label' => t('Line Items'));
       $fields['contribution_line_total'] = array(
           'name' => t('Line Item Amount'),
           'set' => 'lineItems',
@@ -1076,7 +1072,6 @@ function wf_crm_get_fields($var = 'fields') {
         'default' => 1,
         'set' => 'lineItems',
       );
-      // KG - end
       $sets['contributionRecur'] = array('entity_type' => 'contribution', 'label' => t('Recurring Contribution'));
       $fields['contribution_frequency_unit'] = array(
         'name' => t('Frequency of Installments'),
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -516,7 +516,6 @@ function wf_crm_get_fields($var = 'fields') {
       'im' => array('entity_type' => 'contact', 'label' => t('Instant Message'), 'max_instances' => 9, 'custom_fields' => 'combined'),
       'activity' => array('entity_type' => 'activity', 'label' => t('Activity'), 'max_instances' => 30,  'attachments' => TRUE),
       'relationship' => array('entity_type' => 'contact', 'label' => t('Relationship'), 'help_text' => TRUE, 'custom_fields' => 'combined'),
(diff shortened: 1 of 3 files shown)


includes/utils.inc
```diff
@@ -1057,6 +1057,23 @@ function wf_crm_get_fields($var = 'fields') {
         'name' => t('Contribution Source'),
         'type' => 'textfield',
       );
+      // KG - start
+      // LineItems
+      $sets['lineItems'] = array('entity_type' => 'line_item', 'label' => t('Line Items'));
```
colemanw: I feel like the entity type should be 'contribution' but not 100% sure.

colemanw: Also feel like this line belongs at the top with the other line items.

includes/wf_crm_admin_form.inc
```diff
@@ -1031,6 +1030,37 @@ class wf_crm_admin_form {
         }
       }
     }
+    // KG - start
+    // LineItem
+    $c = 1;
```
colemanw: I don't think we need this variable. There's only one contribution. Ever.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1513,6 +1513,25 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         'entity_table' => 'civicrm_contribution',
       );
     }
+    // KG - start
+    // LineItems
+    $fid = "civicrm_1_lineitem_1_contribution_line_total";
+    if (isset($this->enabled[$fid])) {
+      foreach ($this->data['lineitem'][1]['contribution'] as $l => $lineitem) {
+        $fid = "civicrm_1_lineitem_{$l}_contribution_line_total";
```
colemanw: `$1`?

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1513,6 +1513,25 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+        if ($this->getData($fid) > 0) {
+          $this->line_items[] = array(
+            'qty' => 1,
+            'unit_price' => $lineitem['line_total'],
+            'financial_type_id' => $lineitem['financial_type_id'],
+            'label' => wf_crm_aval($this->node->webform['components'], $this->enabled[$fid] . ':name', t('Line item')),
+            'element' => "civicrm_1_lineitem_{$l}",
```
colemanw: I don't think that's a valid name for a variable.

colemanw (commented):
This is shaping up. I think it's all going in the right direction, just needs a little cleanup.

colemanw:
Also, one more useful field for each line_item set would be "Line Item Name".

includes/utils.inc
colemanw: And should be conditional with CiviContribute being enabled :)

KarinG:
Feedback from Jake: "Tested - and looking perfect!."

Thank you for your comments/feedback @colemanw - I'll try finalize this either tonight or tomorrow morning.

includes/wf_crm_webform_postprocess.inc
KarinG: Ah - that's actually a letter lower case 'l' (not a number 1); we're using $c for contact and $n in other places - I used $l for the lineitem number. I'll make that $n (easier to read).

includes/wf_crm_webform_postprocess.inc
KarinG: See above - will change that to $n so it is consistent with what we have in the membership section: 

'element' => "civicrm_{$c}_membership_{$n}",

includes/utils.inc
KarinG: I'll have another look at the entity type [I did start off with 'contribution'. Not sure what made me change it]; and yes I can move that Set + make it conditional.

includes/wf_crm_admin_form.inc
KarinG: I kept it thinking we may want to at some point have more than 1 contribution - but yeah - will simplify.

KarinG:
Re: Line Item Name -> I modeled it after "Contribution Amount" -> to change that Name/Label -> all one needs to do Edit element Line Item Amount -> and that will end up in the CiviCRM Contribution:

For Example:

image

image

image

includes/wf_crm_webform_postprocess.inc
KarinG: Ok - done

includes/wf_crm_webform_postprocess.inc
KarinG: Ok - done

includes/utils.inc
KarinG: Right now the field keys are lineitem n of the contribution:
civicrm_1_lineitem_n_contribution_line_total
civicrm_1_lineitem_n_contribution_financial_type_id
[and that automatically pulls in e.g. all contribution financial types available which is what we want]

I can change the entity from lineitem to contribution - but then we loose the lineitem reference and the keys would be very similar to:
civicrm_1_contribution_1_contribution_total_amount 
(which is the already existing field you can create by checking the box for 'Contribution Amount')

Probably best to stick with
civicrm_1_lineitem_n_contribution_line_total
civicrm_1_lineitem_n_contribution_financial_type_id
to avoid confusion

KarinG: Also if $ent becomes 'contribution' then the number_of suggests we have more than one contribution. We need to be able to store number of lineitems:

elseif ($ent == 'grant' || $ent == 'activity' || $ent == 'case' || $ent == 'lineitem') {
  $this->data[$ent]["number_$key"] = $val;

KarinG: I was able to move the Set to the top though - working on making it conditional now.

KarinG: Almost done. Easier if it was $ent = 'contribution' - going to have one more pass at looking into that.

colemanw:
I've done another code review and think it all looks peachy.
Would you mind removing the `// KG` comments before merging? `Git blame` will show you as the author of those lines anyway.

KarinG:
Yes of course - but I wasn't going to do that until I had it all done :-)

And I have one more concern about $ent - so will have another pass at that. Right now the field_keys are ok - but not perfect (they nest under contact 1); and the lineItem set does not show until after you enable the Contribution section and after you reload the Config form. That bugs me. If it's $ent = 'contribution' it works - except then I don't get the financial_types loaded.

KarinG:
Ok - I just added a final clean up commit. I started seeing literally double :-) when I was reworking things from lineitem entity to contribution entity so decided to leave this as is for now. 

So to summarize two things to keep in mind:
1) lineItems when injected initially will nest under Contact 1 - of course they can be dragged into any position on the webform from there.
2) the lineItem sets only show on reload of the Config form after enabling a Contribution page. That's because the AJAX call on the enable Contribution page will render everything with contribution entity and we've got lineitem entity/sets.

Tested by me and Jake.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1513,6 +1513,23 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
       );
     }
+    // LineItems
+    $fid = "civicrm_1_lineitem_1_contribution_line_total";
+    if (isset($this->enabled[$fid])) {
+      foreach ($this->data['lineitem'][1]['contribution'] as $n => $lineitem) {
+        $fid = "civicrm_1_lineitem_{$n}_contribution_line_total";
+        if ($this->getData($fid) > 0) {
+          $this->line_items[] = array(
+            'qty' => 1,
```
colemanw, line 1523: We could potentially add a field for quantity.

KarinG: It crossed my mind too! But then we'd probably also have to store the priceset in CiviCRM tables - as people will really think these are pricesets now?

PR #160, opened 2018-08-30
Tally line items with javascript

colemanw:
Overview
Follow up on #159 - tallies line items on-the-fly without requiring a page break.

Before
Line items were only tallied properly if they were not on the last page of the webform.

After
Line items tallied on any page.

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -446,6 +446,11 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
           if (!empty($field['data_type'])) {
             $dt = $element['#civicrm_data_type'] = $field['data_type'];
           }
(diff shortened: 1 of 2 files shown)


colemanw:
@KarinG I decied to do this as a followup PR because it will need to be put through some testing. Esp with changing the widget type of contribution amount and line item fields to test that they work as checkboxes, radios, grids, etc.

KarinG:
Ah thank you! I’ll give it a good spin.

KarinG:
This is almost working. 

LineItems on p1:
image

If I move the LineItems to p2: [note the sales tax is not applied]
image

colemanw:
@KarinG if it works except for taxes I say we merge it and file a followup since the js never worked with taxes in the first place.

KarinG:
Perhaps we just need a warning bubble that says “If you’re using Financial Types associated with Sales Taxes you must put Contiribution and/or Membership fields on p1”?

PR #161, opened 2018-09-02
Making Financial Types for Membership configurable.

KarinG:
Overview
This PR makes it possible for admin to configure Financial Type for Membership on the Webform. It emulates how one can do that natively in CiviCRM with Pricesets. 

Before
Financial Type was always taken directly from the Membership Type Configuration in CiviCRM: 
`$membership_financialtype = $this->getMembershipTypeField($type, 'financial_type_id');`

After
Now that Membership Financial Type is configurable - the Webform can be configured to automatically add the Financial Type that has the correct Sales Tax (GST) associated with it - based on the Province the Member lives in - using native Webform conditionals. [Note: could not get the conditionals to work with AJAX select so edited the widget to select options]

image

image

Technical Details
Modeled after participant status (with - Automatic - option as default)
Default behaviour (if no Membership Financial Type on the form) is to take Financial Type directly from the Membership Type Configuration in CiviCRM.

Comment
Will be testing this live on one of our projects this week.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1547,11 +1547,16 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
               }
             }
 
+            $membership_financialtype = $this->getMembershipTypeField($type, 'financial_type_id');
+            if (isset($item['financial_type_id']) && $item['financial_type_id'] !=0 ) {
+              $membership_financialtype = $item['financial_type_id'];
+            };
+
             if ($price) {
               $this->line_items[] = array(
                 'qty' => $item['num_terms'],
                 'unit_price' => $price,
-                'financial_type_id' => $this->getMembershipTypeField($type, 'financial_type_id'),
+                'financial_type_id' => $membership_financialtype,
                 'label' => $this->getMembershipTypeField($type, 'name') . ": " . wf_crm_display_name($this->existing_contacts[$c]),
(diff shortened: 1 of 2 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1547,11 +1547,16 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
               }
             }
 
+            $membership_financialtype = $this->getMembershipTypeField($type, 'financial_type_id');
+            if (isset($item['financial_type_id']) && $item['financial_type_id'] !=0 ) {
+              $membership_financialtype = $item['financial_type_id'];
+            };
```
colemanw, line 1553: You could call `getMembershipTypeField` in an `else` condition here to save calling it for no reason. That's my only nitpick :)

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1547,11 +1547,16 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
               }
             }
 
+            $membership_financialtype = $this->getMembershipTypeField($type, 'financial_type_id');
+            if (isset($item['financial_type_id']) && $item['financial_type_id'] !=0 ) {
```
colemanw, line 1551: One more little nitpick, this line could be better written as `if (!empty($item['financial_type_id'])) {`

PR #165, opened 2018-09-21
Make sure billing parameters are formatted correctly when submitting via a redirect payment processor

mattwire:
Overview
All built-in CiviCRM forms that take payment pass billing params through in the format (eg) billing_state_province-5 rather than just state_province before submitting to a (redirect) payment processor.  Webform_civicrm does not, and this causes problems if the payment processor expects the address to be in the format eg `billing_state_province-5` instead of `state_province`.

This was found while integrating Omnipay Sagepay with webform_civicrm.

Before
Cannot submit payment via a "redirect" payment processor that has been written to expect billing address parameters to be in the "standard" format.

After
Can submit payment via a "redirect" payment processor such as Omnipay Sagepay.  For any processors that are clever enough to look for address fields with different keys there will be no change.

Technical Details
Merge $params with $_REQUEST which contains the required parameters in the expected format.

Comments
@eileenmcnaughton @pradnayak

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1915,6 +1915,14 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params['source'] = $this->settings['new_contact_source'];
     $params['item_name'] = $params['description'] = t('Webform Payment: @title', array('@title' => $this->node->title));
 
+    // We need the submitted billing params from the request (Eg. billing_country_id-5)
+    //  because that's how payment processors expect to find them.
+    foreach ($this->form_state['input'] as $formKey => $formValue) {
+      if (substr($formKey, 0, 8) === 'billing_') {
+        $params[$formKey] = $formValue;
+      }
+    }
+
     if (method_exists($paymentProcessor, 'setSuccessUrl')) {
       $paymentProcessor->setSuccessUrl($this->getIpnRedirectUrl('success'));
       $paymentProcessor->setCancelUrl($this->getIpnRedirectUrl('cancel'));


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1915,6 +1915,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params['source'] = $this->settings['new_contact_source'];
     $params['item_name'] = $params['description'] = t('Webform Payment: @title', array('@title' => $this->node->title));
 
+    // We need the submitted billing params from the request (Eg. billing_country_id-5)
+    //  because that's how payment processors expect to find them.
+    $params = array_merge($params, $_REQUEST);
```
eileenmcnaughton: Just a  hot take - looks like a potential security risk merging 'anything the user wants to input' into params

mattwire: I've changed it to:
```
    foreach ($this->form_state['input'] as $formKey => $formValue) {
      if (substr($formKey, 0, 8) === 'billing_') {
        $params[$formKey] = $formValue;
      }
    }
```
@eileenmcnaughton Is that better?

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1915,6 +1915,14 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params['source'] = $this->settings['new_contact_source'];
     $params['item_name'] = $params['description'] = t('Webform Payment: @title', array('@title' => $this->node->title));
 
+    // We need the submitted billing params from the request (Eg. billing_country_id-5)
+    //  because that's how payment processors expect to find them.
+    foreach ($this->form_state['input'] as $formKey => $formValue) {
+      if (substr($formKey, 0, 8) === 'billing_') {
+        $params[$formKey] = $formValue;
+      }
+    }
```
colemanw, line 1924: I didn't realize Drupal would stick random other stuff in `$form_state['input']` - note that billing fields are just markup on the form they are not actual FAPI fields.
But if this actually works then I approve :)

colemanw (approved):

mattwire:
@KarinG @colemanw Is this ok to merge now?

colemanw:
@KarinG does this look good to you?

KarinG:
Let me QA once more - I think I did but want to be sure. Will do that right now.

KarinG:
Ok:
- this works!
- indeed no harm for any payment processors clever enough to already handle this themselves
- however there are many places in CiviCRM where we always need to check what form we're on so this feels like a workaround
- but then I'm all for getting things to work - and if this helps people payment process with Webform CiviCRM - then it's a good thing
- @colemanw if you agree - then hit Merge!

PS - this (below) is public Contribution vs public Event - I use this bit for all extensions that are trying to have to do something _validateForm - everything is always a little different in different places in CiviCRM 😄 

```
    if ($formName == 'CRM_Contribute_Form_Contribution_Confirm') {
      $financialTypeID = $form->_values['financial_type_id'];
      $totalAmount = $form->_amount;
      $emailPrimary = $form->_defaultValues['email-5'];
    }
    else {
      // we must have an Event form
      $financialTypeID = $form->_values['event']['financial_type_id'];
      $totalAmount = $form->_totalAmount;
      $emailPrimary = $form->_defaultValues['email-Primary'];
    }
```

PR #169, opened 2018-10-15
Fix default selection of multi-select element

jitendrapurohit:
Overview
Multi-select option does not stick on the form when we re-edit.

Before
To reproduce -

1. Create a webform -> Click on Activities tab.
2. Toggle the multi-select field `Assign Activity to` to single select and save `none` option for the input field.

image

2. Save the form and try to edit again. The field selection remains the default one and the setting saved in the above step is lost. 

image

After
The value `none` is saved and retrieved correctly on editing the webform.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1346,6 +1346,9 @@ class wf_crm_admin_form {
         $options = array_keys($options);
         $item['#default_value'] = $options[1];
       }
+      elseif (!empty($this->data[$ent][$c][$table][$n]) && empty($this->data[$ent][$c][$table][$n][$name]) && $item['#default_value'] == 0) {
+        $item['#default_value'] = NULL;
+      }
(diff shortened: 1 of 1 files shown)


yorkshirerose:
Had this problem on two sites:
One is running CiviCRM 5.6.1, Drupal 7.60 and Drupal Webform integration v7-4.23 (not the most recent update of this yet) and the other is CiviCRM 5.7.0 (Drupal 7.59) and Webform CiviCRM integration module 7-4.24. 
It would only allow 'user select' or a contact for 'Assign to' on each activity, and the option to say 'None' would disappear whenever the webform was edited.
Made this change to the file and it looks like that's fixed it - I can't replicate the bug anymore and the drop down for 'Assign to' is behaving predictably now (i.e. I can leave it set to 'None').

KarinG:
Thank you Jitendra and Rose - merging this

PR #170, opened 2018-10-17
Load default location type for address/phone/email configured in civi

jitendrapurohit:
Overview
Load default location type for address/phone/email as per civi configuration.

Before
On webform - default location type is loaded as `Home` even if `Main` is selected as default in `civicrm/admin/locationType?reset=1`

image

After
Default loc type is loaded according to the default configured in civi.


colemanw:
Looks good to me

PR #171, opened 2018-10-19
Prefer case relationship if case id is present in the URL

jitendrapurohit:
Overview
Prefer case relationship to load on a contact if caseid is present in the URL. 

Before
Eg - 

- A has multiple "Referring Officer" relationship with contact B and C.

- The relationship of Contact C is attached to case id 1.

- webform contact input field has a default value set as below -

image

- Now, if a webform is loaded with `case1=1` param in the URL, it loads the contact which is related to contact A, but do not actually prefer the case related contact, which might make more sense to load as a default one?

After
If there are multiple relationship with a contact - load the case related contact if there is a case id in the URL.

Comments
If this doesn't seem to be a safe fix, should we just add a setting on the component page? eg `Load case related contact if case id is present in the URL`.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -881,7 +881,7 @@ function wf_crm_find_relations($cid, $types = array(), $current = TRUE) {
     if ($type_ids) {
       $typeClause = "AND relationship_type_id IN ($type_ids)";
     }
-    $sql = "SELECT relationship_type_id, contact_id_a, contact_id_b
+    $sql = "SELECT relationship_type_id, contact_id_a, contact_id_b, case_id
       FROM civicrm_relationship
       WHERE (contact_id_a = $cid OR contact_id_b = $cid) $typeClause";
     if ($current) {
@@ -893,6 +893,10 @@ function wf_crm_find_relations($cid, $types = array(), $current = TRUE) {
       if (!$allowed || in_array($dao->relationship_type_id . '_' . $a_b, $allowed)) {
         $c = $dao->{"contact_id_$a_b"};
         $found[$c] = $c;
+        if ($type_ids && !empty($dao->case_id) && !empty($_GET["case1"]) && $_GET["case1"] == $dao->case_id) {
+          //Make this contact as a first value in the array.
+          $found = array_merge([$c => $c], $found);
+        }
       }
     }
     $dao->free();


includes/contact_component.inc
```diff
@@ -893,6 +893,10 @@ function wf_crm_find_relations($cid, $types = array(), $current = TRUE) {
       if (!$allowed || in_array($dao->relationship_type_id . '_' . $a_b, $allowed)) {
         $c = $dao->{"contact_id_$a_b"};
         $found[$c] = $c;
+        if ($type_ids && !empty($dao->case_id) && !empty($_GET["case1"]) && $_GET["case1"] == $dao->case_id) {
```
colemanw, line 896: If we were in object context then I would recommend using `$this->data['case'][1]` instead of reaching into `$_GET` but I don't think cases have actually been loaded yet.
```suggestion
        // Prefer case roles over other relationships
        if ($type_ids && !empty($dao->case_id) && wf_crm_aval($_GET, 'case1') == $dao->case_id) {
```

colemanw:
@jamienovick - FYI I think this relates to some other work you were doing.

shitijg:
@jitendrapurohit @colemanw  I think this might override the "Randomize" checkbox that we have as we are enforcing some logic to pick a contact as a default value when there are more than one to select from.

If you're open to let the user select on the webform - another way of doing this is:
In the default value - 
You can set "Set default contact from" = Auto from filters

In the filters section, you can do the same thing as above through the "Relationships to" dropdown. 

You can set the widget at the top as "Autocomplete" or "Select from List"

shitijg:
@jitendrapurohit Also - just to add to that - we're adding "Case Relationship"  as an option in the default value section. It's passed testing for us internally - we'll add a PR to webform_civicrm shortly. 

Hope this solves your use case. 
Thanks,
Shitij

includes/contact_component.inc
jitendrapurohit: yes, cases are not loaded as of now, hence fetching it the way it is done to set case objects.

jitendrapurohit:
Hi @shitijg 

>If you're open to let the user select on the webform - another way of doing this is:

Ah no, we really want the case relationship to load by default along with all the case information.

>@jitendrapurohit Also - just to add to that - we're adding "Case Relationship" as an option in the default value section. It's passed testing for us internally - we'll add a PR to webform_civicrm shortly.

Sure, that sounds like similar to our usecase. Happy to test when the PR is done :-)

shitijg:
@jitendrapurohit https://github.com/colemanw/webform_civicrm/pull/183

jitendrapurohit:
closing in favor of #183

PR #174, opened 2018-10-22
Use Request object and fix called to undefined check_plain

mglaman:
Overview
Improves checking `cid` from the URL and an error when a message is shown for the "not you?" message.

Technical Details
When loading an existing contact from the URL, raw _GET is used instead of the request object. This fixes that, along with a called to the now undefined check_plain

Comments

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -6,8 +6,11 @@
  */
 
 use Drupal\Component\Serialization\Json;
+use Drupal\Component\Utility\Html;
 use Drupal\Component\Utility\NestedArray;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Link;
+use Drupal\Core\Render\Markup;
 use Drupal\Core\Url;
 use Drupal\webform\Plugin\WebformHandlerInterface;
 
@@ -793,7 +796,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
    * @param string $type
    */
   function setMessage($message, $type='status') {
-    if (node_is_page($this->node) && empty($_POST)) {
+    if (empty($_POST)) {
       drupal_set_message($message, $type, FALSE);
     }
   }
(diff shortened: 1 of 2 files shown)


includes/wf_crm_webform_preprocess.inc
```diff
@@ -810,17 +813,18 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
     $message = $this->replaceTokens($message, $contact);
     preg_match_all('#\{([^}]+)\}#', $message, $matches);
     if (!empty($matches[0])) {
-      $q = $_GET;
-      unset($q['q'], $q['cs'], $q['cid'], $q['cid1']);
-      if (empty($_GET['cid']) && empty($_GET['cid1'])) {
-        $q['cid1'] = 0;
+      $query = \Drupal::request()->query;
+      if (empty($query->get('cid')) && empty($query->get('cid1'))) {
+        $query->set('cid1', 0);
```
dsnopek: Is this right? The original code was only changing the local `$q['cid1']` variable, but the new code is changing 'cid1' on the request object (or at least appears to be - perhaps the query is copied? Doesn't look like it, though).

dsnopek: It also seems like it's missing the `unset()`'s too. If those are actually ever passed, this would lead to different URLs being generated.

Maybe it should be doing something like:

```
$q = \Drupal::request->query->all()
```

And then using `$q` same as it did in the old code?

includes/wf_crm_webform_preprocess.inc
```diff
@@ -843,7 +847,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
       if (is_array($value)) {
         $value = implode(', ', $value);
       }
-      $values[] = implode(' &amp; ', wf_crm_explode_multivalue_str(check_plain($value)));
+      $values[] = implode(' &amp; ', wf_crm_explode_multivalue_str(Html::escape($value)));
```
dsnopek: Given all the Twig autoescape stuff that happens in D8, I'd worry that this would lead to double escaping. Should this maybe just plain strings and rely on the Twig automagic?

mglaman: You're right.

* https://www.drupal.org/project/drupal/issues/2089331
* https://www.drupal.org/node/2549395

Deprecation message;

```
Will be removed before Drupal 9.0.0. Rely on Twig's auto-escaping feature, or use the #plain_text key when constructing a render array that contains plain text in order to use the renderer's auto-escaping feature. If neither of these are possible, \Drupal\Component\Utility\Html::escape() can be used in places where explicit escaping is needed.
```

includes/wf_crm_webform_preprocess.inc
mglaman: That works, and makes more sense. I'll revise.

mglaman:
@dsnopek updated to address the review.

mglaman:
FWIW this is a blocker to memberships (at least if signed in)

```
The website encountered an unexpected error. Please try again later.
TypeError: Argument 1 passed to node_is_page() must be an instance of Drupal\node\NodeInterface, instance of Drupal\webform\Entity\Webform given, called in /app/web/modules/contrib/webform_civicrm/includes/wf_crm_webform_preprocess.inc on line 796 in node_is_page() (line 512 of core/modules/node/node.module).
```

dsnopek:
The fixes look good to me! :+1:

dsnopek:
@KarinG I think this one is ready to commit :-)

PR #175, opened 2018-10-23
Move wf_crm_field_options to a service

mglaman:
Overview
This helps begin the refactoring process and cleanup.

Technical Details
There is a lot of D7 code, and this helps to start to bring this into D8 and hopefully streamline some of those code with what D8/OOP gives up.


dsnopek:
This is a great idea, this totally makes sense as a service! The `\Drupal\webform_civicrm\FieldOptions::get()` method is still a monster and should be refactored into more, smaller methods at some point, but that doesn't need to happen in this PR - in fact, it's probably better this way so we're not worried that any of the code got broken in the move.

I haven't tested, but :+1: to a skim through the code!

mglaman:
I didn't want to refactor it too much yet. But I did want to start cleaning things up to make moving forward more easier.

dsnopek:
@KarinG I think this is fine to merge! It's just moving code

KarinG:
Excellent! Thank you @mglaman and @dsnopek

PR #176, opened 2018-10-24
Fixup getExposedOptions

mglaman:
Overview
As part of the work in #174 to get the membership form displaying, the getExposedOptions method needs to be updated. Otherwise, there are no exposed options and PHP notices abound.

Technical Details
This updated the method's logic for new array keys, like `#extra` as things are being remapped.

Comments
This + #174 unblock the membership select


dsnopek:
@mglaman Is there a way to test this without the membership changes? If not, maybe this is better just in with those changes rather than as a separate PR?

mglaman:
@dsnopek no, there isn't. I'll keep this part of those changes.

PR #177, opened 2018-10-26
Support free membership

mglaman:
Overview
This adds a `civicrm_select` element that will hold the `civicrm_live_options` and static options logic. It also fixes element placement in the Webform so we can properly use WebformElement plugins.

Deprecates #174 which contained part of these fixes, but in a smaller delivery.

Technical Details

* Instead of having to hack in `#webform_civicrm` to know the form key, it is properly set
* We can use WebformElement/Element plugins appropriately, so we can begin to remove the preProcess,postProcess,formAlter, etc methods.
* Displays membership options
* Adds membership
* Allows setting as static or live
* Customizing of the option values
* Disabling options


KarinG:
@mglaman - thank you! @dsnopek - let me know when you’ve had a chance to look at this (and Matt’s other PRs); then I’ll give it a spin as well!

mglaman:
@KarinG want me to cherry-pick that commit into its own PR, since this one may take a while to land?

KarinG:
@mglaman - Yes please - good idea.

mglaman:
@KarinG okay, I split that commit out to #179

mglaman:
Okay, this ready for some testing. You can configure the membership select to:

* Sort the options
* Disable options
* Change the label of options

To do:

- [ ] Respect the sorting when re-editing the element
- [ ] Support the "Default" option

src/Plugin/WebformHandler/CivicrmWebformHandler.php
```diff
@@ -79,6 +79,7 @@ public function defaultConfiguration() {
   }
 
   public function alterElements(array &$elements, WebformInterface $webform) {
+    \Drupal::getContainer()->get('civicrm')->initialize();
```
dsnopek: This is a plugin - let's inject the 'civicrm' service instead (although, definitely call the `initialize()` here where it's needed)

src/Plugin/WebformHandler/CivicrmWebformHandler.php
```diff
@@ -91,6 +92,11 @@ public function alterForm(array &$form, FormStateInterface $form_state, WebformS
     $processor->alterForm();
   }
 
+  public function postLoad(WebformSubmissionInterface $webform_submission) {
+    parent::postLoad($webform_submission); // TODO: Change the autogenerated stub
```
dsnopek: Seems like this turned out not to be needed?

dsnopek:
I created a new membership form, where the membership type was "user select" triggering the addition of the "CiviCRM Select" element. At first, I had it set to Live Options, saved that, and experimented with it. Then I set it to Static Options, and messed with that a bit, saved it. Aside from the items called out in Matt's "To do" above, all worked as expected.

However, the resulting Yaml looked a little off:

```
civicrm_1_membership_1_membership_membership_type_id:
  '#type': civicrm_select
  '#title': 'Membership Type'
  '#civicrm_live_options': '0'
  '#options':
    2: Normalny
  '#extra':
    civicrm_live_options: 1
  '#form_key': civicrm_1_membership_1_membership_membership_type_id
```

Notice the '#civicrm_live_options' key at the top-level of this element, and also the 'civicrm_live_options' under '#extra'. This is, in fact, using Static Options. I think the key under '#extra' must have gotten set accidentally earlier on, and then was left there?

includes/utils.inc
```diff
@@ -1055,8 +1055,9 @@ function wf_crm_get_fields($var = 'fields') {
     if (isset($sets['membership'])) {
       $fields['membership_membership_type_id'] = array(
         'name' => t('Membership Type'),
-        'type' => 'select',
+        'type' => 'civicrm_select',
         'expose_list' => TRUE,
+        'civicrm_live_options' => 1,
         'extra' => array('civicrm_live_options' => 1),
```
dsnopek: Hm, maybe this is the source of the weird key in the Yaml?

includes/wf_crm_admin_form.inc
```diff
@@ -1757,36 +1759,45 @@ class wf_crm_admin_form {
...
           '#type' => $enabled_element['type'],
-          '#webform_civicrm' => $enabled_key,
-          // Webform uses YAML dump, which dies on FormattableMarkup.
-          '#extra' => array_map([get_called_class(), 'stringifyFormattableMarkup'], empty($enabled_element['extra']) ? [] : $enabled_element['extra']),
-          '#weight' => empty($enabled_element['weight']) ? 0 : $enabled_element['weight'],
-        ];
+        ]);
+
+        $stub_form = [];
+        $stub_form_state = new FormState();
```
dsnopek, line 1779: Should this maybe use `SubformState`? Or, is it really completely independent of the parent form?

If this was using `SubformState`, I think the incantation is something like:

```
$stub_form_state = SubformState::createForSubForm($stub_form, $form, $form_state);
```

src/Element/CivicrmSelectOptions.php
```diff
@@ -0,0 +1,178 @@
+<?php
+
+namespace Drupal\webform_civicrm\Element;
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Render\Element\FormElement;
+
+include_once __DIR__ . '/../../includes/utils.inc';
```
dsnopek, line 10: What's this `include_once` about?

src/Plugin/WebformElement/CivicrmSelect.php
```diff
@@ -0,0 +1,136 @@
...
+        'value' => '',
+        'empty_option' => '',
+        'empty_value' => '',
+        'options' => [],
+        'options_randomize' => FALSE,
+        'expose_list' => TRUE,
+        'exposed_empty_option' => '- ' . t('Automatic') . ' -',
+        'civicrm_live_options' => 1,
+        'extra' => [
+          'civicrm_live_options' => 1,
```
dsnopek: Or maybe _this_ is where that extra bit in the Yaml is from?

src/Plugin/WebformElement/CivicrmSelect.php
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $form_state) {
+    $form = parent::form($form, $form_state);
+
+    // Get element properties.
+    $element_properties = $form_state->getValues() ?: $form_state->get('element_properties');
+    $wtf_values = $form_state->getValues();
```
dsnopek: `$wtf_values`? :-) I'm assuming that was meant to be `$wf_values` but it's not used anywhere anyway

src/Plugin/WebformElement/CivicrmSelect.php
```diff
@@ -0,0 +1,136 @@
...
+
+    $form['options']['civicrm_live_options'] = [
+      '#type' => 'radios',
+      '#options' => [
+        t('<strong>Static Options</strong> (fully configurable)'),
+        t('<strong>Live Options</strong> (update automatically)'),
+      ],
+      '#description' => Markup::create('<p><div class="live-options-hide">' .
+        t('Drag the arrows to re-order these options. Click the "enabled" checkbox to show/remove an item from the form. Set the label as you want it to appear on the form.') .
+        '</div><div class="live-options-show">' .
```
dsnopek: I'm assuming there's some Javascript that's meant to hide/show these bits based on the 'live-options-hide' and 'live-options-show' classes). It's not working, though. The current version just always shows both of these div's.

src/Plugin/WebformElement/CivicrmSelect.php
mglaman: :) Bad debugging statement left behind.

src/Plugin/WebformElement/CivicrmSelect.php
mglaman: The `extra` can be blown away, now. That was from the D7 work and has now been migrated completely into the element property `civicrm_live_options`

src/Element/CivicrmSelectOptions.php
mglaman: There are functions used in `utils.inc`, since helpers haven't been ported to services, yet.

includes/wf_crm_admin_form.inc
mglaman: I don't think it matters, here, because we just need a FormStateInterface object and values assigned. We we did a SubFormState I'd have to mock the form element parent arrays.

includes/utils.inc
mglaman: Like comment below, the `extra` is legacy and can be removed (yay!)

src/Element/CivicrmSelectOptions.php
dsnopek: Can you add a comment about this by these sort of includes?

includes/wf_crm_admin_form.inc
dsnopek: Ok! Just wanted to make sure we weren't missing something important, since SubFormState does a tiny bit of magic to really make the illusion of a subform work properly. I'll mark this one as resolved.

src/Element/CivicrmSelectOptions.php
mglaman: I added this format to all `include_once`

```
// Include legacy files for their procedural functions.
// @todo convert required functions into injectable services.
```

mglaman:
@dsnopek okay, initial review items addressed.

src/Plugin/WebformHandler/CivicrmWebformHandler.php
dsnopek: Changes look awesome! I can see you removed some other direct calls to pull this service from the container.

dsnopek:
This is looking awesome! All of my review is addressed and the two TODO items from last week are also working great in my testing.

However, I did find what appears to be a bug with live options:

If make a webform that adds a membership, where membership type is "User Select", and view the form, it shows the current membership types in the select. If I go into the CiviCRM backend and add a new membership type, and view the form again, it shows the new membership type too. All good!

But if I edit the membership type element on the webform and save it (with or without changing anything, and certainly leaving it as using "live options"), then the membership types on the form become set in stone. If I add a new membership type in the CiviCRM backend, it doesn't appear on the form, even after clearing caches. I think it's saving the option list, when it should probably just leave that as empty when using live options (and I can see on the YAML that it's saving the option list -- and, strangely, it's not marking when live options are turned on, only when turned off, which is kind of weird but maybe it's tracking differences from the default?). Otherwise, "live options" are action kind of like "static options" without any of the configuration, just a snapshot of the current when the element was edited.

mglaman:
@dsnopek this has been fixed. Essentially, if it had never been edited, there were no known `#options`, which caused it to act "Live". But, once saved it never would "pull fresh" for live. This has been addressed in the last commit.

dsnopek:
This is working in my testing! :+1: 

@KarinG We're all good from our end! This is ready for you to do any additional review/testing and merge :-)

KarinG:
Ok - merging! I'll ask some of our co-partners on this project to also have a look once I have the new -dev posted.

PR #178, opened 2018-10-26
Fix getCreditCardCSSNames() call which is deleted from core civi

jitendrapurohit:
Overview
Fix fatal error on the webform

Before
`getCreditCardCSSNames()` function is deleted from core civi in the PR merged yesterday https://github.com/civicrm/civicrm-core/pull/12615. The core civi PR defines two `protected` function in order to set the js in core. Webform civi still uses the old function call which leads to a fatal error in viewing any of the webform.

After
Replaced the deleted function with the one used in core. The fatal error is fixed.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -867,8 +867,18 @@ abstract class wf_crm_webform_base {
    * @return int|null
    */
   function addPaymentJs() {
-    CRM_Core_Resources::singleton()->addCoreResources();
-    CRM_Financial_Form_Payment::addCreditCardJs(NULL, 'html-header');
+    $currentVer = CRM_Core_BAO_Domain::version();
+    if (version_compare($currentVer, '5.8') <= 0 && method_exists('CRM_Core_Payment_Form', 'getCreditCardCSSNames')) {
+      $credit_card_types = CRM_Core_Payment_Form::getCreditCardCSSNames();
+      CRM_Core_Resources::singleton()
+        ->addCoreResources()
+        ->addSetting(array('config' => array('creditCardTypes' => $credit_card_types)))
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@jitendrapurohit this will be tricky across webform versions - when I do this in an extension I would increment the civi version in the info.xml file so that the change would only apply to the newer versions - I'm not too sure how it gets handled in webform though - presumably different versions of webform support different versions of civi

mattwire:
So both the old and new methods here are not ideal.  They both rely on specific versions of CiviCRM and unsupported access to methods.  We've hit similar issues to this with webform_civicrm before (eg. the switch from validateCreditCard to validatePaymentInstrument).
I'm not sure what is best though - maybe it would be possible to create a copy of this function in webform_civicrm so we don't depend on Civi for it?

eileenmcnaughton:
@mattwire I agree - I would say that I see calling validatePaymentInstrument as supported as we deliberately adopted that into the Payment class and the payment class is a supported interaction point with core - whereas I don't see this function usage as supported  - which probably strengthens your case for copying to core.

As I mentioned I deal with this in 'normal' extensions by incrementing the supported Civi version on a regular basis to one version of he extension isn't trying to work on all versions of Civi - but as a module I don't think webform supports that approach in a manageable way

KarinG:
@jitendrapurohit - thank you for catching this! How about a version of CiviCRM check and then dependent on the outcome -> use either old method or new method?

jitendrapurohit:
Not sure which version was the civi PR got merged in. I've updated it as per the [JIRA](https://issues.civicrm.org/jira/browse/CRM-21311) fix version and an additional check to see if the function exists in the class 👍

PR #179, opened 2018-11-01
Element contains ignored property weight

mglaman:
Overview
Fixes https://www.drupal.org/project/webform_civicrm/issues/3003926

Technical Details
This removes the legacy property being set: weight. Apparently, we cannot define these when building the CiviCRM elements.


KarinG:
Fixes https://www.drupal.org/project/webform_civicrm/issues/3003926 - thank you!

PR #181, opened 2018-11-10
Adding membership Auto-renewal support

omarabuhussein:
This is an alternative to this PR : https://github.com/colemanw/webform_civicrm/pull/133

Overview
Adding Membership Auto-renewal support to webform

Before
This module currently does not provide a way to auto-renew memberships.

After
The ability to set a membership to be auto-renewed is now available, you can either set the membership to be auto-renewed automatically, or not to be auto-renewed (which is the default option), or to allow the user to selected whither to set the membership to be auto renewed or not as u can see below : 

2018-04-27 13_59_31-pp test _ dmaster8

If the membership is set to be auto-renewed automatically, or if the end user selected the membership to be auto-renewed then a recurring contribution record will be created with auto_renew field set to TRUE, and the created memberships will have their contribution_recur_id field refer to the newly created recurring contribution ID.

Here is an example with one/no installments : 

And here is another one with 3 installments :

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -77,6 +77,9 @@ function wf_crm_field_options($field, $context, $data) {
     elseif ($table == 'membership' && $name == 'num_terms') {
       $ret = drupal_map_assoc(range(1, 9));
     }
+    elseif ($table == 'membership' && $name == 'auto_renew') {
+      $ret = array(1 => t('Yes'), 0 => t('No'));
+    }
     // Aside from the above special cases, most lists can be fetched from api.getoptions
     else {
       $params = array('field' => $name, 'context' => 'create');
@@ -1226,6 +1229,12 @@ function wf_crm_get_fields($var = 'fields') {
       $fields['membership_end_date'] = array(
         'name' => t('End Date'),
         'type' => 'date',
+      );
+      $fields['membership_auto_renew'] = array(
+        'name' => t('Auto-renew Membership?'),
+        'type' => 'select',
+        'expose_list' => TRUE,
+        'value' => 0,
       );
     }
     // Add campaign fields
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1198,6 +1198,22 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         $params['contribution_recur_id'] = $this->ent['contribution_recur'][1]['id'];
       }
 
+      // if the membership is to be auto-renewed, we set its related recurring contribution ID
+      if (!empty($params['auto_renew'])) {
+        $contributionID = $this->ent['contribution'][1]['id'];
+        $contribution = wf_civicrm_api('Contribution', 'get', array(
+            'sequential' => 1,
+            'id' => $contributionID,
+            'return' => array('contribution_recur_id'),
+          )
+        );
+
+        if (!empty($contribution['values'][0]['contribution_recur_id'])) {
+          $params['contribution_recur_id'] = $contribution['values'][0]['contribution_recur_id'];
+        }
+      }
+      unset($params['auto_renew']);
+
       $result = wf_civicrm_api('membership', 'create', $params);
 
       if (!empty($result['id'])) {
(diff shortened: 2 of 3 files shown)


KarinG:
Got it @omarabuhussein - I'll look at this early next week. Good weekend!

omarabuhussein:
Thanks @KarinG I Am writing a comment now to explain some stuff. Have a nice weekend you too!

omarabuhussein:
So @KarinG there are a few things I discovered while reworking on it and they are different from the previous PR and some things are different from what we discussed in the call but let me walk through them below : 


1- These lines in the previous PR that you asked about : https://github.com/colemanw/webform_civicrm/pull/133/files#diff-59be2e079a691088e15ab3a80e0640a0R1792 are non relevant to my PR and they where already there (not sure why they are appearing in the diff but they are not in this new PR)

2- Regarding this method: https://github.com/colemanw/webform_civicrm/pull/133/files#diff-59be2e079a691088e15ab3a80e0640a0R1938

And while what I explained in the call was correct regarding having this method to ensure that the frequency interval and unit get selected correctly for the the recurring contribution if there are more than one memberships with different frequency unit or interval (and we decided not to handle this case and to show a warning instead which I is something I did in this PR), but this is not the full story, the other purpose for this method is in case the user did not configure the frequency and interval  on the webform, then this method will calculate it based on the memberships frequency unit and duration, in the new version of this PR though I simplified the method so it only pick the first membership it find and base the calculation on it instead of finding the membership with minimum frequency and duration.


3- The previous PR was only work for "Offline"  (Any payment processor that uses Payment_Manual class) or "Pay Later" cases  because it was part of the ticket I worked on specs to make it only work these cases. But because we want this update to be generic as much as possible I also updated the PR to make sure it work even for online payments.

omarabuhussein:
Now To explain some lines in the code to make it easier  for you to review : 

1- The changes in includes/utils.inc just add the "auto-renew" select box to the webform membership tab.

2- The changes added here : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L1190-L1203

are to attach the created recurring contribution to any membership marked to be auto_renewed, remember that memberships in civicrm does not have auto_renew field and instead they have a field called contribution_recur_id instead.


3- The changes here : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L1755-L1758

are to make sure that auto-renew also work for live payments.

4- The changes here : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L1886-L1889

are to make sure that auto-renew also work for offline (pay-later) payments.


5- The line here : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L1813

just ensure that the recurring contribution to be created will be marked as "auto_renew"


6- This method here : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L2600

checks if either we have more than one installment or if at least one membership is configured to be auto-renewed and calculate some of the recurring contribution parameters if needed (like if there is a membership configured to be auto-renewed but there is no frequency unit is configured on the webform then it will get the value from the membership type duration and unit).

7- This method is called by the one above : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L2625

to check if there is at least one membership is configured to be auto-renewed. If no membership is configured to be auto-renewed then it will return false.

8- This method : https://github.com/colemanw/webform_civicrm/blob/7e76fd5b3f752954fdc870a654059e711189ab72/includes/wf_crm_webform_postprocess.inc#L2644

is the simplified version from this one in the previous PR : https://github.com/colemanw/webform_civicrm/pull/133/files#diff-59be2e079a691088e15ab3a80e0640a0R1938

I explained the reason for this in the previous comment (point 2)


9- The added code in js/webform_civicrm_admin.js file is checks if there are more than one membership configured to be auto-renewed, if so then a warning appear telling the user : 

```
Ensure that the memberships selected to be Auto-Renewed have the same frequency unit and interval or otherwise it might not work well !
```

2018-11-10 02_55_00-mail

colemanw:
@KarinG are you still up for reviewing this?
@omarabuhussein there is a conflict that needs to be resolved. Are you able to rebase this?

colemanw:
@omarabuhussein I've changed the base for this PR from 4.x to 5.x. Are you available to rebase it & fix the merge conflict?

omarabuhussein:
just did @colemanw

jamienovick:
@KarinG @colemanw please do let us know if there is anything we could do to get this reviewed - it provides functionality that is essential to get Membership Extras working (as this provides offline autorenew support) and we have to maintain our own fork in the meantime. If we could get this in that would be huge for us! ps. We have 3 clients using this in Prod with no issues... and another 2 going live soon if that helps!

KarinG:
@jamienovick - this PR is not just Adding autorenewal support but is also rewriting everything and thus touching on re: recurring contributions (including donations, monthly donations). As we’ve discussed in the past we really need PRs to just do one thing so that we can test it.

omarabuhussein:
Not sure if it can be broken down but let me see and come back to you

omarabuhussein:
will take a look and see what I can do

omarabuhussein:
I created this PR : https://github.com/colemanw/webform_civicrm/pull/274

to replace part of what this PR does.

omarabuhussein:
I created this PR : https://github.com/colemanw/webform_civicrm/pull/274

to replace part of what this PR does.

PR #182, opened 2018-11-12
Adding support for field of Type Formula on a page before Payment Page

KarinG:
**Overview**
See issue: https://www.drupal.org/project/webform_civicrm/issues/3012908

I believe this is what it would look like if we were to try solve this issue in Webform CiviCRM module. Persisting line_items between pages required a fair bit of work. If there is an easier way to do this within Webform Calculator module then that could be alternate solution. 

**Webform Calculator module:** webform_calculator-7.x-2.x-dev
**Form Configuration:** Contribution Amount is a field of Type Formula

image

**Before**
Key issue is that the result of the Formula field calculation appears on the Payment Page - but the amount transacted excludes the calculated $amount portion.

image

image

**After**
The amount transacted now includes the calculated $amount portion. Here's a more complex example to also help ensure Memberships, Participant Fees, LineItems continue to get included properly:

image

**Technical Details**
Key is this line to persist data between webform pages:
`    $this->line_items = wf_crm_aval($form_state, 'civicrm:line_items', array());`

This line is already in use in wf_crm_webform_preprocess.inc: https://github.com/colemanw/webform_civicrm/blob/7.x-4.x/includes/wf_crm_webform_preprocess.inc#L28

Using it in wf_crm_webform_postprocess.inc - allows for a field of Type Formula on a page before Payment Page;

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -407,7 +407,6 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
             // Only if the element does not already exist
             $elementfound = $this->searchForLineItem('civicrm_'.$c.'_participant_'.$n.'_participant_'.$id_and_type, $this->line_items);
             if (is_null($elementfound)) {
-              $test = 1;
               $this->line_items[] = array(
                 'qty' => $count,
                 'entity_table' => 'civicrm_participant',
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -75,6 +75,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $this->crmValues = wf_crm_enabled_fields($this->node, $this->rawValues);
     // Even though this object is destroyed between page submissions, this trick allows us to persist some data - see below
     $this->ent = wf_crm_aval($form_state, 'civicrm:ent', array());
+    $this->line_items = wf_crm_aval($form_state, 'civicrm:line_items', array());
 
     $this->hiddenFieldValidation();
     $this->validateThisPage($this->form['submitted']);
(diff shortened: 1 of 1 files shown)


jitendrapurohit:
I was able to replicate this scenario and confirmed that this PR fixes the main issue.

I looked into the changes and added some debug statements into the code(without this PR), my testing result shows that the calculated field is not getting retrieved in the `$form_state['storage']['submitted']` variable avaiable [at this point](https://github.com/colemanw/webform_civicrm/blob/7.x-4.x/includes/wf_crm_webform_postprocess.inc#L71). I wasn't able to invest more on this so not completely sure about the reason. It could be related to webform calculator handling the form field submission(not sure).

I see if it retrieves the value correctly - this issue is fixed and all this processing of the line item manipulation might not be required?

KarinG:
Hi @jitendrapurohit - thank you for having a look! 

I'm not convinced yet where this needs to be fixed either. 

Here was my thought process:

1. Noticed issue -> went straight into webform_calculator module - for a first look - and like you noticed it does not appear to be saving to `$form_state`

2. Then I remembered that the line_item tally on the webform_civicrm payment page is correct. How is that possible? Then I noticed this line in wf_crm_webform_preprocess.inc: https://github.com/colemanw/webform_civicrm/blob/7.x-4.x/includes/wf_crm_webform_preprocess.inc#L28

3. I'm not sure why that was done - but this made me go back to webform_civicrm and decide to extend this preprocess logic to wf_crm_webform_postprocess - essentially using: in L104 -> the line_items between page submissions were already being stored for safe keeping: https://github.com/colemanw/webform_civicrm/blob/7.x-4.x/includes/wf_crm_webform_postprocess.inc#L104

4. but I'm still not sure it's the right path - should we go back to try and fix webform_calculator? What really bugs me is that our payment page total is correct [via this store line_items between page submissions trick] - but our checkout total_amount is not. That has me worried about other things that could go wrong - it is a bit scary to be doing something different in `preprocess` compared to `postprocess`.

KarinG:
@colemanw - if you have a moment could you please let us know what you think is the best way to handle this?

KarinG:
@jitendrapurohit - I've now pulled this into two live projects. One is setting up a complex ticket calculation fee; the other one is calculating payment processing fees and handling and adding that to the total_amount prior to transaction. This is actually with a hidden Processing Fee formula field and working well. 

But still not convinced it's the way to go permanently. What do you think we should do? 

image

jaapjansma:
I have tested this PR. I do think it is a really nice addition. I got it working in the end but got stuck during testing.

I discovered that the fields where the calculation is based on should all be on the same page. Reading the comments above I do think this is a bug in the webform_calculator module. And should be fixed in that module. But I am not sure on what time/moment the maintainer of the webform_calculator module is going to merge the fix. 

So my advice would be to create a workaround in this module and at the same time fix the webform_calculator module. We could then later on remove the work around from this module.

KarinG:
I'm going to close this in favour of: https://www.drupal.org/project/webform_calculator/issues/2393499#comment-13098116

Direct link to our patch there: https://www.drupal.org/files/issues/2019-05-07/webform_calculator_semperit.diff

This is D7WFC - it allows formula fields to be calculated over multiple pages. Tested it with a number of webform civicrm like scenarios. @jitendrapurohit and/or @jaapjansma - if you can give it a try and comment on the drupal.org issue - that would be helpful.

zkrebs:
@KarinG I was able to test your May 7th submission to the webform_calculator module and it works like a charm. Once again - thank you!

petednz:
@zkrebs per comment above yours. pls add your 'reviewed' info to https://www.drupal.org/project/webform_calculator/issues/2393499#comment-13098116

PR #183, opened 2018-11-22
CATL-1020: Load Default Value For A Contact From Case Relationships

vinuvarshith:
Overview
This feature is already existing for non case relationships. We want to load the default value of the existing contact field on a webform through a case relationship
scrn

Before
setup of the webform to test what's existing:
When we go to the webform - just as an example - we can add 2 contacts
Contact 1 can be the current user
We can set contact 2's default value to load from a relationship. This can be done by editing the existing contact field of Contact 2 on the webform, and setting the 'default value' to load from a relationship.
See screenshot.
This, however, only sets about non case relationships. There's currently no way to load a contact from case relationships (case roles)

After
In the first dropdown where we specify the source of loading (Set default contact from) - we currently have 4 options: None, Specified Contact, Relationship To, Auto - from filters

Now there is another option to this dropdown: 'Case Relationship to'
The second field is to select a contact (remains the same)
The third field where we need to specify the relationship - on selecting 'Case Relationship to' - this will now load only case relationships (case roles)
This option will only be available when cases in enabled on the webform.

In addition if only the case id is given in url and if 'contact 1: existing contact' item in webform config is set to load 'defaults' from 'case roles' and 'case client' is set for contact 1, then this will automatically load the first case client of the case passed in the url. (Previously passing only the case would not autofill anything)
Also 'Current User' option is now available for all contacts (Previously it was only available for contact1)

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -41,6 +41,8 @@ function _webform_defaults_civicrm_contact() {
       'default_contact_id' => '',
       'default_relationship' => '',
       'default_relationship_to' => '1',
+      'default_for_case' => '1',
+      'default_case_roles' => '',
       'allow_url_autofill' => TRUE,
       'dupes_allowed' => FALSE,
       'filters' => array(
@@ -216,10 +218,10 @@ function _webform_edit_civicrm_contact($component) {
     '#default_value' => $component['extra']['default'],
     '#parents' => array('extra', 'default'),
   );
-  if ($c == 1 && $contact_type == 'individual') {
+  if ($contact_type == 'individual') {
     $form['defaults']['default']['#options']['user'] = t('Current User');
   }
-  elseif ($c > 1) {
+  if ($c > 1) {
     $form['defaults']['default']['#options']['relationship'] = t('Relationship to...');
     $to = $component['extra']['default_relationship_to'];
     $form['defaults']['default_relationship_to'] = array(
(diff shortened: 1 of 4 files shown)


shitijg:
@jitendrapurohit guess you asked for this?

shitijg:
@colemanw if you can review this one - that will be great.

jitendrapurohit:
I've applied this PR on my review setup. I'll post the test result soon.

includes/contact_component.inc
```diff
@@ -238,6 +240,31 @@ function _webform_edit_civicrm_contact($component) {
       '#parents' => array('extra', 'default_relationship'),
     );
   }
+  
+  $case_count = $data['case']['number_of_case'];
```
jitendrapurohit: I get a notice error on this line while editing a contact component - 

    Notice: Undefined index: case in _webform_edit_civicrm_contact() (line 244 of /Applications/MAMP/htdocs/civicrm-master/sites/all/modules/webform_civicrm/includes/contact_component.inc).

jitendrapurohit (commented):
Looks pretty much great. This does fix the requirement of #171 and much more flexibility for case contacts. I've found a single notice error as of now. Can you pls push an update to fix that?

Thanks for the PR.

vinuvarshith:
@jitendrapurohit The above notice is fixed. Thanks.

shitijg:
@colemanw can we merge this?

PR #184, opened 2018-11-23
Allow negative line items like e.g. discounts.

KarinG:
Overview
Ad-hoc to https://github.com/colemanw/webform_civicrm/pull/159
Allow negative line items like e.g. discounts.

**Example:**

image


colemanw:
Hmm, this may introduce a vulnerability in weborms where users can give themselves discounts by entering a negative number on donation amounts.

KarinG:
That's easy to prevent: the default Minimum for civicrm_1_lineitem_1_contribution_line_total = 0; so in order to allow negative amounts admin needs to hit edit on that form element and remove that Minimum (and could for safety put their max discount in).

colemanw:
Good point. AFAIK that's been the default since forever.

PR #185, opened 2018-11-26
[WIP] Payments

mglaman:
Overview
Adds the ability to use a contribution form to accept payments for a one-off donation or others.

Technical Details

Done / highlights

* Fixes setting `select` value options when inserting elements
* Ensures that the verbatim functionality for simple payments work in D8 like D7.

Comments
- [ ] Page break does not nest Payment elements, so it's one page. This needs to be looked at. Do we fix now in this PR or later?
- [ ] Create elements to mimic the component alters for Build like in D7. This is actually a long-running TODO for the entire project.
- [ ] Try to break it in weird ways


includes/utils.inc
```diff
@@ -926,22 +926,22 @@ function wf_crm_get_fields($var = 'fields') {
         'name' => ts('Contribution Page'),
         'type' => 'hidden',
         'expose_list' => TRUE,
-        'empty_option' => t('None'),
+        'empty_option' => (string) t('None'),
```
dsnopek: All of these instances of casting to a string right away aren't ideal:

> "This object, when cast to a string, will return the formatted, translated string. Avoid casting it to a string yourself, because it is preferable to let the rendering system do the cast as late as possible in the rendering process, so that this object itself can be put, untranslated, into render caches and thus the cache can be shared between different language contexts."

See https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21StringTranslation%21TranslatableMarkup.php/class/TranslatableMarkup/8.6.x 

Could these be handled better later on, rather than cast to a string?

Or, if that's a whole project in itself, maybe we could make an issue to handle this later and then put `// @todo` with references to this issue everywhere that we're casting translatable markup to a string?

includes/wf_crm_admin_form.inc
```diff
@@ -1214,11 +1215,16 @@ class wf_crm_admin_form {
    * Ajax-loaded mini-forms for the contributions tab.
    */
   private function checkSubmissionLimit() {
-    $webform = &$this->node->webform;
+    $webform = $this->webform;
+
+    $anonymoous_access = $webform->checkAccessRules('create', new \Drupal\Core\Session\AnonymousUserSession());
```
dsnopek: Spelling issue: 'anonymoous' should be 'anonymous' (one less 'o')

includes/wf_crm_webform_preprocess.inc
```diff
@@ -214,6 +211,9 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
         // RM: This is needed in order for CiviCRM to know that this is a 'test' (i.e. 'preview' action in CiviCRM) transaction - otherwise, CiviCRM defaults to 'live' and returns the payment form with public key for the live payment processor!
         $contributionCallbackQuery['action'] = CRM_Core_Action::description(CRM_Core_Action::PREVIEW);
       }
+      if ((strpos($contributionCallbackUrl, '/') !== 0)) {
```
dsnopek: A nit, but this is a weird way to check this. Why not just:

```
if ($contributionCallbackUrl[0] != '/') {
```

includes/wf_crm_webform_preprocess.inc
```diff
@@ -615,13 +616,16 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
...
+
+    $build = [
+      '#type' => 'table',
+      '#sticky' => FALSE,
+      '#caption' => $this->contribution_page['title'],
+      '#header' => array(),
+      '#rows' => $rows,
+      '#attributes' => array('id' => "wf-crm-billing-items"),
+    ];
+    return \Drupal::getContainer()->get('renderer')->render($build);
```
dsnopek: Ugh. Similar to the thing about converting translation markup to strings early, is it possible to avoid the early render here and just return the render array? And, again, if doing so would be a big project, it should have an issue to address this later.

dsnopek (commented):
I haven't tested but I read through the code changes. I know this is still early WIP, so some of the things pointed out in my review are probably already on your list, but hopefully this will be at least a little useful. :-)

includes/utils.inc
mglaman: We can't, due to the way Webform dumps to YAML. Any objects explode it. There's somewhere else in the code that walks an object to convert all translatable markup to strings.

Any string in these function, which is translatable, gets dumped into Webform YAML anyways. 

If we remove that, the form dies due to Yaml encoder not being setup to handle objects (I wish it just did the casting for us.)

includes/wf_crm_webform_preprocess.inc
mglaman: That's a better idea (I copied core.) But i fixed the actual callback URL to have a `/`.

includes/wf_crm_webform_preprocess.inc
mglaman: It's a big project because it adds this to the `#prefix` or something. Not ideal, but I don't want to tackle it in this PR.

mglaman:
@dsnopek good news! I was able to have a successful transaction. Can you give this a whirl locally? There's _bugs_ but a basic contribution form, enable the Contribution settings, submit.

- [ ] Page break does not nest Payment elements, so it's one page. This needs to be looked at
- [ ] Create elements to mimic the component alters for Build like in D7
- [ ] Try to break it in weird ways

KarinG:
Awesome! I’ll be taking it for a spin too :-)

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1973,22 +1973,25 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    * @return string
    */
   public function getIpnRedirectUrl($type) {
-    $url = trim($this->node->webform['redirect_url']);
-    if ($url == '<none>' || $type == 'cancel') {
-      $url = url('node/' . $this->node->nid, array('absolute' => TRUE));
-    }
-    elseif ($url == '<confirmation>') {
-      $query = array('sid' => $this->submission->sid);
+    $confirmation_type = $this->node->getSetting('confirmation_type');
```
dsnopek, line 1976: Where is this node coming from? I've been testing Webforms not attached to nodes, which even if attaching Webforms to nodes, that's really just like an entity reference. Is this meant to be looking at setting on the Webform object?

mglaman: `node` == `webform`. I just never changed the property name before. Because in D7 the webform data was saved in the node.

dsnopek: Ooookay. So, the property name is 'node' to avoid having to change so many things, but it's really a webform object. Got it.

We should make an issue to rename that later, perhaps before doing a proper release? It's gonna confuse people in the future.

dsnopek:
I tried to test this locally, but I'm having some issues even configuring it. On 'master', I'm able to select a contribution form from the dropdown on the admin page and some additional options for configuring it appear. But on this branch, I get a `Drupal.AjaxError` exception thrown and the additional fields don't load. The AJAX error is kinda weird too in that it says it got 200 response:

```
An AJAX HTTP error occurred.
HTTP Result Code: 200
Debugging information follows.
Path: /admin/structure/webform/manage/civicrm_contribution/civicrm?ajax_form=1
StatusText: OK
ResponseText: 
```

I think the problem is that there's no response text?

includes/wf_crm_webform_postprocess.inc
KarinG: Issue filed/documented: https://www.drupal.org/project/webform_civicrm/issues/3017497

dsnopek: Thanks, Karin!

includes/utils.inc
dsnopek: Er, ok, so if the untranslated strings are meant to be in Yaml, and something in Webform turns them into translateable strings, then we'll want to NOT use translated strings here. Otherwise, the content will end up being stuck in some language and no longer translateable.

For example, let's say the current user is using the UI is Polish, so `(string) t('None')` becomes the string `'Nic'`. Then `'Nic'` gets saved in the Yaml, and whatever process walks that and makes translateable strings will do the equivalent of `t('Nic')` which won't find anything, and the text is now stuck now in Polish.

includes/wf_crm_webform_preprocess.inc
dsnopek: Ok! This could, at least temporarily, lead to some interesting interactions with the render cache, so I guess we'll need to keep an eye out for that. Here's an issue I just made for it:

https://www.drupal.org/project/webform_civicrm/issues/3018013

mglaman:
@dsnopek with 1911761b66c51aaba6fedf62995f94f743f0c451 the error should be resolved.

includes/utils.inc
mglaman: I removed the workarounds in this PR. Later we should go through `wf_crm_get_fields` and remove all `t()` calls.

includes/wf_crm_webform_preprocess.inc
mglaman: :) This was fixed.

dsnopek:
Sorry for taking so long to get back to this!

I did some manual testing on the latest, and I was finally able to successfully enable contribution on a Webform.

I tested with PayPal first, and everything seemed to work, although, I was testing locally against the PayPal sandbox, so PayPal's IPN couldn't ping my site back and the contribution was marked as "Pending (Incomplete Transaction)" in CiviCRM. Then I tested with iATS, which I was kind of expecting not to work because Matt told me he didn't test it, but it worked all the way through!

@KarinG if you wanted to do some manual testing on this, I think it's ready for that :-)

@mglaman In [this comment](https://github.com/colemanw/webform_civicrm/pull/185#issuecomment-442702159) above, you gave 3 checkboxes of things that still need to be done. Do any of those still need to be done? The first (the page break thing) seems like it may still not be working right? I'm not sure about the alter thing.

What else do we need to do to get this merged?

KarinG:
Yep lots is working! 

1. Transacting Credit Card - this is a live iATS transaction with a fake VISA - and this is perfect:
image

2. and had to check :-) ah looks like me demo/test site is in USD:
image

3. Ah here's an issue: when I have multiple payment processors configured as: user select -> I can toggle on the webform from Pay Later to iATS ACH/EFT to iATS Credit Card - but when selecting Credit Card -> it still expects the 'required' ACHEFT fields.
image

Glad I found an issue :-)

4. Ah - second issue: when trying to edit the payment processor option/default - the invalid #weight property surfaces again (like we saw earlier on in the Contacts section):
image

5. Ok - recurring contributions! Why not... ah - installments is not working yet. When configuring $400 to be paid monthly in 4 installments by credit card [here's the screenshots for it]:

image

image

It should transact $100 and set up a recurring series for another 3 installments $100/each - but it simply transacts the $400 all at once. I bet you that's just not reading the installments # properly (b/c installments = 1 = default => means transact all at once]:
image

6. Installments = 0 => means transact open-ended (no installments) - that's a CiviCRM thing - all we need to do is ensure that we create a recurring series with 0 installments. 

I'm amazed at how well this is already working though! Awesome!

mglaman:
> @mglaman In this comment above, you gave 3 checkboxes of things that still need to be done. Do any of those still need to be done? The first (the page break thing) seems like it may still not be working right? I'm not sure about the alter thing.

I added those to the main body so we can discuss. I'll address them here

> Page break does not nest Payment elements, so it's one page. This needs to be looked at

I tried working on this and had issues. I'd like to merge this PR as is, if we could, and then figure it out. Page breaks do not work like they did in D7 and nest their elements. So.. things get kind of hard.

> Create elements to mimic the component alters for Build like in D7

This might help simplify the page break (one element vs the three various ones.) I tried working on it, but I ended up breaking things. I think it would be best to merge this as is, and then try to improve the elements component.

> Try to break it in weird ways

:) This was for you both, so we can decide if its ready to merge.

> What else do we need to do to get this merged?

I need to look at @KarinG's reported issues

* Ah here's an issue: when I have multiple payment processors configured as: user select -> I can toggle on the webform from Pay Later to iATS ACH/EFT to iATS Credit Card - but when selecting Credit Card -> it still expects the 'required' ACHEFT fields.
* Ah - second issue: when trying to edit the payment processor option/default - the invalid #weight property surfaces again (like we saw earlier on in the Contacts section):
* Ok - recurring contributions! Why not... ah - installments is not working yet. When configuring $400 to be paid monthly in 4 installments by credit card [here's the screenshots for it]:

Do we want to tackle recurring in this initial functionality support?

KarinG:
Hey Matt - don't worry about recurring for this initial commit on Contribution functionality. That was me trying to find something that wasn't working :-) I'm very impressed with how much of this is already working. 

I do think that "`invalid #weight property`" and the "`requiring fields that should not be required`" are two good candidates to try and address at this stage as we'll get people reporting this to us fairly soon after merging this. If you can get to those than we can create separate issues for "`Page break`" and "`Recurring`"

mglaman:
@KarinG sounds good! The invalid properties should be fixed by 9f65900.

I'll work on the required fields goofiness, next.

KarinG:
Great! Very close on merging this... :-)

mglaman:
@KarinG I have set up iATS and submitted to the sandbox both ACH and CC and have not been able to reproduce the required fields problem :/

KarinG:
Ah - ok - let me try re-reproduce this:

1. Payment Processor field is set to - User Select -
image

2. This correctly produces radio buttons on the form for every payment processor that is enabled on the Contribution page configured for this form (KG Test in my case):
image

image

3. Selecting the radio button for iATS ACHEFT - looks nice - 
image

4. But deciding I want to pay by Credit Card -> I'm hitting the radio button iATS Credit Card -> checking out:
image

5. Submitting the form -> produces the validation error
image

6. So the issue only appears when there are "radios" (user-select) on the form. When looking at these radio button option element: option values are 0, 3, and 1 and Value is 3 (which happens to be the ACHEFT payment processor) - I did not edit / configure this element - this is how it ended up on the form.  
image

mglaman:
@KarinG I still cannot reproduce.

I have the same setup as your screenshots. The payment goes through or triggers a fail scenario.

PR #186, opened 2018-11-26
Fix JavaScript errors, update library definitions.

mglaman:
Overview
Fixes some JavaScript errors discovered while working on #185. Fixes invalid reference to D7 Drupal.settings, now drupalSettings

Before
Threw a console warning.

After
No longer throws a console warning.

Comments
Did not try to replicate on 8.x-5.x, but it triggers if enabling a Contribution setting.

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -194,10 +194,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
     // Preprocess contribution page
     if (!empty($this->data['contribution'])) {
       $this->addPaymentJs();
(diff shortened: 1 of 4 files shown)


dsnopek:
I tried to either:

1. Reproduce this issue with 8.x-5.x, or
2. Test the changes added in this PR

... but I wasn't able to either. Without the changes in PR #185, there's no way to save the admin form with a contribution enabled (I get `Error: Call to undefined function webform_component_invoke()`) so there's no place it'll load the new library.

Anyway, all that said, the changes in this PR all look correct. And since I can't get the code to activate, I don't think they break anything new.

So, @KarinG I think this should be safe to merge!

PR #189, opened 2019-01-11
Block interaction with payment form until payment block has fully loaded

mattwire:
Overview
Loading the civicrm payment "billing block" is done via AJAX.  It can take a few seconds to load (sometimes long enough that someone can press the submit button before it has loaded).

Before
Possible to submit the form before payment block has loaded - there is no indication to the user that the form has not finished loading.

After
Not possible to submit form before payment block has loaded - form is "dimmed" until it loads making it very obvious that it has not finished loading.

Technical Details
This uses the jquery blockUI that is already used elsewhere in CiviCRM.  It would be nice to have the loading "spinner" too but I can't work out how to do that!

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -5,6 +5,13 @@ cj(function($) {
     setting = Drupal.settings.webform_civicrm,
     $processorFields = $('.civicrm-enabled[name$="civicrm_1_contribution_1_contribution_payment_processor_id]"]');
 
+  $(document).ajaxStart(function() {
+    $('#billing-payment-block').closest('form').block();
+  })
+  .ajaxStop(function() {
+    $('#billing-payment-block').closest('form').unblock();
+  });
+
   function getPaymentProcessor() {
     if (!$processorFields.length) {
       return setting.paymentProcessor;


js/webform_civicrm_payment.js
```diff
@@ -5,6 +5,13 @@ cj(function($) {
     setting = Drupal.settings.webform_civicrm,
     $processorFields = $('.civicrm-enabled[name$="civicrm_1_contribution_1_contribution_payment_processor_id]"]');
 
+  $(document).ajaxStart(function() {
+    $('#billing-payment-block').closest('form').block();
+  })
+  .ajaxStop(function() {
+    $('#billing-payment-block').closest('form').unblock();
+  });
+
```
colemanw, line 14: I don't think we should block the form for every use of ajax, just the one we care about. So I'd put this into the `loadBillingBlock` function.

colemanw (changes requested):

js/webform_civicrm_payment.js
mattwire: @colemanw I've got something similar for core, and you've hit the issue on the spot.  The problem with loadBillingBlock is it loads way too late on the form to be of any real use and you can still submit the form for a while before it freezes.

The ajaxStart/Stop functions can only be linked to `document`.  And if we want to limit it to only when the billing ajax request is going to be triggered we still want to do it as soon as the form loads at which point we don't know if it's going to have billing info on it.  By targetting #billing-payment-block we are making sure it doesn't target other forms but it will trigger for every ajax request on that form.

colemanw: Well, since this is using `cj` it will only attach to Civi's ajax requests not any initiated by Drupal, so... ok.

PR #190, opened 2019-01-11
Link recurring contribution to membership record if we have one

mattwire:
Overview
If we submit a recurring payment via the webform link it to any memberships being created at the same time.

Before
Recurring contribution not linked to membership - membership will not auto-renew.

After
Recurring contribution is linked to membership - membership will show links to recurring contribution and will auto-renew if the payment processor supports it.

Technical Details
Just add the contribution_recur_id when creating the membership.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1187,6 +1187,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       // The api won't let us manually set status without this weird param
       $params['skipStatusCal'] = !empty($params['status_id']);
 
+      if (isset($this->ent['contribution_recur'][1]['id'])) {
+        $params['contribution_recur_id'] = $this->ent['contribution_recur'][1]['id'];
+      }
+
       $result = wf_civicrm_api('membership', 'create', $params);
 
       if (!empty($result['id'])) {
@@ -1804,6 +1808,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     }
 
     $resultRecur = wf_civicrm_api('ContributionRecur', 'create', $contributionRecurParams);
+    $this->ent['contribution_recur'][1]['id'] = $resultRecur['id'];
 
     // Run the Transaction - and Create the Contribution Record - relay Recurring Series information in addition to the already existing Params [and re-key where needed]; at times two keys are required
     $contributionParams['total_amount'] = $contributionFirstAmount;


colemanw:
Thanks @mattwire - one nitpick, I think for consistency it should be `'contribution_recur'` not `'contributionrecur'` to follow the entity naming convention.

mattwire:
> Thanks @mattwire - one nitpick, I think for consistency it should be 'contribution_recur' not 'contributionrecur' to follow the entity naming convention.

@colemanw Agreed - I've updated it.

KarinG:
After rebasing a couple of clients to 5.x ahead of their new membership season, I discovered that this causes a conflict with the instalment feature that webform provides. 

**Scenario**
1. Membership Type "Professional - Instalments" is configured as "No auto-renew option". It's a 1 year Membership. 
2. Webform allows any type of monies (Contribution, Event, Membership) to be split up into instalments - so that they can be paid e.g. Monthly. Webform recreates a Recurring Series. In this case it's a $129 + GST Membership that can be paid in 12 monthly instalments.
3. Connecting the Recurring series to the Membership (this PR) means that every instalment -> the Professional get +1year added to their Membership. Here are two test cases: 

image

Laura's is created with our previous older 4.x code base - note Auto-renew is OFF. Karin's is create with the new 5.x code base - note Auto-renew got switched ON (even though the Membership Type config in Core says NO) and note that I've received a +1 year every time I trigger recurring contribution to simulate an instalment payment.

A longer term solution will be to untangle the dual purpose of recurring contributions [see many other threads] - but for now: if paying for a Membership in instalments - we don't want these payments to be connected as to Core that appears to translated into -> auto-renewal, but it's not. An auto-renewal would be extending the current 12 monthly instalments to 24 monthly instalments [and +1y of Membership end date].

mattwire:
@KarinG What does the database say? Looking at that screenshot both Laura and Karin have auto-renew switched ON, but there is a problems with the recurring contribution for Laura (it's either failed or cancelled).

Part of the problem here is that webform is trying to do something that is not supported by civicrm core. There has been discussion on and off about being able to control what actually happens to other entities (eg. membership) when a recurring contribution receives another payment and that would seem like the real solution here. Is webform_civicrm still creating MembershipPayment records to link the membership to it's payment?

KarinG:
Ah yes - wrong screenshot to illustrate my point. Laura's recurring contribution got removed just before I took that screenshot. After reverting this PR for their instance everything is normal -> instalments are being processed and we're not ending up with +1y end-date every month. 

Not saying that reverting is the answer! 

CiviCRM natively does have instalments - and it works great for Contributions not tied to Membership. Instalments using webform for Events work fine - but Membership is a 3D problem. Essentially is_recur has become a synonym for autorenewal in Core [right down to the message template]. Whereas I'm interested in:

1 term - pay in 12 instalments
autorenewal adds +1y and 
extends existing recurring series by 12 instalments
or
creates a new recurring series of 12 instalments
so that's **_recurring the recurring series_**

I created instalments originally for a swim club -> a membership $3000 in fees -> giving parents an option to pay in 10 instalments of $300/each. I suppose in Core you would need to create a Membership $300 duration: 1month. But it's not a 1month membership - it's a 10mo commitment. I'm really not sure how we could tackle this in 3D. There is a good use case out there for intstalments. Another client with us sells professional association Memberships - they are not cheap and monthly instalments is of interest to them. I can see why. 

Yes webform_civicrm is hooking up payments for Membership and Events. That too can look funny as you can pay for e.g. 2 events and 1 membership one one webform. Core just has the one payment so it thinks you're owed a refund - everywhere.

PR #191, opened 2019-01-18
Fix state/province from being cleared from addresses when province isn't a submitted field

BrianTheWiz:
## Overview
State/Province are being cleared from a CiviCRM contact's address when a webform is submitted that does not have a State/Province field for that address. Resulting in corrupted data anytime someone submits a form without a State/Province field.

## Before
Create a webform that has some address fields that are not State/Province. Have a CiviCRM contact that has multiple addresses with a State/Province. Submit the form for that contact. All addresses will be cleared.

### Form (home address)
screen shot 2019-01-18 at 4 04 52 pm

### Before Submitting
screen shot 2019-01-18 at 4 06 00 pm

### After Submitting (notice even billing is affected)
screen shot 2019-01-18 at 4 06 47 pm

## After
Now when you submit a webform that has no State/Province, it does not clear the State/Province. Or if you do have a State/Province field, it only modifies it for that specific address.

## Technical Details
We are on CiviCRM 5.8.2. It seems like there was some code in there that was converting the `state_province_id` from an abbr to an id, but it was already an id.

## Comments
First time diving into the code for this module so forgive me if I'm missing a key element here. We just upgraded CiviCRM from 4.6 to 5.8 last weekend and been fighting bugs.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -766,13 +766,6 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $existing = array_merge(array(array()), $result['values']);
         }
         foreach ($contact[$location] as $i => $params) {
-          // Translate state/prov abbr to id
-          if (!empty($params['state_province_id'])) {
-            $config = CRM_Core_Config::singleton();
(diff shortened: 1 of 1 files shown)


colemanw:
Ok I think I see what's going on. It is indeed a problematic regression caused by the location reordering.
Here is a more complete fix: https://github.com/colemanw/webform_civicrm/pull/192

PR #192, opened 2019-01-18
Fix address field regression

colemanw:
Replacement for https://github.com/colemanw/webform_civicrm/pull/191


BrianTheWiz:
Tested this just now. It seems to be working. Tested with and without State/Province. Thank you!

PR #193, opened 2019-01-19
Pay Later option causes error in createDeferredPayment

mglaman:
Overview
When using the "Pay Later" option, \wf_crm_webform_postprocess::createDeferredPayment returned an error. That is because fetching the `0` payment_processor_id for Pay Later errors on

```
      $liveProcessorName = wf_civicrm_api('payment_processor', 'getvalue', array(
        'id' => $params['payment_processor_id'],
        'return' => 'name',
      ));
```

This bypasses the name lookup. The option for Pay Later shouldn't probably be allowed unless Pay Later is enabled on the contribution page. But this does some cleanup and fixes the warning generated.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -5,6 +5,7 @@
  * Front-end form validation and post-processing.
  */
 
+use Drupal\Component\Utility\NestedArray;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element;
 use Drupal\user\Entity\User;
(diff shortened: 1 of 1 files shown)


dsnopek:
I can't for the life of me reproduce this error. :-/ Pay Later seems to be working fine (both user select, and when it's mandated) when I'm testing on the 8.x-5.x branch. But it isn't making things any worse, so...? :-)

@KarinG I know you had this problem - can you test out this PR?

KarinG:
I can't reproduce this error either. I can get Pay Later to work (by itself) - and all looks good after submission:

image

KarinG:
PS -   private function createDeferredPayment() went in not too long ago - trying to remember the details as to why. But how odd that you're running into issues/bugs and vice versa!

mglaman:
@KarinG basically it's a WARNING. So it's not always visible. The problem is when trying to fetch the name of the payment processor with ID 0. Or my CiviCRM install doesn't have a payment processor of `0` to fetch the name. So it passes an error array and not proper int.

KarinG:
Ah ok - I did not get a warning/notice in the dblog either - wicked... Shall I merge it @dsnopek ?

dsnopek:
> Shall I merge it @dsnopek ?

Well, the changes also do some good modernizing to this bit of code (switching `$_POST` to looking at the request object, etc) that will need to be done eventually anyway. So, if @mglaman can reproduce the original issue, and the changes are positive anyway... yeah, why not!

PR #194, opened 2019-01-19
Custom Groups + Fields and Fieldsets

mglaman:
Overview
Fixes the creation of fieldsets and errors when saving custom fields.

Fieldsets required nesting values, unlike D7 where it just worked by the order of which elements are displayed. That makes this all a bit more complicated.

Before
```
The website encountered an unexpected error. Please try again later." Error: Call to undefined function webform_component_invoke() in wf_crm_admin_form::addFieldset() (line 2068 of /Applications/MAMP/htdocs/drupal853.local/modules/webform_civicrm/includes/wf_crm_admin_form.inc)
```

After
screen shot 2019-01-18 at 10 34 01 pm

Technical Details
Ensures that fieldsets are created when elements are updated. 

Comments
Needs testing to shake loose bugs.


dsnopek:
I did some testing of this...

One problem I hit right away, is that I can't create custom fields when the 'webform_civicrm' module is enabled - it just hangs after hitting "Save". The field gets created, but isn't completely functional.

I think this is due to `webform_civicrm_civicrm_postSave_civicrm_custom_field()` or `webform_civicrm_civicrm_post()` which are CiviCRM hooks that attempt to support the 'Add dynamically' feature of custom fields. I think we need to fix this and probably make 'Add dynamically' work to really support custom fields.

But if I uninstall webform_civicrm, create some custom fields, and then re-install it, I can get custom fields on the form when I use a simple Alphanumeric / Test field. However, it shows the field outside the field site rather than inside it:

selection_359

It also didn't save the content that I put in there to the CiviCRM contact!

I didn't test any other field types or input methods because it was such a pain to add/remove custom fields with the first bug I mentioned, but we should test out the field types and input methods to make sure all kinds of custom fields can be shown.

colemanw:
Those 2 hooks are probably using code that references the d7 webform data structure and need to be updated.

colemanw:
Honestly the "dynamic custom fieldset" feature is pretty obscure, so if removing it is easier than supporting it, just remove it.

mglaman:
We could just "stop" the insert of fieldsets and maybe make a note "Don't use fieldsets or any element grouping" until it can be refactored. It will be a pretty heavy load and change how the entire Webform's elements are inserted.

KarinG:
Thanks for digging into this Matt - I agree - let's stop insertion and park the fieldset issue for now. As you said custom fields are working. Let's move on to other bits like e.g. Events first.

mglaman:
Okay, I will push a commit which stops the fieldset generation and ping here when that is done!

KarinG:
Thanks Matt!

mglaman:
@KarinG @dsnopek okay, my last commit disables fieldset generation. Field info is also moved to a service so it can start to be better maintained in OOPy fashion.

includes/wf_crm_admin_form.inc
```diff
@@ -1712,15 +1719,17 @@ class wf_crm_admin_form {
...
+                list(, $c, $ent) =  Utils::wf_crm_explode_key($key);
                 $type = in_array($ent, self::$fieldset_entities) ? $ent : 'contact';
                 $create = !isset($previous_data['data'][$type][$c]);
                 /*
     list(, $c, $ent) =  wf_crm_explode_key($field_key);
     $type = in_array($ent, self::$fieldset_entities) ? $ent : 'contact';
     return !isset($this->node->webform_civicrm['data'][$type][$c]);
                  */
-                self::insertComponent($field, $enabled, $this->settings, !isset($previous_data['data'][$type][$c]));
+                // Temp disabled to prevent fieldsets.
```
dsnopek: If this is going to get merged like this, we should probably but a more formal `@todo` referring to an issue here?

includes/wf_crm_admin_form.inc
```diff
@@ -2002,10 +2008,11 @@ class wf_crm_admin_form {
       $field['value'] = $settings[$field['form_key']];
     }
     // Create fieldsets for multivalued entities
-    if ($ent != 'contribution' &&
-      ($ent != 'participant' || wf_crm_aval($settings['data'], 'participant_reg_type') == 'separate')
+    if ($ent !== 'contribution' &&
+      ($ent !== 'participant' || wf_crm_aval($settings['data'], 'participant_reg_type') === 'separate')
     ) {
-      self::addFieldset($c, $field, $enabled, $settings, $ent, $create_fieldsets);
+      // Temp disable fieldsets.
```
dsnopek: Same here.

includes/wf_crm_webform_ajax.inc
```diff
@@ -22,7 +24,7 @@ class wf_crm_webform_ajax extends wf_crm_webform_base {
       drupal_json_output(array('' => t('- first choose a country -')));
       exit();
     }
-    drupal_json_output(wf_crm_get_states($input));
+    drupal_json_output(Utils::wf_crm_get_states($input));
```
dsnopek, line 27: Perhaps this was updated in a search-and-replace? If so, I guess that's fine. However, `drupal_json_output()` isn't a thing in Drupal 8, so if this code actually executes, it'd be a fatal error.

webform_civicrm.module
```diff
@@ -27,16 +28,6 @@ define('WEBFORM_CIVICRM_DEFAULT_CONTACT_ID', 1);
...
-  $items['node/%webform_menu/civicrm'] = array(
-    'title' => 'CiviCRM',
-    'page callback' => 'drupal_get_form',
-    'page arguments' => array('wf_crm_configure_form', 1),
-    'access callback' => 'wf_crm_admin_access',
-    'access arguments' => array(1),
-    'file' => 'includes/wf_crm_admin_form.inc',
-    'weight' => 3,
-    'type' => MENU_LOCAL_TASK,
-  );
```
dsnopek, line 39: It seems sort of odd to remove this item, when none of this code should be executed anyway? Not really problem, though.

dsnopek (changes requested):
I skimmed through the code changes, and had a couple minor comments. None of that is a big deal.

However, I also did some manual testing, and the value for the custom field isn't actually getting saved in my testing.

I just created a simple alphanumeric field on Contacts, and then added it to the Webform. It gets shown on the form fine, and submission doesn't crash. However, the value for that custom field doesn't get saved in CiviCRM, and strangely not on the Webform Submission either.

includes/wf_crm_webform_ajax.inc
mglaman: That was intended. I fixed the one function call but left this to error for whenever it bubbles up.

mglaman:
I'm debugging why fields are not saving. When we hit `\Drupal\webform_civicrm\Plugin\WebformHandler\CivicrmWebformHandler::preSave` and call `\wf_crm_webform_postprocess::preSave`, the provided webform submission has the custom field values filtered out.

This is really weird. The values exist in the HTTP POST and form state input, but not in the values

In `\Drupal\Core\Form\FormBuilder::handleInputElement` the form state has no value, so it takes the element default:

```
    // Set the element's value in $form_state->getValues(), but only, if its key
    // does not exist yet (a #value_callback may have already populated it).
    if (!NestedArray::keyExists($form_state->getValues(), $element['#parents'])) {
      $form_state->setValueForElement($element, $element['#value']);
    }
```

For some reason the Form API does not like the custom elements..

🤦🏼‍♂️ because the elements were embedded with a `#value` attribute....

```
civicrm_1_contact_1_cg1_custom_1:
  '#type': textfield
  '#value': ''
  '#data_type': String
  '#form_key': civicrm_1_contact_1_cg1_custom_1
  '#extra':
    width: 20
  '#title': 'Test Field 1'
civicrm_1_contact_1_cg1_custom_2:
  '#type': textfield
  '#value': ''
  '#data_type': String
  '#form_key': civicrm_1_contact_1_cg1_custom_2
  '#extra':
    width: 20
  '#title': 'Testing Field 2'
```

mglaman:
@dsnopek addressed and reformatted the `@todos` for fieldsets.

This also fixes the values not saving properly. This was due to the custom field elements having the `#value` property filled vs `#default_value`.

webform_civicrm.module
dsnopek: @mglaman This one is still unaddressed. Is this change even needed for anything? It seems super arbitrary, because this hook_menu() will never be run. If we're gonna remove something from hook_menu() why not the whole thing? Or just leave it be?

dsnopek:
Huzzah! I had to delete and re-create my Webform, but after the latest changes my custom field saved correctly on the contact and in the Webform submission. I didn't test anything but an Alphanumeric field, so I don't know if other field types work or not, but I know there's a desire to get this merged so I don't think that need hold this up (we can fix the other field types in other PRs if necessary).

dsnopek (approved):

mglaman:
It was an out of scope fix. We've replaced that menu handler, but the other
ones need review still

PR #196, opened 2019-01-31
Only show custom reference option if such a field exists

colemanw:
Overview
We recently introduced the ability to use a custom reference field as the default value for an existing contact. This prevents a potential bug with the existing contact configuration form.

Before
The existing contact configuration form may fail validation and not save if no contact reference custom fields exist.

After
Form saves without errors.

See https://www.drupal.org/project/webform_civicrm/issues/3024260


colemanw:
@vinuvarshith FYI

colemanw:
Merging as reviewed by Oliver Gibson.

PR #197, opened 2019-02-13
Add financialType_name in param transfer checkout

rubofvil:
This param is used to add the description in some payment processors online and to be set correclty is required in method doPayment/doTransferCheckout.

I saw that is algo set in https://github.com/colemanw/webform_civicrm/blob/7.x-4.x/includes/wf_crm_webform_postprocess.inc#L2046, but in this point is not sending to the processor.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1926,6 +1926,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $params['amount'] = $params['total_amount'];
 
     $params['financial_type_id'] = $this->contribution_page['financial_type_id'];
+    
+    // Some payment processors expect this to be set (eg. redsys)
+    $financialTypeDetails = wf_civicrm_api('FinancialType', 'get', array('return' => "is_deductible,name",'id' => $this->contribution_page['financial_type_id']));
+    $params['contributionType_name'] = $params['financialType_name'] = $financialTypeDetails['values'][$this->contribution_page['financial_type_id']]['name'];
 
     $params['source'] = $this->settings['new_contact_source'];
     $params['item_name'] = $params['description'] = t('Webform Payment: @title', array('@title' => $this->node->title));


mattwire:
contributionType_name is a deprecated parameter, if a payment processor requires it then it is the payment processor that should be updated to use financialType

rubofvil:
Thanks @mattwire for the comment, i updated the title to make more clear. 
I added both "contributionType_name" and "financialType_name" to make more compatible.

KarinG:
Hi Ruben - I believe Matt means to say that this needs to be addressed in the payment processor extension code itself; payment processors typically don’t care if you use your VISA to buy donations, tickets, toques or mittens.

sluc23:
@KarinG our first approach was to include this param in payment processor code ([Redsys](https://github.com/ixiam/com.ixiam.payment.redsys))
But then we saw this line, 
https://github.com/colemanw/webform_civicrm/blob/7.x-4.x/includes/wf_crm_webform_postprocess.inc#L2046

```php
$financialTypeId = $params['financial_type_id'];
$financialTypeDetails = wf_civicrm_api('FinancialType', 'get', array('return' => "is_deductible,name",'id' => $financialTypeId));
// Some payment processors expect this to be set (eg. smartdebit)
$params['financialType_name'] = $financialTypeDetails['values'][$financialTypeId]['name'];
$params['financialTypeID'] = $financialTypeId;
```

which looks like it's been added for one specific payment processor needs. This implementation it's at half-way to send the param to the payment processor, so we thought on completing it.

Please confirm you don't want to include this feature in the module (that in my pov will be useful for all payment processors that need to have the **financial type label** as parameter),  and we'll code it into the Redsys source.

Tx

mattwire:
@sluc23 The problem we have in CiviCRM with payment processors is that parameters were never clearly defined and they are different depending on when you submit via event registration, contribution page and webform_civicrm.  So, less parameters is better as it is less work to ensure they remain supported.

The "ideal" way to handle this is to create more `get` functions directly in `CRM_Core_Payment` like this: https://lab.civicrm.org/extensions/stripe/blob/master/CRM/Core/Payment/Stripe.php#L920 (but I haven't submitted that because I didn't write a unit test yet!).

sluc23:
ok, moving this to the payment processor core then. We'll close the PR.
tx for the input

KarinG:
@sluc23 - sorry we could not get this in. But it is a better solution you handle this in the payment processor code.

PR #198, opened 2019-02-20
CATL-1032: Ability To Use Case Tags In Webforms

vinuvarshith:
Overview
We would like to give the user the ability to add case tags while filling a form.
Users can already use 'Tags' which are for contacts from the 'contact 1' tab in the webform configurations. This PR adds a similar feature in the Cases tab for Case Tags.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -54,12 +54,12 @@ function wf_crm_field_options($field, $context, $data) {
     elseif ($name == 'privacy') {
       $ret = wf_crm_get_privacy_options();
     }
-    elseif ($table === 'other' || $table === 'case') {
-      if ($field['table'] === 'tag') {
+    elseif ($table === 'other' || ($table === 'case' && strpos($name, 'tag_') === 0)) {
+      if (isset($field['table']) && $field['table'] === 'tag') {
         $split = explode('_', $name);
         $ret = CRM_Core_BAO_Tag::getTags("civicrm_{$ent}", $ret, wf_crm_aval($split, 1), '- ');
       }
-      elseif ($field['table'] === 'group') {
+      elseif (isset($field['table']) && $field['table'] === 'group') {
         $ret = wf_crm_apivalues('group', 'get', array('is_hidden' => 0), 'title');
       }
     }
@@ -806,16 +806,6 @@ function wf_crm_get_fields($var = 'fields') {
       'table' => 'group',
       'expose_list' => TRUE,
     );
-    $tagsets = array('' => t('Tag(s)')) + CRM_Core_BAO_Tag::getTagSet('civicrm_contact');
-    foreach ($tagsets as $pid => $name) {
-      $fields['other_tag' . ($pid ? "_$pid" : '')] = array(
-        'name' => $name,
-        'type' => 'select',
-        'extra' => array('multiple' => 1, 'civicrm_live_options' => 1),
-        'table' => 'tag',
-        'expose_list' => TRUE,
-      );
-    }
     $fields['activity_activity_type_id'] = array(
       'name' => t('Activity # Type'),
       'type' => 'select',
@@ -887,7 +877,9 @@ function wf_crm_get_fields($var = 'fields') {
         'integer' => 1,
       ),
     );
+    $tag_entities = array('other', 'activity');
     if (isset($sets['case'])) {
+      $tag_entities[] = 'case';
       $case_info = new CRM_Case_XMLProcessor_Process();
       $fields['case_case_type_id'] = array(
         'name' => t('Case # Type'),
@@ -940,17 +932,6 @@ function wf_crm_get_fields($var = 'fields') {
         'name' => t('Case # Details'),
         'type' => 'textarea',
       );
-      $tagsets = array('' => t('Tag(s)')) + CRM_Core_BAO_Tag::getTagSet('civicrm_case');
-      foreach ($tagsets as $pid => $name) {
-        if (!$pid) continue;
-        $fields['case_tag' . ($pid ? "_$pid" : '')] = array(
-          'name' => $name,
-          'type' => 'select',
-          'extra' => array('multiple' => 1, 'civicrm_live_options' => 1),
-          'table' => 'tag',
-          'expose_list' => TRUE,
-        );
-      }
       // Fetch case roles
       $sets['caseRoles'] = array('entity_type' => 'case', 'label' => t('Case Roles'));
       // Use the vanilla civicrm_api for this because it will throw an error in CiviCRM 4.4 (api doesn't exist)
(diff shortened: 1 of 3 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -1301,6 +1301,67 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+                'entity_table' => 'civicrm_case',
+              );
+              $i = 0;
+              foreach ($removeTags as $val) {
+                $remTagsParams['tag_id_'.$i] = $val;
+                $i++;
+              }
+              $result = wf_civicrm_api('EntityTag', 'delete', $remTagsParams);
+            }
+          }
```
colemanw: Wow that's a very long block of code being added here just to process tags. Isn't there already a helper function for this? If not there should be.

includes/utils.inc
```diff
@@ -54,7 +54,7 @@ function wf_crm_field_options($field, $context, $data) {
     elseif ($name == 'privacy') {
       $ret = wf_crm_get_privacy_options();
     }
-    elseif ($table === 'other') {
+    elseif ($table === 'other' || $table === 'case') {
```
colemanw: Looks to me like this will break all other pseudoconstant lookups for cases (as the `elseif ($table === 'case)` condition short-circuits all case fields.

Reading the code, I think we are actually safe to remove this outer conditional wrapping groups and tags, so we can just do `elseif ($field['table'] === 'tag') {` and `elseif ($field['table'] === 'group') {`

colemanw (changes requested):
I would like to see more reuse of code for handling tag saving, as it should be basically the same as for contacts.

Also, since activity tags are exactly the same as case tags, I think this PR ought to add support for both, and it should require almost no extra code.

colemanw:
Thanks for working on this @vinuvarshith. Let me know if those comments make sense.
Here's a gist for how to add the fields for all 3 types of tags without repeating any code.
https://gist.github.com/colemanw/a6621c4b08017f5d0e509830f2173667

includes/wf_crm_webform_postprocess.inc
colemanw: Found it. Look at function `wf_crm_webform_postprocess::addOrRemoveMultivaluedData`

includes/utils.inc
vinuvarshith: Actually I have fixed this in my local but didnt update the PR. I have updated this now. Let me know if this is better or can we just remove the outer ($table === 'other') wrapper.

vinuvarshith:
@colemanw 
Thank you for reviewing and for the gist.
I have addressed your comments and updated the code. Pls check and let me know if this is ok.
Thanks

includes/wf_crm_webform_postprocess.inc
vinuvarshith: I have updated the code to use 
wf_crm_webform_postprocess::addOrRemoveMultivaluedData
This also needed some changes to 'addOrRemoveMultivaluedData' function.

includes/utils.inc
```diff
@@ -54,12 +54,12 @@ function wf_crm_field_options($field, $context, $data) {
     elseif ($name == 'privacy') {
       $ret = wf_crm_get_privacy_options();
     }
-    elseif ($table === 'other') {
-      if ($field['table'] === 'tag') {
+    elseif ($table === 'other' || ($table === 'case' && strpos($name, 'tag_') === 0)) {
```
colemanw: This disallows us from getting activity tags.
Could you please take my previous suggestion and remove the outer `elseif`.

colemanw:
It looks like you've got activity tags exposed to the webform now but haven't yet added postprocessing for them. This is moving in the right direction though.

vinuvarshith:
@colemanw I have added support for activity tags too. Please check.
Thanks

includes/utils.inc
vinuvarshith: Updated as suggested.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1442,6 +1450,33 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+      }
+    }
+    $allTags = civicrm_api3('EntityTag', 'getoptions', [
+      'field' => "tag_id",
+    ])['values'];
+    $allTags = array_keys($allTags);
+    $removeTags = array_diff($allTags, $tagValues);
+    $removeTags = array_combine($removeTags, $removeTags);
+    $this->addOrRemoveMultivaluedData('tag', $entityType, $entityId, $tagValues, $removeTags);
+  }
```
colemanw, line 1475: I'm confused about what's going on with this function.
`addOrRemoveMultivaluedData` is sufficient to handle checkbox-style tagging widgets. It looks like you're trying to support something else - free tagging perhaps? Can you give more details about how this is being used?

colemanw (commented):
This PR is really looking good now! The code is nice & clean and the support for case/activity tags is a good feature.

However I'm not sure what exactly the `handleEntityTags` function is doing. Can you clarify?

colemanw (dismissed):

colemanw:
@vinuvarshith this is waiting on your response.

colemanw:
@vinuvarshith ping

includes/wf_crm_webform_postprocess.inc
jamienovick: @vinuvarshith Can you respond to coleman here please?

jamienovick:
@vinuvarshith Can you respond to coleman here please?

includes/wf_crm_webform_postprocess.inc
vinuvarshith: @colemanw
What I am trying to do here is to support adding and untagging. With 'addOrRemoveMultivaluedData' we pass $removeTags param which should have the tags to be removed.
**What I tried previously:**
So if we try to remove a tag from webform by unchecking the checkbox, that value is not recorded and so is not handled. Here I am getting all tag ids, calculating a diff between that tags to be added and all tags which I consider as tags to be removed.
**Updated logic:**
Now I have updated this code to get the values from $this->crmValues. Is this ok? Should I also get the checked tag values from $this->crmValues too instead of params?

vinuvarshith:
@colemanw 
I am so sorry. I somehow missed the notifications and completely forgot about this PR.
I have updated and commented about 'handleEntityTags'. Pls check and let me know if I can make it better.
Thanks for your time in reviewing this PR. 

Apologies once again

includes/wf_crm_webform_postprocess.inc
colemanw: Thanks for the clarification.
It still appears to me that you are reinventing the wheel here, and it's a much bigger wheel than necessary. The function `saveGroupsAndTags` calculates the `$add` and `$remove` values with 2 lines of code. Why not do it that way?

See https://github.com/colemanw/webform_civicrm/blob/b54ddfd6374a80089bbe03b12be55ac85c20786b/includes/wf_crm_webform_postprocess.inc#L860-L873

colemanw (changes requested):
I've commented on the `handleEntityTags` function and I think it can be further simplified.
Once that is done I can help you rebase this PR over to the 7.x-5.x branch as we're not going to be adding any new features to the 4.x branch.

includes/wf_crm_webform_postprocess.inc
vinuvarshith: Thanks for pointing this out. I have checked and fixed 'handleEntityTags' based on 'saveGroupsAndTags'. Tested in a clean setup and this works fine for me.
@colemanw

vinuvarshith:
@colemanw I have replied and updated the code based on your comments. Pls let me know how to proceed.

colemanw:
@jamienovick I have rebased this into the 5.x branch, and made some additional tweaks to it. Could you please have someone review/test #268

PR #199, opened 2019-02-22
Issue 2852365 prevent membership overwite

jaapjansma:
Overview
This PR fixes the issue over overwriting existing memberships. 

See https://www.drupal.org/project/webform_civicrm/issues/2852365

Before

There was an issue with the webform civicrm module overwriting existing membership records. 

After

Overwriting of existing membership is now a setting on the form, to overwrite an activity membership of the same type with the selected status. 

screenshot from 2019-02-22 14-57-12

In addition it is also possible to update an existing membership by the url. E.g. by passing membership1_1=_id_ for the first membership of the first contact. For the second membership of the first contact use _membership1_2_.

Notice for upgrading

When this fix is merged there should be a notice for upgrading existing installations. Installations which used webforms to renew existing memberships should do some extra configuration after upgrading. The configuration is setting the Update Existing Membership field to a set of configured statuses.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -973,14 +973,36 @@ class wf_crm_admin_form {
         $this->form['membership'][$c]['membership'][$fs] = array(
           '#type' => 'fieldset',
           '#title' => t('Membership !num for !contact', array('!num' => $n, '!contact' => wf_crm_contact_label($c, $this->data, 'wrap'))),
-          '#attributes' => array('id' => $fs, 'class' => array('web-civi-checkbox-set')),
-          'js_select' => $this->addToggle($fs),
         );
+        $this->form['membership'][$c]['membership'][$fs]["membership_{$c}_membership_{$n}_settings_existing_membership_status"] = array(
+          '#type' => 'select',
+          '#title' => t('Update Existing Membership'),
+          '#options' => array('' => '- ' . t('None') . ' -') + wf_crm_apivalues('membership', 'getoptions', array('field' => 'status_id')),
+          '#default_value' => wf_crm_aval($this->data, "membership:{$c}:{$n}_settings_existing_membership_status", array()),
+          '#multiple' => TRUE,
+        );
+        $this->help($this->form['membership'][$c]['membership'][$fs]["membership_{$c}_membership_{$n}_settings_existing_membership_status"], 'existing_membership_status');
+        $membership_type = wf_crm_aval($this->data, "membership:{$c}:membership:{$n}:membership_type_id");
         foreach ($this->sets as $sid => $set) {
-          if ($set['entity_type'] == 'membership') {
-            foreach ($set['fields'] as $fid => $field) {
-              $fid = "civicrm_{$c}_membership_{$n}_$fid";
-              $this->form['membership'][$c]['membership'][$fs][$fid] = $this->addItem($fid, $field);
+          if ($set['entity_type'] == 'membership' && (!$membership_type || empty($set['sub_types']) || in_array($membership_type, $set['sub_types']))) {
+            $fs1 = "membership_{$c}_membership_{$n}_fieldset_$sid";
+            if ($sid == 'membership') {
+              $pos = &$this->form['membership'][$c]['membership'][$fs];
+            }
+            else {
+              $pos = &$this->form['membership'][$c]['membership'][$fs]['wrap'];
+            }
+            $pos[$fs1] = array(
+              '#type' => 'fieldset',
+              '#title' => $set['label'],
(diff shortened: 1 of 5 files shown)


petednz:
@jitendrapurohit can you give this a test next week if no-one else gets to it first

KarinG:
@jaapjansma and @petednz - is this issue also happening if Membership Type in CiviCRM Core is configured as belonging to two separate orgs? Core will by default upsell/downsell Memberships that belong to the same org. I have Webforms that are selling more than 1 Membership at the time [e.g. add a Club Membership to your Provincial Membership] - and key to make that work was to have the Membership Types configured to belong to two separate orgs.

jaapjansma:
@KarinG It is still possible to do upselling, e.g. meaning chaning a membership from Silver to Gold. 
You can also use the webform to mix upselling and add new memberships. 

I have not tested this issue when the membership types belong to different organizations. That is not the case at the clients implementation.

jaapjansma:
@jitendrapurohit @petednz thanks for giving this a test!

KarinG:
Sorry let me reword my concerns more clearly:

CiviMember native CiviCRM-> does not allow more than one Membership per Contact - unless the Membership are of different Membership Types belonging to two different Organizations. If your org's name would be: MyOrg - then you need one Membership Type configured to belong to organization MyOrg and one Membership Type configured to belong to organization MyOrg-Magazines 

So yes we could decide to modify this behaviour in Webform CiviCRM module - allowing two Membership Types belonging to the same organization to exist on a single Contact record - but my concern is that this essentially means MyOrg should not be using any CiviCRM native front-end of back-office pages to sell or renew Memberships anymore because the unexpected existence of multiple Memberships of different Type but all belonging to the same MyOrg will no doubt case major issues that likely would result in loss of Membership data.

I like using Webform CiviCRM to do things that otherwise would be impossible to do in Core - to extend the capabilities/ways in which admins can use CiviCRM, but in this case we would be creating a structural/conceptual conflict.

jaapjansma:
I am not sure if I follow you. 
For example I have implemented an situation like this before:
Joe Smit has two memberships: one is the General Member and the other is the Magazin member. Both membership belong to the same organisation e.g the organisation MyOrg.

Are you saying that this is incorrect CiviCRM implementation?

mattwire:
@KarinG I don't think that's quite right. It's quite easy to create multiple memberships of both the same and different types (on the same org) in CiviCRM for a contact.

KarinG:
Ok I need proof :-) CiviCRM has always done upsell/downsell - of Membership Types belonging to the same org. I'd be happy if that's no longer the case.

jaapjansma:
With this PR you can still configure the form for upselling or downselling a specific membership. Meaning that in CiviCRM the membership type gets changed on the membership record.

KarinG:
I understand @jaapjansma - I just need to make sure that this will not cause data that will cause havoc in CiviCRM later when an admin tries to put that Membership type on a native contribution page or renews from civicrm back-end later on.

jaapjansma:
@KarinG How could I help you with getting this PR merged?

KarinG:
Hoi @JaapJansma - I have an idea - will DM you.

KarinG:
Hoi @jaapjansma,

Happy to schedule a GoTo for some time next week. Here's my summary:

**1. In CiviCRM Membership - Core:** if both CiviCoop - General Membership Type as well as CiviCoop - Magazines Membership Type belong to the Same Org in the Membership Type Config:

**a) I purchase a CiviCoop - General Membership:**
image

**b) back to the Contribution form trying to add on the CiviCoop - Magazines Membership:**
image

**c) result is -> my CiviCoop - General Membership is reTyped to a CiviCoop - Magazines Membership and the Expiry date is extended by +1y:**

image

**2. The only way to have two Memberships on a Contact's record is if the Membership Type is Configured with another org** -> look at it as a Department. So e.g. I reconfigure the Membership Types:

CiviCoop - General is with Org: CiviCoop 
CiviCoop - Magazines is with Org: CiviCoop - Publishing 
(a separate organization in CiviCRM)

**Then repeating the above steps results in the behaviour you want** where a Contact has two Memberships at the same time:

image

**3. En this is how things work right now with Webform CiviCRM** (using the Membership APIs). 

**4. So my concern is** -> sure it would be nice if we could have two Memberships on a contact record and both Memberships belonged to the same Org, but it introduces the risk that if an admin were to use a native Contribution page in future -> the behaviour would be core CiviCRM's and the Memberships would start blending together. 

**5. The workaround -> is to make your Magazines Membership attached to your Org's Publishing Department.** That's safe and effective and does not cause conflicts with how Core handles Membership.

jaapjansma:
Just discussed this PR with @KarinG and I came to the conclusion that this PR tries to fix something which should be fixed in CiviCRM it self. If it should be fixed (as she explained that the behaviour is intended).

PR #200, opened 2019-02-28
cherry pick of 'Fix getCreditCardCSSNames() which is deleted from core civi' from 7.x-4.x

johncronan:
This is identical to the fix in the 7.x branch, and is necessary for payment to work, with any Drupal 8 version.


KarinG:
Hey @johncronan - thanks for this! Yes, is required for CiviCRM versions > 5.7.x
[cc: @dsnopek @mglaman ]

PR #202, opened 2019-03-28
Use the api to load contact names instead of db query

colemanw:
Overview
This swaps a db query for an api call, and also improves the efficiency of `wf_crm_apivalues()`.


colemanw:
I've tested this out by stepping through values with the debugger and confirm there's no change.

KarinG:
Hi @colemanw please see: https://civicrm.stackexchange.com/questions/29144/undefined-function-in-webform-civicrm

PR #203, opened 2019-03-28
Another sql query bites the dust

colemanw:
Overview
Simple sql -> api swap.


colemanw:
Have verified this works in my debugger. The variable gets assigned the same value before/after the patch.

PR #205, opened 2019-03-29
Support larger number of max_instances for activities.

KarinG:
Supporting larger number of max_instances of activities for larger projects.


colemanw:
Looks good.

KarinG:
Thanks. It's on a live site happily creating up to 35 activities with one form submission.

colemanw:
Could be bumped up to 50 even. In theory these fields could be unlimited but in practice we tend to hit up against PHP's max allowed POST size when forms get too massive which is why I think reasonable limits are a good thing.

KarinG:
Absolutely.

KarinG:
required max_input_vars to be bumped up quite a bit :-)

PR #206, opened 2019-03-29
Prepopulate billing fields

colemanw:
Overview
Time-saver for users filling out the contribution form - prepopulates it with previously-entered name & address.

Before
Billing fields are always blank for anonymous users.

After
Billing fields are pre-filled with name & address entered on previous pages of the webform.

To Recreate
------------------

1. Create a webform with name & address on page 1, and contribution page on page 2.
2. Fill out form as anon user.
3. Without patch, billing fields on page 2 are blank.
4. With patch, billing fields on page 2 will reflect data entered on page 1.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -98,6 +98,9 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           }
         }
       }
+      if (!$this->billing_params && !empty($this->data['contact'][1])) {
+        $this->prefillBillingFields();
+      }
     }
     // Even though this object is destroyed between page submissions, this trick allows us to persist some data - see above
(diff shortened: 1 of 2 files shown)


aydun:
It works ok for the advertised case of anonymous users, but it causes problems for authenticated users.

For authenticated users, it loads the billing address and then overwrites the name and address fields that are present on the form which presents several issues:
1) the address fields on the form are used in preference to an existing billing address
2) if not all the address fields used by the billing block are also on the form the result is a mix of fields.  

It may be best to only do this prepopulating only if there is no existing billing address.

Also related, where the contact is specified by ?cid1=xxx on the url, the billing address loaded is that of the logged-in user, not the cid1.  See https://www.drupal.org/project/webform_civicrm/issues/3093045

PR #207, opened 2019-04-03
Fatal error with Activities, Fix tokens, Improved Select/Radio/Checkbox and live options support

mglaman:
Overview
Add support for Activities

Screen Shot 2019-05-01 at 11 49 44 AM

An Activity element allows for static and live options.

They persist. However, the label doesn't change to be the single value. (need to see if this is how D7 worked or not.)

Screen Shot 2019-05-01 at 11 50 21 AM

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1978,22 +1978,7 @@ class wf_crm_admin_form {
       if ($options = Utils::wf_crm_field_options($field, 'component_insert', $settings['data'])) {
         $field['options'] = $options;
         $field['extra']['items'] = wf_crm_array2str($options);
-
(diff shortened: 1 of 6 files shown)


KarinG:
Thank you @mglaman - lots here. I've tested it - works as you described (radio buttons when multiple activities and a single checkbox when one activity is enabled in the Element's config). Going to merge it in. 

Re: labels in D7: here's a screenshot of D7WFC - these are dozens of single (user select and only one option enabled) activities (with some CSS to align the checkboxes)

image

This is activity 4's config e.g. (I set &nbsp for the label of the webform element itself in this case; option Label is rendered on the form to the right of the checkbox - and that could be different from Item - Label = Item on first config but then users can override/change them)

image

PR #208, opened 2019-04-04
 use flattened webform elements (so that civicrm elements can be inside a flexbox)

johncronan:
When going through the webform elements to match against wf_crm_get_fields, we should use getElementsDecodedAndFlattened instead of just getElementsDecoded, so that flexboxes don't get in the way.


johncronan:
I tried to rebase this, so that it shows only my commit, but I couldn't figure it out. :/

Please let me know, if it is an issue. I would be happy to recreate the PR more carefully.

KarinG:
Hi @johncronan - can you please be more specific as to which CiviCRM elements? -> I am able to put e.g. CiviCRM Email and CiviCRM Address Elements into a Flexbox:

image

image

johncronan:
I found that I wasn't making it through the validate function in the postprocess .inc file, specifically where it calls out to validateBillingFields and then checks that there is an email present. So it may not bug out unless using a contribution.

KarinG:
Ah ok - can you get me a screenshot of that error/log message for good measure?

johncronan:
I'll attach here. You can go to dev.acreresidency.org/support/donate and use a test credit card number, if you're interested in seeing it in action.

Screen Shot 2019-04-04 at 6 29 07 PM

KarinG:
Excellent - thank you. I now understand the issue. Let me reproduce this.

KarinG:
Ok - I've reproduced the issue:
I put the email field in a flexbox and tried to make a contribution. Error: "An Email address is required to complete this transaction".  John uses Stripe. I hit the Error with iATS Payments as well. 

@johncronan 's PR Fixes the issue. I can now transact monies with my Email field nested in a flexbox. 

@dsnopek - is this the correct solution? Or should we look at fixing this in the validation itself?

mattwire:
@KarinG Was the issue here that the email field was being passed through to CiviCRM as an array? eg. [0 => test@test.com].  As I have an open issue on the stripe extension re. webform and email: https://lab.civicrm.org/extensions/stripe/issues/37

KarinG:
@mattwire - no I don't think that's related. The error that pops up is from the validation (still) on the webform side. And I also hit it with iATS Payments.

KarinG:
Hey @johncronan - if you could please close this PR and create a new one with just a single commit (with the edit to utils.inc) then I'll merge this in and you'll get the credits.

PR #209, opened 2019-04-05
Fixed file access issue according to new security update in CiviCRM.

agilewarealok:
Overview
Related to Issue: https://www.drupal.org/project/webform_civicrm/issues/3037226

When using Webform CiviCRM to create an Activity for a Case. Since CiviCRM 5.10.3, the file attachments can no longer be opened from a Case Activity using the Activity View. The URL to the file on the Activity no longer returns the file

**Steps to reproduce:**

1. Create an Activity for a CiviCRM Case using Webform
2. Upload and attach a file to the Activity using Webform
3. Open CiviCRM, locate the Activity and view the file
4. File URL does not work, file is not returned

Before
File URL does not work and file is not returned. 

After
File URL working as expected. 

Comments
_Agileware Ref: CIVIWF-3_

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -948,10 +948,11 @@ abstract class wf_crm_webform_base {
       $entity_id = $this->ent[$entity][$n]['id'];
     }
     if (!empty($file[$val])) {
+      $fileHash = CRM_Core_BAO_File::generateFileHash($entity_id, $val);
       return array(
         'data_type' => 'File',
         'name' => CRM_Utils_File::cleanFileName($file[$val]['uri']),
-        'file_url'=> CRM_Utils_System::url('civicrm/file', "reset=1&id={$val}&eid={$entity_id}", TRUE),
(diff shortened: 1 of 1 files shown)


colemanw:
This is a backwards-incompatible change with CiviCRM < 5.10 so I think we'd better go ahead and create a new 5.x branch of this module.

colemanw:
I've created the branch (something we've been discussing for a while but this was a good catalyst for getting it done), and changed this PR to be against the new 7.x-5.x branch. Is anyone able to test this out?

jitendrapurohit:
We weren't able to replicate the loading of file issue on the `View activity` page. It seems it forms the URL with the help of code inside civicrm which is already fixed to include file hash?

This problem occurs when an activity is loaded on the webform using the `aid` param passed in the URL.

eg `http://site-name.com/node/42?aid=688`

The attachment loaded from the activity has the URL created from the function fixed in this PR. I was able to replicate the error in this scenario and confirm that the change here fixes the URL generated.

Before this PR, it returns an error -

image

After applying the change, the file can be opened correctly. So +1 to get this merged in webform civi

Thanks @agilewarealok

jitendrapurohit (approved):

jusfreeman:
@jitendrapurohit thanks for taking time to test this one 👍

PR #210, opened 2019-04-23
Fix fatal error after contact merge

jitendrapurohit:
Overview
Fix fatal error after contact merge - https://www.drupal.org/project/webform_civicrm/issues/3044643

Before
Fatal Error after performing contact merge in civicrm

After
Works fine.

Technical Details
Webform civi code tries to update the old contact id in submitted data to new value. It made use of `webform_component`  table which is not present in the updated drupal webform. 

This PR indents to do the same using the updated code format.

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -469,10 +469,28 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
       ->condition('contact_id', '%-' . $old_id . '-%', 'LIKE')
       ->execute();
     // Update contact reference field data
-    db_query("UPDATE {webform_submitted_data} d, {webform_component} c SET d.data = :new
-      WHERE d.data = :old AND d.cid = c.cid AND d.nid = c.nid AND c.type = 'civicrm_contact'",
-      array(':new' => $new_id, ':old' => $old_id)
-    );
+    $webforms = \Drupal\webform\Entity\Webform::loadMultiple();
+    foreach ($webforms as $webform) {
+      $elements = $webform->getElementsInitializedAndFlattened();
+      foreach ($elements as $key => $value) {
+        if ($value['#type'] == 'civicrm_contact') {
+          $query = \Drupal::entityQuery('webform_submission')
+            ->condition('webform_id', $webform->id());
+          $submissions = $query->execute();
+          foreach ($submissions as $sid) {
+            //Load submission and update old contact id with the new id.
+            $submission = \Drupal\webform\Entity\WebformSubmission::load($sid);
+            $data = $submission->getData();
+            if (!empty($data[$key]) && $data[$key] == $old_id) {
+              $data[$key] = $new_id;
+              // Set submission data.
+              $submission->setData($data);
+              $submission->save();
+            }
+          }
+        }
+      }
+    }
   }
 }
 


jackrabbithanna:
I'm sure that this code does what it says it will do. I am a little hesitant to recommend this course of action. In this particular case, a direct query may be better IMO because it is much more efficient. The problem with this solution is scaling. What happens if you have 100 webforms or 1 million submissions. That may seem a little extreme, and maybe it is, but even with 1 webform and 100K contacts, the submissions can add up real quick. Merging contacts can already get to be sort of intensive, and now add on top of that this sifting through all webforms and all submissions. There can be arguments against a direct query of course, updating the submission in that way will not invoke any hooks or trigger any events. Perhaps a webform could have a setting where the admins get to choose if the submissions are even updated or not, maybe it defaults to "do". Maybe even the method of updating could be chooseable...but that seems like some complex stuff to try to explain with good UI/UX. Regardless to what the stakeholders decide to do, it's not bad to have these types of internal discussions and think about scaling, and choice.

jitendrapurohit:
Agree with the above thought @jackrabbithanna I've recently started using d8 with webform and looks like both the table that were used before are removed from webform now.

I think the table that we need to update is `webform_submission_data` where `value` column matches the old contact id and `name` column should end with `_contact_existing` string. Sounds a little hackish but I think it should not affect any other scenarios. Previous comparison of "type = civicrm_contact" is not possible  I guess.

jackrabbithanna:
The type = civicrm_contact condition was against the webform_components table in the query for D7 version. I'm just starting to get involved with the D8 version of this project so I have some catching up and research to do.

webform_civicrm.module
```diff
@@ -469,10 +469,28 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
       ->execute();
     // Update contact reference field data
-    db_query("UPDATE {webform_submitted_data} d, {webform_component} c SET d.data = :new
-      WHERE d.data = :old AND d.cid = c.cid AND d.nid = c.nid AND c.type = 'civicrm_contact'",
-      array(':new' => $new_id, ':old' => $old_id)
-    );
+    $webforms = \Drupal\webform\Entity\Webform::loadMultiple();
+    foreach ($webforms as $webform) {
+      $elements = $webform->getElementsInitializedAndFlattened();
+      foreach ($elements as $key => $value) {
```
mglaman, line 475: See `wf_crm_admin_form::handleDynamicCustomField`.

```
    /** @var \Drupal\webform\WebformInterface[] $webforms */
    $webforms = \Drupal\webform\Entity\Webform::loadMultiple();
    foreach ($webforms as $webform) {
      $handler_collection = $webform->getHandlers('webform_civicrm');

      if (!$handler_collection->has('webform_civicrm')) {
        continue;
      }
```

webform_civicrm.module
```diff
@@ -469,10 +469,28 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+        if ($value['#type'] == 'civicrm_contact') {
+          $query = \Drupal::entityQuery('webform_submission')
```
mglaman, line 477: Entity query is deprecated for retrieving the query directly from the entity storage.

mglaman (changes requested):

petednz:
@jitendrapurohit is this something you need to be reviewing?

KarinG:
Closing this as I think this issue has been resolved with: https://github.com/colemanw/webform_civicrm/pull/283

PR #211, opened 2019-04-30
Fix removal of civicrm handler when a field is unselected

jitendrapurohit:
Overview
Unselecting a single field from a civi setting page on a webform results in all civi fields being lost

Before
To replicate -

- Create a webform with 3 basic contact fields - first name, last name and job title.
- Save.
- Open civicrm tab and unselect job title and re-save.
- The complete civicrm fields are removed from the webform.

After
works fine.

Comments
Drupal Issue - https://www.drupal.org/project/webform_civicrm/issues/3051616

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -177,6 +177,7 @@ class wf_crm_admin_form {
     $this->form['delete']['#value'] = t('Remove Fields and Save Settings');
     $this->form['disable']['#value'] = t('Leave Fields and Save Settings');
     $this->form['cancel']['#value'] = t('Cancel (go back)');
+    $this->form['form_state'] = $this->form_state;
     return $this->form;
   }
 
@@ -1617,9 +1618,8 @@ class wf_crm_admin_form {
       foreach ($delete_me as $key => $id) {
         list(, $c, $ent, $n, $table, $name) = explode('_', $key, 6);
         $info = '';
-        if ($ent == 'contact' || $ent == 'participant') {
-          $info = '<em>' . wf_crm_contact_label($c, wf_crm_aval($this->node, 'webform_civicrm:data'));
-        }
+        $element = $this->webform->getElement($id);
+        $info = '<em>' . $element['#title'];
         if ($info && isset($this->sets[$table]['max_instances'])) {
           $info .= ' ' . $this->sets[$table]['label'] . ' ' . $n;
         }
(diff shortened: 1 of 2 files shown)


KarinG:
Reproduced the issue - it looks like the problem is that when unselecting a property of e.g. contact_1 -> it's remove all contact_1 fields - not just civicrm_1_contact_1_contact_job_title

image

includes/wf_crm_admin_form.inc
```diff
@@ -177,6 +177,7 @@ class wf_crm_admin_form {
     $this->form['delete']['#value'] = t('Remove Fields and Save Settings');
     $this->form['disable']['#value'] = t('Leave Fields and Save Settings');
     $this->form['cancel']['#value'] = t('Cancel (go back)');
+    $this->form['form_state'] = $this->form_state;
```
mydropwizard-support: Stashing the full FormState object on the form doesn't feel right. If we have to stash something on the form, maybe something like:

```
$this->form['#form_state_values'] = $this->form_state->getValues();
```

Or, ideally, if there were any other way to pass the necessary values to the right place without putting them on the form, that'd be best.

dsnopek, line 180: Stashing the full FormState object on the form doesn't feel right. If we have to stash something on the form, maybe something like:

```
$this->form['#form_state_values'] = $this->form_state->getValues();
```

Or, ideally, if there were any other way to pass the necessary values to the right place without putting them on the form, that'd be best.

mglaman: This shouldn't be done at all

src/Form/WebformCiviCRMSettingsForm.php
```diff
@@ -70,6 +70,10 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
     $webform = $this->getWebform();
     $handler_collection = $webform->getHandlers('webform_civicrm');
     $values = $form_state->getValues();
+    if (!isset($values['nid']) && !empty($form['form_state'])) {
```
mglaman, line 73: We have the form state, we should use it instead of fetching from the form object.

mglaman: I see. the form state somehow has different values. probably the ones from the confirm form style setup.

mglaman: See https://github.com/colemanw/webform_civicrm/pull/212/commits/d45f8518444656da862d5ec8059bb2bf2605ea9e

includes/wf_crm_admin_form.inc
mglaman: Somewhere we already stash the original valuess in `vals` in form state storage where we can pull them.

PR #212, opened 2019-05-03
Properly handle the confirmation form when a field is removed.

mglaman:
Builds on #211. Can't push there. Adds commit to handle submission.


KarinG:
@mglaman thank you! I tested this just now and it is working well - it definitively solves https://www.drupal.org/project/webform_civicrm/issues/3051616 and I also tested for multiple contacts that it only removes the indicated field for the Contact specified [and keeps it for other Contacts]. 

The only improvement I can think off is that we add the form_key to the screenshot below (this is helpful when there are multiple contacts)-> so in my case that would then have been:

Job Title - civicrm_2_contact_1_contact_job_title

image

PR #214, opened 2019-05-08
Adjust the checkbox label to be the option's label

mglaman:
Overview
When a single option is chosen for a CiviCRM Options element, it converts to a checkbox. The checkbox label, however, is the element's title - not the option. For activities this means the checkbox is always labeled "Activity Type". Now it matches the value.

Before
Screen Shot 2019-05-07 at 10 09 28 PM

After
Screen Shot 2019-05-07 at 9 59 49 PM

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1472,28 +1472,27 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   * Process the submission into the details of the activity.
   */
  private function formatSubmissionDetails(&$params, $activity_number) {
-   /*
     // Format details as html
     $params['details'] = nl2br(wf_crm_aval($params, 'details', ''));
     // Add webform results to details
     if (!empty($this->data['activity'][$activity_number]['details']['entire_result'])) {
+      /*
+      // @todo port
       module_load_include('inc', 'webform', 'includes/webform.submissions');
       $submission = webform_submission_render($this->node, $this->submission, NULL, 'html');
       $params['details'] .= drupal_render($submission);
+      */
     }
     if (!empty($this->data['activity'][$activity_number]['details']['view_link'])) {
-      $params['details'] .= '<p>' . l(t('View Webform Submission'), "node/{$this->node->nid}/submission/{$this->submission->sid}", array(
-          'absolute' => TRUE,
-          'alias' => TRUE
-        )) . '</p>';
+      $params['details'] .= '<p>' . $this->submission->toLink(t('View Webform Submission'), 'canonical', [
+        'absolute' => TRUE,
+      ])->toString() . '</p>';
     }
     if (!empty($this->data['activity'][$activity_number]['details']['edit_link'])) {
-      $params['details'] .= '<p>' . l(t('Edit Submission'), "node/{$this->node->nid}/submission/{$this->submission->sid}/edit", array(
-          'absolute' => TRUE,
-          'alias' => TRUE
-        )) . '</p>';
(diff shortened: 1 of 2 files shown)


KarinG:
The label looks great - but on Submit I get a illegal choice in the last Activity on the form (if I have 4 it's the 4th - in this screenshot it's the 2nd (Phone Call);

image

image

My form is quite complex (also has Contribution and other bits on it) - are you able to submit your form and create two Activities in the Contact's CiviCRM record?

mglaman:
🤔 I'll make a more complex form and see.

KarinG:
I made a simple form -> and getting the same error:

image

mglaman:
🤦‍♂ I just realized why! Because the form thinks it is supposed to be a select list, but we are altering it to be a checkbox.

src/Plugin/WebformElement/CivicrmOptions.php
```diff
@@ -148,6 +148,11 @@ public function prepare(array &$element, WebformSubmissionInterface $webform_sub
     // A single static radio should be shown as a checkbox
     if (!$use_live_options && !$as_list && count($element['#options']) === 1) {
       $element['#type'] = 'checkbox';
+      // Reset the element label, the checkbox label, to be the first option's value.
+      $element['#title'] = reset($element['#options']);
+      // Remove the options array to prevent invalid validation.
+      // @see \Drupal\Core\Form\FormValidator::performRequiredValidation
+      unset($element['#options']);
```
mglaman, line 160: Keeping this around was causing the following logic check to pass

```
if (isset($elements['#options']) && isset($elements['#value'])) {
```

which caused the validation failure.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1472,6 +1472,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   * Process the submission into the details of the activity.
   */
  private function formatSubmissionDetails(&$params, $activity_number) {
+   /*
```
mglaman: Commented this out. We can keep, just need to generate the correct route. But now it will submit.

mglaman (commented):
@KarinG you should be able to test now. I'm going to give it some more work. It should submit. I need to check multiple activities more.

mglaman:
@KarinG bonus: fixed some other bugs I missed (checkbox return value, was always `0` or `1`, now it is the proper activity ID.) And submission view or edit links are now in the activity details.

The missing piece is

```
      module_load_include('inc', 'webform', 'includes/webform.submissions');
      $submission = webform_submission_render($this->node, $this->submission, NULL, 'html');
      $params['details'] .= drupal_render($submission);
```

I'm going to push a commit for it, but due to some constraints I won't be able to test it right away.

KarinG:
Fantastic - I’ll test it!

KarinG:
Thanks @mglaman 

I've run some real use case scenarios on it:

**1. Confirming multiple activities** [2 activities / checkboxes - as described above in https://github.com/colemanw/webform_civicrm/pull/214#issuecomment-490701805 ] can be submitted and are recorded properly in CiviCRM:

image

**2. I've also tested 3 Activities: 2 to display as checkboxes and 1 to display as a select** [and a subject as a bonus] -> this is working too 👍 

image

image

**3. I'm going to merge this as it's working. One thing I'd like you to look at is adding an option to display a CiviCRM Options list as a Select list or a Radio button (as it is now).** That makes for much nicer forms if the list is long. In D7WFC -> this is (when editing the webform element) under Display options -> 
image

mglaman:
@KarinG yeah that Listbox was a D7 option of Webform whereas D8 has you select the type of related element. We'll need to just add it as an element setting

PR #215, opened 2019-05-08
use flattened webform elements (so that civicrm elements can be inside a flexbox)

johncronan:
When going through the webform elements to match against wf_crm_get_fields, we should use getElementsDecodedAndFlattened instead of just getElementsDecoded, so that flexboxes don't get in the way.

This is rebased version of #208.


KarinG:
Thanks!

PR #219, opened 2019-05-20
Sync state/province with the country on the form

kainuk:
Overview
Synchronizes the states (or the provinces, or the departments) with the country when both address fields are on the form

Before
The state select list is filled when the form opens with the state list of the default country (on a update it is the statelist of the selected country)

After
When the country is updated the state list is also updated.

Technical Details
Just removed some obsolete brackets

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -341,9 +341,10 @@ var wfCivi = (function ($, D, drupalSettings) {
   }
 
   function countrySelect() {
+
     var name = parseName($(this).attr('name'));
     var countryId = $(this).val();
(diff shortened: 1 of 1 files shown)


KarinG:
Thanks! @kainuk - Billing Name and Address State/Province happily changes with Country but the elements as put on the form - are still disconnect.  

My default country is United States -> 

swapping that to Country = Netherlands in Billing Name and Address -> all is well -> the Dutch provinces are loaded and I can select one

swapping that to Country (in the field below the Billing Name and Address - [this field is put on the form by selecting the checkbox for Country in the CiviCRM -> Contact tab]: I can select Country = Netherlands but I still have the US States (on the right)

image

KarinG:
Ok - I got it working on a fresh instance [also with Bartik] - that's good enough for me. Thanks @kainuk - for you first D8WFC PR!

image

KarinG:
I gave it another test run - and this is working well. Thanks @kainuk !

PR #220, opened 2019-05-21
Add `aslist` as Listbox support

mglaman:
Overview
Adds "Listbox" option for CiviCRM Options element

Allows controlling the output of the country field, for example, from <select> to <radios> or back.


KarinG:
There is is -> 
image

And it works great:
image

PR #221, opened 2019-05-23
Cleanup contact relationship tab in the administration form

kainuk:
Overview
When more contacts are selected in the administration form the option is offered to create a relationship. In rendering this screen not all the variables are substituted.

Before

1. In the label HTML markup is shown.
2. In the select the placeholders are shown instead of the contacts

before

After
after


KarinG:
Nice @kainuk - tested:

image

PR #222, opened 2019-05-27
WIP - Fix to ensure add element button works

stesi561:
Overview
TODO

Before
Clicking on the Add element doesn't work properly in some instances.
image

After
Clicking on the Add element works as expect.

Technical Details
TODO

@jitendrapurohit has made this fix however is currently away. Posting here in case someone else also finds this issue in the interim.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -506,6 +506,7 @@ function wf_crm_array2str($arr) {
  *   Result of API call
  */
 function wf_civicrm_api($entity, $operation, $params) {
+  \Drupal::service('civicrm')->initialize();
   $params += array(
     'check_permissions' => FALSE,
     'version' => 3
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -9,6 +9,8 @@ use Drupal\webform\Plugin\WebformElementBase;
 use Drupal\webform\WebformSubmissionInterface;
 use Drupal\webform_civicrm\Utils;
 
+include_once __DIR__ . '/../../../includes/utils.inc';
+
 /**
  * Provides a 'civicrm_options' element.
  *


includes/utils.inc
```diff
@@ -506,6 +506,7 @@ function wf_crm_array2str($arr) {
  *   Result of API call
  */
 function wf_civicrm_api($entity, $operation, $params) {
+  \Drupal::service('civicrm')->initialize();
```
mglaman, line 509: This needs to go into `CivicrmOptions::prepare` which is being invoked for some reason during the add element process.

src/Plugin/WebformElement/CivicrmOptions.php
```diff
@@ -9,6 +9,8 @@
 use Drupal\webform\WebformSubmissionInterface;
 use Drupal\webform_civicrm\Utils;
 
+include_once __DIR__ . '/../../../includes/utils.inc';
```
mglaman, line 12: Valid! Per the get field options method which invokes the helper function.

mglaman (changes requested):
We can prevent a lot of initialization calls and checks if we adjust the invocation to when it is needed

includes/utils.inc
stesi561: Do you want me to update and test this or is this something you have already done?

KarinG: I’ll do that - and test it -I was able to reproduce this issue

mglaman: :) thanks @KarinG. Only the author and maintainers can push to PRs

jitendrapurohit: Is it already modified and pushed as a PR? would be happy to do a quick test if it is needed :)

KarinG: @jitendrapurohit - sorry - no I don’t think I got to it before holidays and still catching up. If you can make the edit - I promise it’s top of the list for me to test / merge.

jitendrapurohit: I've submitted a new one - https://github.com/colemanw/webform_civicrm/pull/242

KarinG:
Closing in favour of: https://github.com/colemanw/webform_civicrm/pull/242

PR #223, opened 2019-06-06
d8WC - Fix issue in adding file fields on webform

jitendrapurohit:
Overview
File fields are not added on d8 webform.

Before
To replicate -
- Create a custom field in civi with type = file.
- Create a webform and add this field from the civi settings page.
- The file field is incorrectly added with type = generic element.

image

After
Delete the incorrect element and add the file field again after this patch. It is added correctly.

image


KarinG:
@jitendrapurohit - tested/confirmed/merged :-)

PR #227, opened 2019-06-17
Fix fatal error when custom fields are Edit-ed after they have been placed on the webform

jitendrapurohit:
Overview
Fatal error fixes when custom fields are added to the webform

Before
> Error: Class name must be a valid object or a string in civicrm_api3_generic_getoptions()

Above error is noticed on the webform when custom fields are added to it. To replicate -

- Add custom field with type select on the webform.
- Edit the component from the webform page.

After
Fixed.


KarinG:
Ah reproduced it ->

The custom field are successfully added to the webform:

image

image

But when I hit Edit to configure them - yep: "Error: Class name must be a valid object or a string in civicrm_api3_generic_getoptions()"

Sorry - looks like this PR needs a rebase.

jitendrapurohit:
looks like this is handled in the latest update. I'm not able to replicate this now. closing.

KarinG:
Awesome!

PR #228, opened 2019-06-17
Use api for wf_crm_state_abbr function

kainuk:
Overview
The wf_crm_state_abbr function used a direct DB query. Now it uses only the API.

Technical Details
Nothing is changed in functionality

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -23,19 +23,26 @@ use Drupal\webform_civicrm\Utils;
  * @return string or integer
  */
 function wf_crm_state_abbr($input, $ret = 'abbreviation', $country_id = NULL) {
-  $input_type = $ret == 'id' ? 'abbreviation' : 'id';
-  $sql = "SELECT $ret FROM civicrm_state_province WHERE $input_type = %1";
-  $vars = array(1 => array($input, $ret == 'id' ? 'String' : 'Integer'));
+  $params = [];
+  if ($ret == 'id') {
+    $params['abbreviation'] = $input;
+  }
+  else {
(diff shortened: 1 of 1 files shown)


colemanw:
Thanks for this @kainuk - and congrats on your 1st PR to the project :tada:

includes/utils.inc
```diff
@@ -23,19 +23,26 @@ use Drupal\webform_civicrm\Utils;
...
+    $params['id'] = $input;
+  }
   if ($ret === 'id') {
     if (!$country_id || $country_id === 'default') {
-      $config = CRM_Core_Config::singleton();
-      $country_id = (int) $config->defaultContactCountry;
+      $settings = wf_civicrm_api('Setting', 'get', [
+        'sequential' => 1,
+        'return' => 'defaultContactCountry',
+      ]);
```
colemanw, line 38: So in the 7.x-5.x branch we have a function `wf_crm_get_civi_setting()` which was introduced in 3da7c660406217a23d759dcf2772b7221aafec11. Looks like that hasn't yet been ported to the 8.x branch but it would be nice. Not a blocker to this PR tho.

mglaman (approved):
+1 on this end.

includes/utils.inc
colemanw: I've submitted a PR to port those changes so actually this PR will need rebasing once #239 is merged.

colemanw:
I went ahead and committed a version of this to the 7.x-5.x branch, and then cherry-picked that into #239 so it would merge cleanly.

PR #229, opened 2019-06-17
Better feedback on removing CiviCRM fields

kainuk:
Overview
On deleting CiviCRM fields, confirmation of the user is asked. The description of this field is improved (or made the same as in Drupal 7 webforms).

To replay:
------------------------------------------------------------------------------------------
Create a form with 2 contacts. Label the contact Parent and the second contact Child. Add an address to the first contact. Start deleting fields (so you get the feedback). Delete
* The gender of the parent
* The external id of the child
* The city of the address

Before
before
The label of the contact is not used, which makes it less clear which field it is.

After
after

Technical Details
This PR also removes the use of `$this->node->` on line 1628, that causes a warning

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1620,12 +1620,12 @@ class wf_crm_admin_form {
         list(, $c, $ent, $n, $table, $name) = explode('_', $key, 6);
         $info = '';
         $element = $this->webform->getElement($id);
-        $info = '<em>' . $element['#title'];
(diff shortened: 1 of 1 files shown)


KarinG:
Great - confirmed/tested -> and merging.

image

kainuk:
;-)

mglaman:
@KarinG was this merged?

KarinG:
Oops - forgot to actually hit Merge. Sorry fellows!

PR #230, opened 2019-06-19
Converted wf_crm_get_matching_rules to api only

kainuk:
Overview
Another function that is converted to api only

To Test
-----------------------------------------
This is the place where the function is shown in the interface
matching-rule

This PR can be tested with The function can be tested with https://github.com/colemanw/webform_civicrm/pull/233 .

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -287,10 +287,10 @@ function wf_crm_get_matching_rules($contact_type) {
   static $rules;
   $contact_type = ucfirst($contact_type);
   if (!$rules) {
-    $rules = array_fill_keys(array('Individual', 'Organization', 'Household'), array());
-    $dao = CRM_Core_DAO::executeQuery('SELECT * FROM civicrm_dedupe_rule_group');
-    while ($dao->fetch()) {
-      $rules[$dao->contact_type][$dao->id] = $dao->title;
+    $rules = array_fill_keys(['Individual', 'Organization', 'Household'], []);
+    $values = wf_crm_apivalues('RuleGroup', 'get');
+    foreach ($values as $value) {
+      $rules[$value['contact_type']][$value['id']] = $value['title'];
     }
   }
   return $rules[$contact_type];


mglaman (approved):
Looks good to me, code-wise. Did not manually test but I also wouldn't be able to easier test.

colemanw:
I've tested this out by backporting it to the 7.x-5.x branch and have cherry-picked it in. Works great.
I just have a small code-style nitpick I'd suggest before merging this.

includes/utils.inc
```diff
@@ -287,10 +287,10 @@ function wf_crm_get_matching_rules($contact_type) {
   static $rules;
   $contact_type = ucfirst($contact_type);
   if (!$rules) {
-    $rules = array_fill_keys(array('Individual', 'Organization', 'Household'), array());
-    $dao = CRM_Core_DAO::executeQuery('SELECT * FROM civicrm_dedupe_rule_group');
-    while ($dao->fetch()) {
-      $rules[$dao->contact_type][$dao->id] = $dao->title;
+    $rules = array_fill_keys(['Individual', 'Organization', 'Household'], []);
+    $values = wf_crm_apivalues('RuleGroup', 'Get', []);
```
colemanw: Typically the api action is all lowercase, and this function's 3rd param already defaults to `[]` so I'd write this line as

```suggestion
    $values = wf_crm_apivalues('RuleGroup', 'get');
```

KarinG: Hoi Klaas: I tried to commit Coleman's suggestion (a new button popped up in GitHub) - but I don't have permissions - which is odd b/c I know I can do that from the command line.

colemanw:
@KarinG I've rebased in the suggested change. Good to merge.

KarinG:
Awesome!

includes/utils.inc
kainuk: Hi @KarinG, I see you already solved it (I was on holiday in Italy without laptop)

PR #231, opened 2019-06-22
WIP: line item total text from label fix, membership line items display amount updates

jackrabbithanna:
Overview
Updates line item table to show membership line items for membership fee fields. Updates contribution line item text to pull from contribution amount field label.
Seems to make memberships connect to contributions fine, MembershipPayment Records are created (although did this really make that work?)

Sorry about the fields being black background in the screenshots, that's my Gnome/GTK theme being wonky
The membership fee field has a value of 15
The membership fee 2 field has a value of 25
Contribution amount field value of 0

Not sure about why table line items listing is backwards from field order.

Before
Membership line items don't show and totals don't update when fee fields are changed.
membership-line-item-table-before

After
Membership line items show and totals update
membership-line-item-table-after

Contribution contains membership line items.
Membership records tied to contributions with MembershipPayments

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -461,6 +461,10 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
           if ($table == 'contribution' && in_array($name, array('line_total', 'total_amount'))) {
             $element['#attributes']['class'][] = 'contribution-line-item';
           }
+          // For membership line-items
+          if ($table == 'membership' && in_array($name, array('fee_amount'))) {
+            $element['#attributes']['class'][] = 'contribution-line-item';
+          }
           // Provide live options from the Civi DB
           if (!empty($element['#civicrm_live_options']) && isset($element['#options'])) {
             $params = [
@@ -571,10 +575,11 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
   private function displayLineItems() {
     $rows = [];
     $total = 0;
+    $elements = $this->node->getElementsDecoded();
+
     // Support hidden contribution field
     $fid = 'civicrm_1_contribution_1_contribution_total_amount';
     if (!$this->line_items && isset($this->enabled[$fid])) {
-      $elements = $this->node->getElementsDecoded();
       $field = $elements[$this->enabled[$fid]];
       if ($field['#type'] === 'hidden') {
         $this->line_items[] = [
@@ -583,6 +588,24 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
           'element' => 'civicrm_1_contribution_1',
           'label' => !empty($field['name']) ? $field['name'] : t('Contribution Amount'),
         ];
+      }
+    }
+
+    // add membership line items if membership fee form elements are hidden
+    for ($c=1; $c<=30; $c++) {
+      for ($m=1 ; $m<=9; $m++) {
+        $membership_fid = 'civicrm_' . $c . '_membership_' . $m . '_membership_fee_amount';
+        if (!empty($this->enabled[$membership_fid]) && !empty($elements[$this->enabled[$membership_fid]])) {
+          if ($field['#type'] === 'hidden') {
+            $field = $elements[$this->enabled[$membership_fid]];
+            $this->line_items[] = [
+              'line_total' => $field['value'],
+              'qty'        => 1,
+              'element'    => 'civicrm_' . $c . 'membership_' . $m,
+              'label'      => !empty($field['name']) ? $field['name'] : t('Membership Fee'), // update for membership type id in label
+            ];
+          }
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
So this is not the result of figuring out how it worked in the D7 version. 
In the D7 version a line item will appear the line item listing table, with the membership type listed as the text. 

The code above doesn't do that, but does work with multiple relationships, renewals update the end dates. 

If this lil bit of code makes it work here in this PR, and if this is not how it's done in the D7 .. then there must be just some other lil bit of code that needs updating .. because the post processing works out fine here....

I'm trying to figure out what was missing the D8 version that was in the D7 version to  make membership_fee_amount fields trigger a line item in the line item listing.

In this code, all we're doing is setting a class for the membership_fee_amount option, which then triggers javascript to include an entry for it in the table, and on submit, that is good enough to have post processing work.

includes/wf_crm_webform_preprocess.inc
```diff
@@ -461,6 +461,10 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
           if ($table == 'contribution' && in_array($name, array('line_total', 'total_amount'))) {
             $element['#attributes']['class'][] = 'contribution-line-item';
           }
+          // For membership line-items
+          if ($table == 'membership' && in_array($name, array('fee_amount'))) {
```
jackrabbithanna, line 465: This bit is what actually made it work.

js/webform_civicrm_payment.js
```diff
@@ -100,7 +100,7 @@ cj(function($) {
   function calculateLineItemAmount() {
     var fieldKey = $(this).data('civicrmFieldKey'),
       amount = getFieldAmount(fieldKey),
-      label = $(this).closest('div.webform-component').find('label').text() || Drupal.t('Contribution'),
+      label = $('label[for=' + this.id + ']').text() || Drupal.t('Contribution'),
```
jackrabbithanna, line 103: This bit here is to pull the text in the line item table from the label of the input element for the webform field.

jackrabbithanna:
It may help me if you do test this, even though a WIP and adding new changes that weren't in the D7 version, and detect any differences in how the data is setup on form submit. is there any minor detail about the accounting wrong? If now then this is totally a form field options or preprocess issue, or javascript

I'm not figured out yet how the membership fee amount was getting into the list of line items in the D7 version. 
This is the commit that brought the membership fee field into the codebase:
https://github.com/colemanw/webform_civicrm/commit/54ad5e6100ad300d9a0a6e391dc2cf8685c16c42

Every single bit there, is the same in D8 version (the bit in D7 in  includes/utils.inc is now in D8 in src/Plugins/Fields.php)

There is definitely javascript that reads for form elements with class .contribution-line-item .. and then the javascript includes them in the table of line items list. 

It would make sense to me that there be "one" way of adding line items...but definitely that's not the case in D7 version.

I'll keep looking but it's non-obvious  ... Pretty much everywhere I look trying to see the Field or Field options are setup for membership_fee_amount, they are the same in the D8 version as D7 ...

It's got to be in field, field options, or preprocessing, or javascript where the problem is to do an exact copy over from how it was done in the D7 version
This is because in displayLineItems() method, there were line item values in $this->line_items to iterate over to add to the table . It's only by adding a class to the membership fee amount that the javascript would add the line items.
I haven't seen any membership specfic javascript for adding membership line items to the table list
Also once, there is a line item in the table, all the post processing for the membership starts to work out great...

Basically the way I'm adding the line item for memberhsip is the same way coleman did it when he added the "Additional line items"
He added support for additional line items here:
https://github.com/colemanw/webform_civicrm/commit/925cdcf930a9c77f7016c5c434eb1aace77b7a86
Then he added the field options and js to do that tally here:
https://github.com/colemanw/webform_civicrm/commit/20833779e4762d39e852dab9e19f36128d93fce3

So membership stuff must be much older, and done a different way...

There's got to be really only on lil thing missing in D8, but it is seeming likely to me now that something is just different. 

I'm also concerned that the page break isn't working, or I wonder if that change is affecting the thing. Like I'm wondering if the page break worked, would the line item table pick everything up right? IDK, just a thought in my brain. 

In D7, it warns that you should put the Membership Fee field BEFORE the payment page break. 
"In order for line-items to be displayed correctly to the user, it is recommended to add a page break in-between Contribution Billing Fields and the following:"

With it not working in D8, perhaps the reason there were no membership line items, is because that would happened in validation / submission of page 1...

BUT with this new way here in this code, it would work when the Membership Fee is on the same page as contribution amount, AFTER the page break...

MAYBE we should have this PR and THEN also when the page break works, it would work as it did before. when there are two separate pages?

Additional line item show on the second page, after the page break, which necessitates the javascript that adds them to the line item table listing.

So perhaps it's better we put the membership fee amount field on the second page as well, and let this PR will work, if there is no page break, or if the membership fee amount is on the second page.

If the membership fee is on the first page, then when that is working, post process of the first page will create the m embership line item from the fee amount fields, and these additions won't hurt either, because the membership fee amount field will not be on the second page of the form, so the javascript won't engage for that, updating any membership line items in the table. 

I think getting the page break stuff back to working is now a prerequisite for this feature, at least if you want to keep it very much the same to how the D7 version worked.

Using this PR makes it work if the membership fee amount is on the second page, OR if the page break feature isn't working.

KarinG:
@jackrabbithanna - going to close this WIP as we can process paid Memberships at this point. That's a good thing!

PR #232, opened 2019-07-02
Replaced wf_crm_get_events by an api only function

kainuk:
Overview
The function wf_crm_get_events is replaced by an version that only uses the CiviCRM Api

Technical Details
The function can be tested with https://github.com/colemanw/webform_civicrm/pull/233 .

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -40,45 +40,56 @@ function wf_crm_state_abbr($input, $ret = 'abbreviation', $country_id = NULL) {
 
 /**
  * Get list of events.
- * FIXME use the api for this.
  *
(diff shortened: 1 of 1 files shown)


colemanw:
I've cherry-picked this into the 7.x-5.x branch and manually tested it. It's also reassuring to know that a semi-automated test is in the works.

IMO #239 should be merged into the 8.x-5.x branch before this one. It will create some small conflicts due to the settings retrieval. Cherry-picking a2bcf02 as a replacement for this PR would solve that.

colemanw:
Ok I've rebased this. Good to merge.

PR #233, opened 2019-07-02
Tests for the Api function replacements

kainuk:
Overview
Function replacements are difficult to test. No before and after situation exists. The idea is that everything stays the same. I created a test script that compares the old function with the replaced one (I just copied the old function in the test).

The tests use existing CiviCRM data, so you have the test data must be created manually. The tests do not change data. The tests are for: 

* https://github.com/colemanw/webform_civicrm/pull/230 
* https://github.com/colemanw/webform_civicrm/pull/232

Comments
I marked this as PR draft. The tests are not to be intended to be a lasting part of the code. After all they only a code change, not a part of the functionality.

### DIFF
--- a/tests/src/Kernel/ApiReplacementTest.php
+++ b/tests/src/Kernel/ApiReplacementTest.php
@@ -2,10 +2,15 @@
 
 namespace Drupal\Tests\webform_civicrm\Kernel;
 
+use CRM_Core_Config;
 use CRM_Core_DAO;
+use CRM_Utils_Date;
 
 class ApiReplacementTest  extends CiviCRMTestBase {
 
+  /**
+   *
+   */
   protected function setUp() {
     parent::setUp();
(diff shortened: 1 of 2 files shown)


KarinG:
I wish there was a Cool label - thanks for sharing! Are there any other direct DB queries left that still need replacing?

colemanw:
This is a good model for how unit tests could be added to this module. So far this module has no unit tests at all and I'd like to see that number grow > 0.
@kainuk would you be willing to start us off with a very simple test. Say, create a PR with a test that simply runs `wf_crm_get_fields()` and verifies that it outputs some data. Then we would be up to 1 unit test in the suite! Maybe on our way toward 2!!

colemanw:
No activity on this PR in a long time. I'd still love to see some tests for this module though.

PR #234, opened 2019-07-13
Add ability to create new case based on existing case

bmango:
Overview
Add ability to create a new case based on an existing case.

I have a use case where users need to create the same type of case every two years using a webform. The case includes about 300 custom fields. To make it more user-friendly, instead of letting them start from scratch, I want to pre-fill the form based on their previous case data.

This is a very simple addition that I think adds some more flexibility to webform_civicrm and will benefit others with similiar use-cases.

Comments
This is my first PR! :)

### DIFF
--- a/includes/wf_crm_admin_help.inc
+++ b/includes/wf_crm_admin_help.inc
@@ -378,7 +378,7 @@ class wf_crm_admin_help {
     print '<p>' .
       t('If a matching case of the chosen type already exists for the client, it will be autofilled and updated.') .
       '</p><p>' .
-      t('Note: a case can also be autofilled by passing "case1", etc. in the url. To create a new case based on an existing case also append "newcase1=yes" etc.') .
+      t('Note: a case can also be autofilled by passing "case1", etc. in the url. To create a new case based on an existing case also append "newcase1=1" etc.') .
       '</p>';
   }
 
--- a/includes/wf_crm_admin_help.inc
+++ b/includes/wf_crm_admin_help.inc
@@ -384,11 +384,11 @@ class wf_crm_admin_help {
 
   public static function duplicate_case_status() {
     print '<p>' .
-      t('This forces a new case to be created based on an existing case.') .
+      t('Choosing this option means a new case will always be created even when an existing case has been selected. If an existing case has been selected the data for this case will NOT be updated.') .
       '</p><p>' .
       t('Useful if you want to pre-fill a form with existing case data, allow the user to make updates and then create a new case with their updates.') .
       '</p><p>' .
-      t('Note: Populate the existing case by selecting an option from the Update Existing Case drop-down or by passing "case1=[caseid]" in the url etc') .
+      t('Note: Populate the existing case by selecting an option from the Update Existing Case drop-down or by passing "case1=[caseid]" in the url etc.') .
       '</p>';
   }
 
--- a/includes/wf_crm_admin_help.inc
+++ b/includes/wf_crm_admin_help.inc
@@ -378,7 +378,17 @@ class wf_crm_admin_help {
     print '<p>' .
       t('If a matching case of the chosen type already exists for the client, it will be autofilled and updated.') .
       '</p><p>' .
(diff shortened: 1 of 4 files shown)


colemanw:
Congrats on your first PR! :tada:

bmango:
Thanks!

includes/wf_crm_webform_preprocess.inc
```diff
@@ -660,6 +660,10 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
             $item = wf_civicrm_api('case', 'getsingle', array('id' => $id));
             if (array_intersect((array)wf_crm_aval($item, 'client_id'), $clients)) {
               $this->ent['case'][$n] = array('id' => $id);
+              // Check if want to create new case based on existing case
+              if(isset($_GET["newcase$n"]) && $_GET["newcase$n"] == 'yes') {
+                $this->ent['case'][$n]['new'] = 'yes';
```
colemanw: Generally booleans passed into the url are either "1" or "0" instead of yes or no.

colemanw:
Seems like a useful feature.
In terms of implementation, doing it as a url arg is nice & simple, but I'd be tempted to make it a backend configuration option instead (as a checkbox on the "Cases" tab of the Webform-CiviCRM page). That would seem more consistent with the way other things are done in this module.

includes/wf_crm_webform_preprocess.inc
bmango: Thanks, I have updated that now so the boolean uses 1 instead of yes.

bmango:
Very glad you like the feature. 
Yes, I think you are right about using backend configuration. It will make the feature more noticeable/understandable to users. I will look into adding that on.

bmango:
I have updated the code to use a checkbox on the Cases tab instead of using a parameter in the url. It was not as difficult as I thought it would be! Just worked it out using trial and error and plenty of dpms :) I think it definitely works better as a configuration option.

includes/wf_crm_admin_help.inc
```diff
@@ -378,7 +378,17 @@ class wf_crm_admin_help {
...
       t('If a matching case of the chosen type already exists for the client, it will be autofilled and updated.') .
       '</p><p>' .
-      t('Note: a case can also be autofilled by passing "case1", etc. in the url. To create a new case based on an existing case also append "newcase1=1" etc.') .
+      t('Note: a case can also be autofilled by passing "case1", etc. in the url.') .
+      '</p>';
+  }
+
+  public static function duplicate_case_status() {
+    print '<p>' .
+      t('This forces a new case to be created based on an existing case.') .
```
colemanw: The language here suggests that an existing case is required, but your code suggests it is not. It looks like the behavior if an existing case is not found would be to just create a new case from scratch (so it is not "forced" to find an existing case in order for the form to function). Is that right?

colemanw:
Nice work figuring out the configuration option!

bmango:
Thanks for your comments. You are absolutely right about the new behaviour, an existing case is not required. I have updated the help message to correct that although the message is getting quite long now.

colemanw:
Long, but helpful and accurate :)

includes/wf_crm_webform_postprocess.inc
```diff
@@ -1260,7 +1260,8 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       if (is_array($data) && !empty($data['case'][1]['client_id'])) {
         $params = $data['case'][1];
         // Set some defaults in create mode
-        if (empty($this->ent['case'][$n]['id'])) {
+        // Create duplicate case based on existing case if 'duplicate_case' checkbox is checked
+        if (empty($this->ent['case'][$n]['id']) || $this->data['case'][$n]['duplicate_case'] == 1) {
```
colemanw: For safety let's just change this line to

```suggestion
        if (empty($this->ent['case'][$n]['id']) || !empty($this->data['case'][$n]['duplicate_case'])) {
```
And then I think we're good to merge :)

bmango:
Many thanks for reviewing the updates! I have committed your suggestion.

PR #236, opened 2019-07-15
Fix a bug when new multi-value fields are added to the webform.

aydun:
Overview
See https://www.drupal.org/project/webform_civicrm/issues/3065727

Steps to reproduce:
- create multi-value set of custom fields on Contacts.  Add some fields
- add data to a contact
- create a webform, in CiviCRM tab add custom fields just created
- view the webform with ?cid=  for the contact you added data to

Before
The existing values are not loaded.
Go back to the webform and resave the config.
Then reload the webform - should show values.

After
The values are loaded as expected

Technical Details
Previously, the 'createmode' keys are wrongly converted to arrays since the
code path for `!isset($enabled[$key])` is followed. This
results in multi-value fields not being loaded by the webform.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1656,6 +1656,11 @@ class wf_crm_admin_form {
         if (substr($key, 0, 7) == 'civicrm') {
           ++$i;
           $field = wf_crm_get_field($key);
+          if (substr($key, -11) === '_createmode') {
+            // Update webform's settings with 'Create mode' value for custom group.
+            $this->settings['data']['config']['create_mode'][$key] = $val;
+          }
+
           if (!isset($enabled[$key])) {
             $val = (array) $val;
             if (in_array('create_civicrm_webform_element', $val, TRUE) || (!empty($val[0]) && $field['type'] == 'hidden')) {
@@ -1686,10 +1691,6 @@ class wf_crm_admin_form {
             webform_component_update($component);
           }
 
-          if (substr($key, -11) === '_createmode') {
-            // Update webform's settings with 'Create mode' value for custom group.
-            $this->settings['data']['config']['create_mode'][$key] = $val;
-          }
         }
         // add empty fieldsets for custom civicrm sets with no fields, if "add dynamically" is checked
         elseif (strpos($key, 'settings_dynamic_custom') && $val == 1) {


KarinG:
Hi Aiden - I’ll have a look at this next week. We’ve been using custom fields w/ multiple values / create for several clients recently - all create mode;

aydun:
Thanks Karin.  For this, I'm using create/edit mode.  The problematic code path when adding a new field on the admin form is: https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_admin_form.inc#L1659-L1660 followed by https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_admin_form.inc#L1691 so the stored value is an array.

The cast at https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_admin_form.inc#L1691 means the array is interpreted as 1 - ie create mode.

KarinG:
Current behavior is that the webform does not load any values and I actually understood that was by design (as it would not be clear which existing data from a multi data set it should load). And there are use cases where one does not want to load any. How do you think we should handle that?

aydun:
Hmm - maybe we're talking about different things?  It is loading fine once the flag is fixed.  Isn't that the point of the createmode flag?  "Create/Edit" loads the existing values so they can be edited, and "Create" doesn't load, just adds new values.  https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_webform_preprocess.inc#L481-L487

As regards which values to load - that's like other places where there are multiple values/records to load: create multiple similar fields and webform populates them with different values in some unspecified order.

KarinG:
@aydun - I can not reproduce this.

**Form configured to expect ?cid1=** so nothing loads without that argument present:

image

**with ?cid1=4 in the URL** -> all fields load:

image

**Qoli -> qo20 contains multiple values** -> it appears to load the first entry in the table for it:

image

**Switching to Create Only Mode -> with ?cid1=4 in the URL** indeed shows a blank field for qo20 -> filling in 4 -> correctly produces:

image

**Deleting the first record from the table in CiviCRM** - leaving 8 and 4 for score & resetting the webform to Create/Edit mode for the multi-value fields -> now (as expected) loads:

image

aydun:
@KarinG - I've pulled the latest 7.x-5.x and still seeing the same behaviour.

Note that resaving the form corrects the problem - it only shows directly after the fields have been added to the form and saved the first time.  Going back to the Civi tab and resaving with no changes corrects it.

KarinG:
Ah - ok - trying again. Ah yes seeing it. Ok merging. Thanks @aydun

PR #237, opened 2019-07-25
Fixed roles retrieval in webform

Yevalnud:
Overview
Case roles in webform weren't being retrieved correctly from Civi when the label and the name were different.

Before
Under the case tab in webform, the correct case roles were not being retrieved. 

After
Correct case roles now being retrieved under the case tab in webform.

Technical Details
Used label and name to check against relationship type name stored in case type definition.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -941,7 +941,7 @@ function wf_crm_get_fields($var = 'fields') {
       foreach (wf_crm_apivalues('case_type', 'get') as $case_type) {
         foreach ($case_type['definition']['caseRoles'] as $role) {
           foreach (wf_crm_get_relationship_types() as $rel_type) {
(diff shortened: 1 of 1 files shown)


pradpnayak:
Similar [issue](https://civicrm.stackexchange.com/questions/31493/assigning-case-roles-via-webform) was reported in SE today. Thanks @Yevalnud for fixing this.

adlapidem:
Patch fixes the problem in my D7 4.7.x site in which custom Case Roles were not loaded for assignment in Webform on the CiviCRM config tab. Good timing.

colemanw:
@demeritcowboy do you support merging this?

demeritcowboy:
As an improvement yes. Wondering if pulling it into a separate isRoleEqual() or something function in utils.inc would make it easier to change later if the pattern gets reused? Thanks for asking.

PR #239, opened 2019-08-01
Use api to retrieve civi settings

colemanw:
This is a cherry-pick of 3da7c660406217a23d759dcf2772b7221aafec11 from the 7.x-5.x branch.

The goal is to use the api more and reduce usage of Civi internals like `CRM_Core_BAO_Setting` and `CRM_Core_Config`.


KarinG:
Awesome! :-) I'll give it a test when I get back at my desk.

PR #242, opened 2019-08-02
Fix add element on webform

jitendrapurohit:
Overview
Fix add element button on webform.

Before
Clicking on the Add element doesn't work.

image

After
Clicking on the Add element works as expected.

Comments
ping @KarinG

### DIFF
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -9,6 +9,8 @@ use Drupal\webform\Plugin\WebformElementBase;
 use Drupal\webform\WebformSubmissionInterface;
 use Drupal\webform_civicrm\Utils;
 
+include_once __DIR__ . '/../../../includes/utils.inc';
+
 /**
  * Provides a 'civicrm_options' element.
  *
@@ -133,6 +135,7 @@ class CivicrmOptions extends WebformElementBase {
    * {@inheritdoc}
    */
   public function prepare(array &$element, WebformSubmissionInterface $webform_submission = NULL) {
+    \Drupal::service('civicrm')->initialize();
     $as_list = !empty($element['#extra']['aslist']);
     $is_multiple = !empty($element['#extra']['multiple']);
     $use_live_options = !empty($element['#civicrm_live_options']);


KarinG:
Thanks for moving that into `CivicrmOptions::prepare` !

KarinG:
I have at times run into this - but I've not run into this for some time now. I don't know why that is. 

Right now I can hit Add Element -> on here and add Buttons, e.g. 
image

image

Is this still an issue @jitendrapurohit ? If so please give me exact steps to reproduce and I'll prioritize it.

mglaman:
Per my comments in #222, I see this as something which should still exist when a CiviCRM Options element is on the form

KarinG:
Odd - even when adding a CiviCRM Options element (a select to the form) -> 

image

Is this one of those scenarios where better in than not b/c there will some scenarios where this will resurface?

jitendrapurohit:
I was able to reproduce this on a clean updated `8.x-5.x` branch of this repo(without any patch applied).

I simply added a NEW webform and tried to click on the "Add Element" button without adding any civi fields from the settings tab.

image

KarinG:
Ok trying to reproduce this now installing a brand new webform and webform_civicrm for a brand new client (Alan has them on CiviCRM 5.15.1):

root@964fcf326132:/var/www/drupal# composer require 'drupal/webform:^5.4'

root@964fcf326132:/var/www/drupal/web/modules/contrib# git clone https://github.com/colemanw/webform_civicrm.git --branch 8.x-5.x

I am able to create a webform and I can hit Add Element and add a checkbox without having visited the CiviCRM tab (CiviCRM processing is still disabled);

image

Happy to leave this issue open (I do believe you!)  - but I know you understand I can't merge until I can reproduce.

jitendrapurohit:
The only difference I see between our screenshots is the way the webform is rendered. My admin theme was set to "Seven" and switched to Barthik to see if it makes any difference, but it still replicates for me :). Happy to leave this open to let others try and comment 👍

jitendrapurohit:
I did a pull today and got some changes from the time this PR was created. The Add element is working which I think is auto-fixed by latest update. I'm closing this and will re-open if I see it again. 

Thanks :)

KarinG:
Happy dance!

PR #243, opened 2019-08-02
Remove unused function wf_crm_get_activity_types

colemanw:
This ports commit 5a4e48a from the 7.x branch to remove a function that's no longer used.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -124,44 +124,6 @@ function wf_crm_get_surveys($act = array()) {
 }
 
 /**
- * Get list of activity types
- * @param string $case_type_name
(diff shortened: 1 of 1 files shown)


KarinG:
What was this function replaced with?

colemanw:
This function hasn't been touched since 2014. Somewhere along the line it was made obsolete as more fields were standardized to use api.getoptions to fetch option lists. Reading the current code it looks like that generic handler is used as the `activity_type_id` field isn't even declared on the admin form (it's just loaded from metadata like every other field).

KarinG:
Ah got it. Thank you!

PR #244, opened 2019-08-02
Remove legacy stuff

colemanw:
This is a port of c21b844f05da60e293da8feb5bb6793071f49d63 from the 7.x branch, with the one difference that it updates the minimum WEBFORM_VERSION to 5.


includes/wf_crm_webform_preprocess.inc
```diff
@@ -784,17 +776,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
       }
     }
     if ($items) {
-      // Todo: Remove when we drop support for 4.4 - https://www.drupal.org/node/2832331
-      if (version_compare($this->civicrm_version, '4.6.0', '<')) {
```
KarinG, line 788: Remove when we drop support for 4.4 - about time :-) Thank you for the summer clean up and for also porting it to our D8WFC.

PR #246, opened 2019-08-03
Switch to api for loading event/participant info

colemanw:
Overview
Removes a direct query and internal function call in favor of the api.

Before
BAO & SQL :(

After
API :)

Technical Details
Renames some of the array keys we were passing around within the webform to match the standard keys from the api.


colemanw:
Merging based on D8 review of the same changes.

PR #247, opened 2019-08-03
Switch to api for loading event/participant info

colemanw:
Overview
Removes a direct query and internal function call in favor of the api.

Before
BAO & SQL :(

After
API :)

Technical Details
Renames some of the array keys we were passing around within the webform to match the standard keys from the api.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -416,14 +416,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         continue;
       }
       $participants = current(wf_crm_aval($this->data['contact'], "$c:contact", array()));
-      $firstName = CRM_Utils_Array::value('first_name', $participants);
-      $lastName = CRM_Utils_Array::value('last_name', $participants);
 
-      $participantName = "{$firstName} {$lastName}";
+      $participantName = ($participants['first_name'] ?? '') . ' ' . ($participants['last_name'] ?? '');
       if (!trim($participantName)) {
-        $participantName = CRM_Utils_Array::value('webform_label', $participants);
+        $participantName = $participants['webform_label'] ?? NULL;
         if (!empty($this->existing_contacts[$c])) {
-          $participantName = CRM_Contact_BAO_Contact::displayName($this->existing_contacts[$c]);
+          $participantName = wf_crm_apivalues('contact', 'get', $this->existing_contacts[$c], 'display_name')[0];
         }
       }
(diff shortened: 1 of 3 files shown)


includes/wf_crm_webform_postprocess.inc
```diff
@@ -487,12 +486,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         if ($event['ended']) {
           form_set_error($eid, t('Sorry, you can no longer register for %event.', array('%event' => $event['title'])));
```
kainuk: `form_set_error` is removed from Drupal8 - you can get this working by using  `$this->form_state->setErrorByName` (The form_set_error was already there before the port)

includes/wf_crm_webform_postprocess.inc
```diff
@@ -487,12 +486,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
         }
-        elseif ($event['max_participants'] && $event['count'] > $event['remaining']) {
-          if (!empty($event['full'])) {
-            form_set_error($eid, '<em>' . $event['title'] . '</em>: ' . $event['full_message']);
```
kainuk, line 492: again `form_set_error`. The drupal 8 alternative does not support html so the emphasis tags   give the following effect. 
Selection_003

includes/wf_crm_webform_postprocess.inc
```diff
@@ -487,12 +486,12 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+        elseif (!empty($event['max_participants']) && $event['count'] > $event['available_places']) {
+          if (!empty($event['is_full'])) {
+            form_set_error($eid, '<em>' . $event['title'] . '</em>: ' . $event['event_full_text']);
           }
           else {
-            form_set_error($eid, format_plural($event['remaining'],
```
kainuk, line 495: And this is the last `form_set_error`.

kainuk:
Just tested it with the replaced `form_set_error`. Event full works.

kainuk:
Line items is still a bit a mystery for me, but with this PR will be a improvement.

colemanw:
@kainuk so is this ready to merge in your opinion?

kainuk:
Hi @colemanw I created a review, but I forgat to submit it. I have added a number of Drupal 8 changes. I added them as a diff file to this comment.
[PR247diff.txt](https://github.com/colemanw/webform_civicrm/files/3559729/PR247diff.txt)

colemanw:
@kainuk I've manually applied your patch but in the future you can just submit a PR against a PR.

KarinG:
Thank you both!

PR #248, opened 2019-08-03
Use attachment api instead of internal functions

colemanw:
Overview
Use the attachment api to future-proof file handling.

Before
More code, less stable.

After
Less code, more stable.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -942,19 +942,14 @@ abstract class wf_crm_webform_base {
         'icon' => $val,
       );
     }
-    $file = wf_crm_apivalues('file', 'get', $val);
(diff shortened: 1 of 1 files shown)


colemanw:
To test this PR, add a custom field of type "file" for an existing contact. It should save files and then show the existing file when reloading the webform. Behavior should be unchanged before/after this PR.

KarinG:
ok - tested: I can upload a File via the Webform -> and I can view it in CiviCRM! Merging this. 

PS - Going back to the webform -> there is no indication/flag that I've already just uploaded a File -> so I can (accidentally) overwrite it.

KarinG:
And another one done! How many are left?

colemanw:
Nice. #251 should be fairly simple to review...

PR #249, opened 2019-08-03
Use attachment api instead of internal functions

colemanw:
Overview
Use the attachment api to future-proof file handling.

Before
More code, less stable.

After
Less code, more stable.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -930,18 +930,14 @@ abstract class wf_crm_webform_base {
         'icon' => $val,
       );
     }
-    $file = wf_crm_apivalues('file', 'get', $val);
-    $entity_id = '';
-    if ($entity && $n && (strpos($fieldName, 'custom_') === 0 || strpos($fieldName, 'file_') === 0)) {
-      $entity_id = $this->ent[$entity][$n]['id'];
-    }
+    $file = wf_crm_apivalues('Attachment', 'get', $val);
     if (!empty($file[$val])) {
-      return array(
+      return [
         'data_type' => 'File',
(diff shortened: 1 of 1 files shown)


includes/wf_crm_webform_base.inc
```diff
@@ -930,18 +930,14 @@ abstract class wf_crm_webform_base {
...
-      $entity_id = $this->ent[$entity][$n]['id'];
-    }
+    $file = wf_crm_apivalues('Attachment', 'get', $val);
     if (!empty($file[$val])) {
-      return array(
+      return [
         'data_type' => 'File',
-        'name' => CRM_Utils_File::cleanFileName($file[$val]['uri']),
-        'file_url'=> CRM_Utils_System::url('civicrm/file', "reset=1&id={$val}&eid={$entity_id}", TRUE),
-        'icon' => file_icon_url((object) array('filemime' => $file[$val]['mime_type'])),
```
kainuk, line 943: The current code does not work with Drupal8, because it does not have the function file_icon_url anymore. I cannot find a simple replacement function. I did find, however a function file_icon_class, that can be used to assign a html class, so that the icon location can be picked up from the css file.

colemanw: That sounds reasonable. Is there already a css rule supplied by Drupal's stylesheet or do we need to add it to `webform_civicrm_forms.css`? Note there are already a few rules in that stylesheet pertaining to file icons so those can be modified to suit D8.

colemanw: @kainuk could you do a patch or a PR for this?

kainuk: Hi @colemanw, the fileInfo does not work at all in Drupal 8. I will create a seperate PR for that, so I can test this PR.

PR #251, opened 2019-08-04
Use api to retrieve relationship data

colemanw:
Overview
Another sql -> api swapout

Before
SQL :(

After
API :)

Technical Details
Since the api automatically fetches custom data, I was able to remove some extra code that did that separately.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -887,55 +887,56 @@ function wf_crm_update_existing_component(&$component, $enabled, $data) {
  * @return array
  */
 function wf_crm_find_relations($cid, $types = array(), $current = TRUE) {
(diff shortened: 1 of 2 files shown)


colemanw:
To test this PR, expose some relationships to the form, including custom relationship fields. They should load and save correctly before/after the patch.

KarinG:
@kainuk - are you able to port this one to D8 as well? Then I’ll review both in one go.

colemanw:
@KarinG already done: #252

KarinG:
Doh - ok will test this :-)

KarinG:
Testing -> https://github.com/colemanw/webform_civicrm/pull/252

PR #252, opened 2019-08-04
Use api to retrieve relationship data

colemanw:
Port of #251


KarinG:
Ok - confirmed this is working!

image

image

PR #253, opened 2019-08-06
Switch to api for fetching tags

colemanw:
Overview
Another BAO -> API conversion.

Before
Used BAO function to fetch tags.

After
Wrote our own function to get tags via api.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -347,7 +347,7 @@ function _webform_edit_civicrm_contact($component) {
     '#type' => 'select',
     '#multiple' => TRUE,
     '#title' => t('Tags'),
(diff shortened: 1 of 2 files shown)


colemanw:
@KarinG the simple way to test this PR is to create a bunch of nested tags and tagsets, then load the "CiviCRM" tab of a webform and inspect the tags and tagsets therein. It should look exactly the same before/after:

image

KarinG:
Ok done! Tested:
image

KarinG:
**PS** - Note: https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/utils.inc#L157 is incompatible with PHP 5.6

colemanw:
@KarinG that's ok, CivICRM now requires PHP 7.

PR #255, opened 2019-08-10
Empty submission for Datetime custom field resulting in default (now)…

deepak-srivastava:
… getting saved in Civi

Overview
If there are civi datetime custom fields included in webform, empty submission of fields results in storing submission date (today's) in civi.
Note this only happens if there is time section included. If it's just date field, it works fine.

Before
image

image

Aug 10 17:01:15  [info] $$data before saveCustomDate() = Array
(
.............
    [cg6] => Array
        (
            [1] => Array
                (
                    [custom_11] => 20190813011500
                    [custom_12] => 20190810
                    [custom_13] => 20190810
                )
        )
)

After
image

Aug 10 17:03:49  [info] $$data before saveCustomData() = Array
(
.............
    [cg6] => Array
        (
            [1] => Array
                (
                    [custom_11] => 20190813011500
                    [custom_12] => 
                    [custom_13] => 
                `)`
        )
)

Technical Details
The code which fudges date and time together, populates date with default (Today) before merging with time, causing the problem. 
The date section on the other hand doesn't use default, therefore any date fields without time-part works fine.
There might be some original intention which i'm unaware of. Even if there is, it's not consistent with how just date fields work.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2357,8 +2357,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         // Fudge together date and time fields
         if ($field['type'] === 'time' && substr($name, -8) === 'timepart') {
           $name = str_replace('_timepart', '', $name);
-          // Add date (default to today)
-          $date = wf_crm_aval($this->data, "$ent:$c:$table:$n:$name", date('Ymd'));
+          $date = wf_crm_aval($this->data, "$ent:$c:$table:$n:$name", '');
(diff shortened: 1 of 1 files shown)


KarinG:
Ah - thanks for catching this inconsistency. If one really wanted a default - one could and should configure that on the webform element itself (set default to 'now').  

@colemanw - can you think of any reason why we should not merge this?

KarinG:
Thanks Deepak!

kainuk:
Thanks @deepak-srivastava, just installed your PR on a customer site that has an event with many date fields.

PR #256, opened 2019-08-12
Fix file upload jQuery selector

colemanw:
Overview
Fixes the jQuery selector used to target file fields. Now works better across themes.

Before
Files not always shown correctly.

After
Files shown correctly.

Technical Details
Using the *= to match class works better as class attributes can be in any order


KarinG:
Awesome!! Now works on bootstrap themes; live on client site so merging this.

PR #257, opened 2019-08-12
Fix file upload jQuery selector

colemanw:
Using the *= to match class works better as class attributes can be in any order


colemanw:
@KarinG already merged #256 and this is exactly the same.

PR #259, opened 2019-08-17
Contact allow url autofill

mglaman:
Webform removes default values from the element, assuming to keep YAML small. Using `getElementProperty` on the element plugin we can get proper values. This has enabled prefilling properly.


KarinG:
Awesome - this is working (with and without flexboxes). Documenting my testing w/ flexbox scenario:

image

image

KarinG:
Thanks Matt!

PR #245, opened 2019-08-02
Fetch fields with api instead of sql query

colemanw:
Overview
----------------------------------------
Replaces a whopping sql query with an api call.

Before
----------------------------------------
SQL :(

After
----------------------------------------
API :)

PR #260, opened 2019-08-20
Refactor getCustomData function to use api instead of BAO

colemanw:
Another api refactor. I've tested this by stepping through the code in multiple scenarios and the new function output is identical to the old.


KarinG:
I put custom fields of various types on the form:
integer; varchar(500) for encrypted field); date field, file, radio buttons, custom data containing multiple records

and all is well - with custom fields (form, submit, load/fill)

KarinG:
PS - and helped me identify two small bugs in my deur_encrypt module :-)

KarinG:
Over view of Custom Fields part of testing https://github.com/colemanw/webform_civicrm/pull/260 https://github.com/colemanw/webform_civicrm/pull/245

image

PR #261, opened 2019-08-22
Begin support for autocomplete

mglaman:
Screen Shot 2019-08-30 at 10 54 18 PM

Autocomplete of contacts is now available. 

Need to test the CID which is set for the element value. (The cid is set to the element's value)

~~Fields do not populate. (the values do come over the AJAX request.)~~ Selecting a contact and removing it will clear contact fields.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -623,7 +623,7 @@ function wf_crm_contact_access($component, $filters, $cid) {
   // Create new contact doesn't require lookup
   $cid = (string) $cid;
   list(, $c, ) = explode('_', $component['#form_key'], 3);
-  if ($cid && !empty($component['#none_prompt']) && !empty($component['#allow_create']) && strpos($cid, '-') === 0) {
+  if (!empty($component['#none_prompt']) && !empty($component['#allow_create']) && $cid && strpos($cid, '-') === 0) {
     return Html::escape($component['#none_prompt']);
   }
   if (!is_numeric($cid)) {
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -623,7 +623,7 @@ function wf_crm_contact_access($component, $filters, $cid) {
   // Create new contact doesn't require lookup
   $cid = (string) $cid;
   list(, $c, ) = explode('_', $component['#form_key'], 3);
-  if (!empty($component['#none_prompt']) && !empty($component['#allow_create']) && $cid && strpos($cid, '-') === 0) {
+  if ($cid && !empty($component['#none_prompt']) && !empty($component['#allow_create']) && strpos($cid, '-') === 0) {
     return Html::escape($component['#none_prompt']);
   }
   if (!is_numeric($cid)) {
@@ -809,9 +809,13 @@ function wf_crm_find_relations($cid, $types = array(), $current = TRUE) {
  *   Api params
  */
 function wf_crm_search_filters($node, array $component) {
+  /** @var \Drupal\webform\Plugin\WebformElementManagerInterface $element_manager */
+  $element_manager = \Drupal::service('plugin.manager.webform.element');
+  $contact_element = $element_manager->getElementInstance($component);
   $params = ['is_deleted' => 0];
-  $params['contact_type'] = $component['contact_type'];
-  foreach ($component['filters'] as $key => $val) {
+  $params['contact_type'] = $contact_element->getElementProperty($component, 'contact_type');
+  $filters = $contact_element->getElementProperty($component, 'filters');
+  foreach ($filters as $key => $val) {
     if (is_array($val)) {
       $val = array_filter($val);
     }
--- a/includes/wf_crm_webform_ajax.inc
+++ b/includes/wf_crm_webform_ajax.inc
@@ -29,7 +29,7 @@ class wf_crm_webform_ajax extends wf_crm_webform_base {
    * @return mixed
    */
   private function getParameter($name){
-    return $this->requestStack->getCurrentRequest()->get($name);
+    return $this->requestStack->getCurrentRequest()->query->get($name);
   }
 
   /**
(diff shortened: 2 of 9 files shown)


includes/contact_component.inc
```diff
@@ -623,7 +623,7 @@ function wf_crm_contact_access($component, $filters, $cid) {
   // Create new contact doesn't require lookup
   $cid = (string) $cid;
   list(, $c, ) = explode('_', $component['#form_key'], 3);
-  if (!empty($component['#none_prompt']) && !empty($component['#allow_create']) && $cid && strpos($cid, '-') === 0) {
+  if ($cid && !empty($component['#none_prompt']) && !empty($component['#allow_create']) && strpos($cid, '-') === 0) {
```
mglaman: Undo this flip to make the diff easier.

includes/contact_component.inc
```diff
@@ -623,7 +623,7 @@ function wf_crm_contact_access($component, $filters, $cid) {
...
     return Html::escape($component['#none_prompt']);
   }
   if (!is_numeric($cid)) {
```
mglaman: 🤔 it's never numeric when searching, what went wrong here in previous code changes

includes/contact_component.inc
```diff
@@ -809,9 +809,13 @@ function wf_crm_find_relations($cid, $types = array(), $current = TRUE) {
  *   Api params
  */
 function wf_crm_search_filters($node, array $component) {
+  /** @var \Drupal\webform\Plugin\WebformElementManagerInterface $element_manager */
+  $element_manager = \Drupal::service('plugin.manager.webform.element');
+  $contact_element = $element_manager->getElementInstance($component);
```
mglaman, line 814: Need to make a helper for this, since we have so many properties

includes/wf_crm_webform_ajax.inc
```diff
@@ -180,7 +184,7 @@ class wf_crm_webform_ajax extends wf_crm_webform_base {
       return TRUE;
     };
 
-    if (!empty($webform->getElementDecoded($fid)['#private'])) {
+    if (!empty($webform->getElement($fid)['#private'])) {
```
mglaman: This is possibly a false positive because default fields are never present. Review.

webform_civicrm.libraries.yml
```diff
@@ -22,10 +22,13 @@ payment:
...
-# _webform_render_civicrm_contact
-# only used for autocomplete https://github.com/colemanw/webform_civicrm/blob/8f285b94c4ba444fc894b24f2d23a7c52c92731b/includes/contact_component.inc#L384
-token-input:
+civicrm_contact:
   css:
     component:
       css/token-input.css: {}
+  js:
+    # @todo is this a bad assumption? is there a way to load via CiviCRM helpers?
+    /libraries/civicrm/packages/jquery/plugins/jquery.tokeninput.js: {}
```
mglaman, line 31: Is there a way to load this via CiviCRM helpers? Because we can no longer inline JavaScript or dynamically load files.

KarinG: @colemanw - in house CiviCRM JavaScript hero :-), do you know the answer to Matt's question?

colemanw: Does this line work? If so then the only change I'd suggest is to change the hardcoded `/libraries/civicrm` to the real path of the civicrm module.

includes/wf_crm_webform_ajax.inc
mglaman: Need to get the element plugin for this.
```
  /** @var \Drupal\webform\Plugin\WebformElementManagerInterface $element_manager */
  $element_manager = \Drupal::service('plugin.manager.webform.element');
  $contact_element = $element_manager->getElementInstance($component);
```

webform_civicrm.libraries.yml
mglaman: Using `hook_libraries_build` or `hook_libraries_alter` we could set it based on the CiviCRM resources setting. We would just need to grab the `Resource URL` setting and dynamically adjust when libraries are built.

The default was taken based off of https://gitlab.com/roundearth/drupal-civicrm-project

mglaman: Is this it?

```
$userFrameworkResourceURL = CRM_Core_Config::singleton()->userFrameworkResourceURL;
```

mglaman: Found it out :)

```
/**
 * Implements hook_library_info_alter().
 */
function webform_civicrm_library_info_alter(array &$libraries, $extension) {
  if ($extension === 'webform_civicrm') {
    $user_framework_resource_url = CRM_Core_Config::singleton()->resourceBase;
    unset($libraries['civicrm_contact']['js']['/libraries/civicrm/packages/jquery/plugins/jquery.tokeninput.js']);
    $libraries['civicrm_contact']['js'][$user_framework_resource_url . 'packages/jquery/plugins/jquery.tokeninput.js'] = [
      'preprocess' => FALSE,
      'minified' => FALSE
    ];
  }
}

```

Kept the default and added the dynamic definition in the hook. I think this comment can resolved.

KarinG:
Awesome! Roger that:

image

Merging!

colemanw:
Beautiful

kainuk:
Just tried it, works very well.

KarinG:
Did first name and last name fill for you?

kainuk:
I tested it on small form with first and last name and a custom field with file properties. It all works.

KarinG:
I just cleared caches again - and the autocomplete search is working -> but the First Name and Last Name are not filled in. 

Example: 
image

This is my YAML:
```
  '#type': civicrm_contact
  '#title': 'Existing Contact'
  '#widget': autocomplete
  '#search_prompt': 'Find Contact'
  '#none_prompt': '+ Create new +'
  '#results_display':
    display_name: display_name
  '#allow_create': 1
  '#contact_type': individual
  '#extra': {  }
civicrm_1_contact_1_contact_first_name:
  '#type': textfield
  '#contact_type': individual
  '#form_key': civicrm_1_contact_1_contact_first_name
  '#extra':
    width: 20
  '#title': 'First Name'
civicrm_1_contact_1_contact_last_name:
  '#type': textfield
  '#contact_type': individual
  '#form_key': civicrm_1_contact_1_contact_last_name
  '#extra':
    width: 20
  '#title': 'Last Name'
```

KarinG:
@kainuk - you're not just seeing first_name and last_name b/c it's you and the element config still has a current user in there somewhere?

KarinG:
PS - I double checked my console - no error messages

kainuk:
@KarinG I did the following test.
- database with three contacts
- form with three fields. First Name, Last Name, Gender (just by the selection in the CiviCRM settings)
- The autocomplete property on the existing field is set (but after saving the element, I also press save elements on the form)
- Test on the View tab of webforms (so I am logged in).

Start Test
- Form is prepopulated with my first, name, last name and gender.
- I remove my name by pressing the x on the end of the existing bo.
- Now I can select one of the other contacts - and the fields are all populated.

When I log out, the autocomplete does not work. I suppose that has to do with the authorization

KarinG:
I can't get it to fill the fields. If leave the default to existing user it pre-populates the form - but I can't get the First Name and Last Name fields to clear/change to another Contact I search for in the autocomplete. I wonder what's different.

KarinG:
Figured it out! Confirming it’s all working now.

jitendrapurohit:
Great work :)

It looks like it does not work on webform having multiple contacts? i.e, If the type of the second `Existing Contact` element is updated to autocomplete, it is not displayed on the form as the first element.

This seems to be because the [relationship part](https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/src/Plugin/WebformElement/CivicrmContact.php#L387-L400) is not defined for the contact field.

PR #263, opened 2019-08-26
Update wf_crm_admin_form.inc to fix a wrong link generation in a help message

uniquename:
Overview
Fixed a wrong link generation in a help message

Before
Link was not created properly leading to a non working a tag.

After
Link is create in a corrext way.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1528,7 +1528,7 @@ class wf_crm_admin_form {
         ];
         $options = array('attributes' => array('target' => '_blank'));
         $url = Url::fromUri('https://docs.civicrm.org/sysadmin/en/latest/integration/drupal/webform/#cloning-a-contact', $options);
-        $link = Link::fromTextAndUrl('more info', $url)->toString();
+        $link = Link::fromTextAndUrl(t('more info'), $url)->toString();
         \Drupal::messenger()->addStatus(t('Tip: Consider using the clone feature to add multiple similar contacts.  (@link)', ['@link' => $link]));
       }
     }
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -7,6 +7,7 @@ use Drupal\Core\Url;
 use Drupal\webform\Utility\WebformArrayHelper;
 use Drupal\webform\WebformInterface;
 use Drupal\webform_civicrm\Utils;
+use Drupal\Core\Link;
 
 /**
  * @file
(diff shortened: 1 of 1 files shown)


includes/wf_crm_admin_form.inc
```diff
@@ -1525,8 +1526,10 @@ class wf_crm_admin_form {
           'civicrm_' . $c . '_contact_1_contact_first_name' => 'create_civicrm_webform_element',
           'civicrm_' . $c . '_contact_1_contact_last_name' => 'create_civicrm_webform_element',
         ];
-        $link = [':link' => 'href="https://docs.civicrm.org/sysadmin/en/latest/integration/drupal/webform/#cloning-a-contact" target="_blank"'];
-        \Drupal::messenger()->addStatus(t('Tip: Consider using the clone feature to add multiple similar contacts. (<a :link>more info</a>)', $link));
+        $options = array('attributes' => array('target' => '_blank'));
+        $url = Url::fromUri('https://docs.civicrm.org/sysadmin/en/latest/integration/drupal/webform/#cloning-a-contact', $options);
+        $link = Link::fromTextAndUrl('more info', $url)->toString();
```
colemanw: Should 'more info' be wrapped in `t()` here?

uniquename: Yes, it should.

includes/wf_crm_admin_form.inc
```diff
@@ -1525,8 +1526,10 @@ class wf_crm_admin_form {
           'civicrm_' . $c . '_contact_1_contact_first_name' => 'create_civicrm_webform_element',
           'civicrm_' . $c . '_contact_1_contact_last_name' => 'create_civicrm_webform_element',
         ];
-        $link = [':link' => 'href="https://docs.civicrm.org/sysadmin/en/latest/integration/drupal/webform/#cloning-a-contact" target="_blank"'];
-        \Drupal::messenger()->addStatus(t('Tip: Consider using the clone feature to add multiple similar contacts. (<a :link>more info</a>)', $link));
+        $options = array('attributes' => array('target' => '_blank'));
+        $url = Url::fromUri('https://docs.civicrm.org/sysadmin/en/latest/integration/drupal/webform/#cloning-a-contact', $options);
```
mglaman, line 1530: URL generation is not required, this HTML could go into a FormattableMarkup string (like sprintf style)

mglaman (commented):
This could be simplified to just the HTML string in the formattable markup, there is no need to use URL generation since it is a remote URL.

includes/wf_crm_admin_form.inc
uniquename: well, just found it not working at all, so i wanted to give a quick fix. don't have more time...

mglaman:
:( I didn't mean to discourage, @uniquename I just was providing some feedback so we could provide a fix and clean up the existing code.

PR #264, opened 2019-09-01
FileInfo does not work 

kainuk:
Overview
Custom Fields in CiviCRM can contain files and civicrm_webforms makes it possible to upload these files. Unfortunately when you try to edit the webforms results, and error is shown.

To reproduce.
1. Create a custom file field on a contact.
2. Create a webform with this field.
3. Use the form to upload a file
4. Go to webform results.
5. Select the edit button: Error `Call to undefined function file_icon_url() in wf_crm_webform_base->getFileInfo()`

After the PR
File details are shown (Icon and filename link). See picture
Selection_004

Technical Details
Drupal 8 does not work with file icons anymore used the function `file_icon_class`

Drupal 8 does not allow attachment of php generated jQuery snippets to a component. So Javascript with paramaters that are only known at runtime must done in different way. Solution.
* put the parameters in the drupal settings.
* add the JQuery call to the drupal behaviors

Known Issues
------------------------------------------
Clicking on the filelink does not work. The hope is that the following PR will improve this: https://github.com/colemanw/webform_civicrm/pull/249

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -912,7 +912,7 @@ abstract class wf_crm_webform_base {
         'data_type' => 'File',
         'name' => CRM_Utils_File::cleanFileName($file[$val]['uri']),
         'file_url'=> CRM_Utils_System::url('civicrm/file', "reset=1&id={$val}&eid={$entity_id}", TRUE),
-        'icon' => file_icon_url((object) array('filemime' => $file[$val]['mime_type'])),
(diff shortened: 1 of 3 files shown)


KarinG:
@kainuk - goedenmorgen! Ik ga hier vandaag naar kijken

KarinG:
Fantastic @kainuk - and as always bedankt for your details description/steps. I was able to reproduce it and with your PR -> I can Edit/Save:
 
image

PR #265, opened 2019-09-03
Use attachment api instead of internal functions + including file_icon_class

kainuk:
The same as https://github.com/colemanw/webform_civicrm/pull/249 but now file_icon_class instead of file_icon_url. So now it runs on Drupal 8

This also fixes the known issue of https://github.com/colemanw/webform_civicrm/pull/264. Now it is possible to download the file shown by the fileInfo. Unfortunately the attachment api generates a absolute url instead a relative one. This causes Chrome to open a seperate window.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -930,18 +930,14 @@ abstract class wf_crm_webform_base {
         'icon' => $val,
       );
     }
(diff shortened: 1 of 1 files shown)


colemanw:
Code looks good to me :+1: 
@KarinG want to test it out on D8?
And if that works then #248 could also be merged for D7.

KarinG:
Thank you! Works great - I can upload/download files :-)

PR #266, opened 2019-09-06
Fix payment_processor_id in test mode so CiviCRM always uses the correct processor

mattwire:
Overview
webform_civicrm uses the live payment_processor_id internally and calls various CiviCRM functions based on that ID.  It attempts before final payment to select the test ID but this is not always reliable and causes problems with various payment processors (eg. stripe) which cannot be used in test mode without custom code to handle webform_civicrm submissions.

See https://www.drupal.org/project/webform_civicrm/issues/2420809 and https://www.drupal.org/project/webform_civicrm/issues/2756937

Before
Wrong payment_processor_id used in test mode (the live one is used)

After
The correct payment_processor_id is used throughout.

Technical Details
I didn't convert the settings because this would require updating existing configurations.  So instead I update the payment_processor_id immediately after the settings have been loaded.

Comments
@KarinG @eileenmcnaughton

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -948,4 +948,22 @@ abstract class wf_crm_webform_base {
     return CRM_Core_Key::get('CRM_Contribute_Controller_Contribution', TRUE);
   }
 
+  /**
+   * Historically webform_civicrm was configured with the live payment processor ID and the is_test flag.
+   * But this is not how CiviCRM expects it to work and this can cause problems where payments use the live processor
+   *   instead of test.  So we "fix" the processor ID here to be the correct one for live/test.
+   * An "improved" fix would involve changing the configuration saved by the user but this avoids that.
+   *
+   * @throws \CiviCRM_API3_Exception
+   */
+  protected function fixPaymentProcessorID() {
+    // Check for is_test and payment_processor_id (pay later = 0 is set to '' here. is_test has no meaning for pay later).
+    if (!empty($this->data['contribution'][1]['contribution'][1]['is_test'])
+        && !empty($this->data['contribution'][1]['contribution'][1]['payment_processor_id'])) {
+      $paymentProcessor = \Civi\Payment\System::singleton()->getById($this->data['contribution'][1]['contribution'][1]['payment_processor_id']);
+      $paymentProcessor = \Civi\Payment\System::singleton()->getByName($paymentProcessor->getPaymentProcessor()['name'], TRUE);
+      $this->data['contribution'][1]['contribution'][1]['payment_processor_id'] = $paymentProcessor->getPaymentProcessor()['id'];
+    }
+  }
+
 }
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -43,6 +43,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $this->node = $node;
     $this->settings = $node->webform_civicrm;
     $this->data = $this->settings['data'];
+    $this->fixPaymentProcessorID();
     $this->enabled = wf_crm_enabled_fields($node);
     $this->all_fields = wf_crm_get_fields();
     $this->all_sets = wf_crm_get_fields('sets');
@@ -766,7 +767,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         }
       }
     }
-    
+
     foreach (wf_crm_location_fields() as $location) {
       if (!empty($contact[$location])) {
         $existing = array();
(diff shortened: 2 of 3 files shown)


colemanw:
@mattwire I changed the base of this PR to `7.x-5.x` because we're not planning another 4.x release. But now it needs rebasing...

mattwire:
@colemanw Thankyou - now rebased

mattwire:
@colemanw Is this ok to merge now?

KarinG:
This needs to be tested - does this require 15330 referenced above? Does this work for ESR (5.13.x for one more release)?

eileenmcnaughton:
Code makes sense to me

eileenmcnaughton:
Re 15330 - no that is not a pre-requisite for this

colemanw:
@mattwire I'm happy to merge this as it aims to fix a bug that only affects test processors but shouldn't affect live payments. Are you able to rebase it?

mattwire:
@colemanw Thankyou - I've rebased

includes/wf_crm_webform_base.inc
```diff
@@ -948,4 +948,22 @@ abstract class wf_crm_webform_base {
...
+   * Historically webform_civicrm was configured with the live payment processor ID and the is_test flag.
+   * But this is not how CiviCRM expects it to work and this can cause problems where payments use the live processor
+   *   instead of test.  So we "fix" the processor ID here to be the correct one for live/test.
+   * An "improved" fix would involve changing the configuration saved by the user but this avoids that.
+   *
+   * @throws \CiviCRM_API3_Exception
+   */
+  protected function fixPaymentProcessorID() {
+    // Check for is_test and payment_processor_id (pay later = 0 is set to '' here. is_test has no meaning for pay later).
+    if (!empty(wf_crm_aval($this->data, 'contribution:1:contribution:1:is_test'))
```
colemanw: Using `wf_crm_aval` inside an `empty()` is kinda redundant.

```suggestion
    if (!empty($this->data['contribution'][1]['contribution'][1]['is_test'])
```

includes/wf_crm_webform_base.inc
```diff
@@ -948,4 +948,22 @@ abstract class wf_crm_webform_base {
...
+        && !empty(wf_crm_aval($this->data, 'contribution:1:contribution:1:payment_processor_id'))) {
```
colemanw:
```suggestion
        && !empty($this->data['contribution'][1]['contribution'][1]['payment_processor_id'])) {
```

colemanw: @mattwire this should be faster for PHP to parse and easier for humans to parse too.

includes/wf_crm_webform_base.inc
```diff
@@ -948,4 +948,22 @@ abstract class wf_crm_webform_base {
...
+      $paymentProcessor = \Civi\Payment\System::singleton()->getById(wf_crm_aval($this->data, 'contribution:1:contribution:1:payment_processor_id'));
```
colemanw: Ok we've already established that it's not empty.
```suggestion
      $paymentProcessor = \Civi\Payment\System::singleton()->getById($this->data['contribution'][1]['contribution'][1]['payment_processor_id']);
```

colemanw:
@mattwire thanks for putting up with my nitpicks.

eileenmcnaughton:
@colemanw I think this one is important enough (despite the fact it languished for a year) to trigger a webform_civicrm release - you could probably up the minimum civi version when you do as it's probably better for ESR sites not to get any updates

aydun:
This works for me if both the payment processor and mode are set in the CiviCRM / Contribution tab of webform.  But setting either to "User Select" does not work - ie attempts to use Live mode.

Also noted that Test Mode requires zip code (digits only), Live Mode requires Postcode (correctly for my setup).

mattwire:
> Also noted that Test Mode requires zip code (digits only), Live Mode requires Postcode (correctly for my setup).

@aydun This is stripe elements specific I think?

> But setting either to "User Select" does not work - ie attempts to use Live mode.

@aydun so better than before? The issue with the "User Select" I think requires a different fix because I think it's loaded via javascript.

KarinG:
Turns out I have a client who always uses test mode on webforms and that works properly; it grabs the test credentials for the selected payment processor; is this a Stripe specific issue?

aydun:
> > Also noted that Test Mode requires zip code (digits only), Live Mode requires Postcode (correctly for my setup).
> 
> This is stripe elements specific I think?

@mattwire Not sure, but it's a default implementation on a  basic webform.

> > But setting either to "User Select" does not work - ie attempts to use Live mode.
> 
> so better than before? The issue with the "User Select" I think requires a different fix because I think it's loaded via javascript.

@mattwire Yes, better than before (ie can do test Stripe transactions in some configurations), but more is needed to work with the 'user select' javascript scenarios.

> Turns out I have a client who always uses test mode on webforms and that works properly; it grabs the test credentials for the selected payment processor; is this a Stripe specific issue?

@KarinG Don't know whether this is Stripe specific, but Stripe happens to be the processor in use here.

KarinG:
I can confirm this is working with PayPal Pro and iATS Payments - explain to me why this does not work for Stripe?

PR #267, opened 2019-09-09
Tags & Groups support

mglaman:
This PR begins adding support for Tags & Groups. The first step is the removal of a call to `drupal_map_assoc`.

However, when the Tags element is added the live/static options is missing


KarinG:
Shall I merge this in or would you like to make this a WIP?

mglaman:
Let's keep it a WIP for right now and see what other errors might crop up, try to get a few merged at once

mglaman:
@KarinG so it seems to "just work" for tags and groups with this fixed. We could merge this up and provide User Select support in another PR

mglaman:
@KarinG so it seems to "just work" for tags and groups with this fixed. We could merge this up and provide User Select support in another PR

KarinG:
Great! I'm going to confirm this right now.

KarinG:
Confirmed ->  Contacts coming through this Form are added to Group Administrators and receive a Tag Webform. 

image

PR #268, opened 2019-09-13
Ability To Use Case Tags In Webforms

colemanw:
Overview
Enables tags & tagsets for activities & cases

This is a reviewers cut of #198


jamienovick:
@colemanw on it - just trying to work out how best to get this deployed on our infra :)

jamienovick:
@colemanw This has been tested and works correctly:

- Case Tags are exposed in the webform configuration screens
- When saved the tags are created
- When returning to the form with case ID passed in URL the tags are correctly selected
- If I save changes then these are reflected in the case

Let me know if you have any other test cases you would like me to check.

Thanks

PR #269, opened 2019-09-16
Use field options service for options elements

mglaman:
Instead of using custom logic in the elements, use the field options service which has proper handling for the different tables.

This was uncovered with tags. Specifically this handling

```
        if ($field['table'] === 'tag') {
          $split = explode('_', $name);
          $ret = wf_crm_get_tags($ent, wf_crm_aval($split, 1));
        }
```


KarinG:
This is great -> 

Webform is a Tag
Vanilla is a Custom Field -> select -> configured not to show as a select list -> and only one option is enabled. 

This is exactly what that should look like. And I did confirm it all works too! 

image

image

PR #270, opened 2019-09-18
Fix incorrect setting of payment instrument

jitendrapurohit:
Overview
Fix setting of incorrect payment instrument. 

Before
While doing a contribution using a payment processor with billing mode = 4, eg omnipay pxpay processor, webform civi initially creates a pending payment and then completes it using the ipn/webhook details. When the initial payment is processed, the payment instrument is incorrectly set as the default method(Check) instead of getting the preferred value from the payment processor.

And when the contribution is completed, civicrm thinks that the payment instrument is changed and hence creates an additional financial transaction which can be viewed on the view contribution page-

image

To replicate -

Create a simple webform with contribution page having payment processor which navigates to its site for payment.

After the payment is done and webform confirmation page is displayed, note the multiple row display on the view contribution page. 

After
The initial payment gets a correct payment method value and on completion, no processing is done as before and civi directly sets the status to completed.

The below row is displayed after doing payment using the same processor.

image

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -766,7 +766,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         }
       }
     }
-    
+
     foreach (wf_crm_location_fields() as $location) {
       if (!empty($contact[$location])) {
         $existing = array();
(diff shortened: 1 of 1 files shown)


jamienovick:
Hi @jitendrapurohit . How is this different from:
https://github.com/colemanw/webform_civicrm/pull/180

Thanks

J

jitendrapurohit:
Ah, didn't know there is a similar PR already existing. And the steps of capturing this issue seemed from the opposite direction as it was displaying multiple rows on the view contribution page which I think is missing a mention on that PR. 

The patch doesn't apply directly so I tried to manually apply the changes and confirmed that it fixes this part too. So that could be an additional reason to accept it. I'm closing this in favor of the same.

Thanks @jamienovick

PR #272, opened 2019-09-25
Handle Activity/Case tags

colemanw:
This is an untested port of #268

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -931,14 +931,18 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         $params['method'] = t('Webform');
         break;
       case 'tag':
-        $api = 'entity_tag';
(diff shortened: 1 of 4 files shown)


colemanw:
Anyone using D8 webform and able to test this out?

KarinG:
I'm using D8 webform :-) but I've never used Case with it (not even in D7), so not sure what I what I would need to check/keep an eye etc.

colemanw:
@KarinG this adds support for both activity and case tags. Since the code is the same for both I think it's fine to just test one of them (esp since this patch was already accepted for the 7.x branch). Try this:

- Create a couple tags for activities
- Create a webform and expose the activity tag fields
- Try tagging and untagging activities via webform

KarinG:
Ok I've tested Create a webform and expose the activity tag fields and am able to Tag activities via webform. I've not tried untagging as I'd be surprised it we could load existing Activities at this point, but that's ok. Merging this!

PR #273, opened 2019-10-04
Pagebreaks and fieldsets

mglaman:
Fieldset generation and contribution page break support. Ensures the contribution page break is inserted properly. Adds fieldset generation so that the fieldset can be easily changed into a page, for proper page breaks.

Needs experimenting. And currently, the fieldset generation is defaulted to TRUE. The check done in D7 might not be needed in D8. Tries to prevent adding the fieldset if it was already possibly added and removed.

Screen Shot 2019-10-03 at 10 41 56 PM

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1729,7 +1729,7 @@ class wf_crm_admin_form {
                  */
                 // @todo Properly handle fieldset creation.
                 // self::insertComponent($field, $enabled, $this->settings, !isset($previous_data['data'][$type][$c]));
-                self::insertComponent($field, $enabled, $this->settings);
+                self::insertComponent($field, $enabled, $this->settings, TRUE);
                 $created[] = $field['name'];
                 if (isset($field['civicrm_condition'])) {
                   $this->addConditionalRule($field, $enabled);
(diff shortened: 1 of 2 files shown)


KarinG:
Hi @mglaman - ok tested this!-> Indeed now creates a Fieldset for Contact 1 and it nests Contribution Page and Contribution Amount under Payment (which is a Wizard page). 

**One small edit:** swap Contribution Page and Contribution Amount (so amount fields shows before the credit card fields).

**One larger edit** required to make the Page flow work -> as soon as we inject contribution_page_break -> everything ahead of that contribution_pagebreak Wizard page  also needs to -> see screenshot. Not sure though how we can do this? Should we do this? It's easy to add the Wizard page manually and indent the elements. But it will look odd if someone does not know to do this. 

image

mglaman:
@KarinG 

1. `contribution_total_amount` _is_ supposed to be added before `contribution_total_amount`. We can't specify the weight when building the form, so it's ordered in that specific fashion in the Fields.php array. I'll have to see.

2. 😬I don't know what to do. We could try to wrap all existing elements into a new page? So that it becomes two pages by default? Make it a "documentation WTF" for now?

KarinG:
I’m totally ok to make this a documentation issue. Merging this - as it is an improvement over what we had.

mglaman:
@KarinG sounds good. I have been doing some review and I don't think there is a very good way to control the order of elements when added unless we re-do a lot of `\wf_crm_admin_form::postProcess`. Which wouldn't be bad but I don't know if it would be most effective.

I spent some time reviewing Webform UI

PR #275, opened 2019-10-10
Improving autorenewal user experince

omarabuhussein:
Overview
Adding a warning when the admin try to configure a webform with more than one membership with autorenew support, telling the user that both frequency interval and unit of the membership types should be the same in order for the autorenewal to work well.

Before
Admins can configure two memberships to autorenew but where each membership type has different frequency interval and unit. 

After
Admins can still do the above but a warning will appear warning for the admin : 

after

Notes
-----------
The only relevant commit in this PR is this one : 

https://github.com/colemanw/webform_civicrm/pull/275/commits/0ec4f9b19e27cf11410cea86b7221b38b961969a

the other two appear here because I created this branch from this PR branch : https://github.com/colemanw/webform_civicrm/pull/274

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -75,6 +75,9 @@ function wf_crm_field_options($field, $context, $data) {
     elseif ($table == 'membership' && $name == 'num_terms') {
       $ret = drupal_map_assoc(range(1, 9));
     }
+    elseif ($table == 'membership' && $name == 'auto_renew') {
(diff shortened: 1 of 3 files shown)


omarabuhussein:
@KarinG 

Here is the 2nd PR, the only relevant commit is this one : https://github.com/colemanw/webform_civicrm/commit/0ec4f9b19e27cf11410cea86b7221b38b961969a

PR #276, opened 2019-10-24
Fix relationship if multiple contact elements are included in the form

jitendrapurohit:
Overview
Fix relationship if multiple contact elements are included in the form

Before
- Multiple contact elements do not work on d8 webforms. I've added 2 contact existing field on the below webform, but only first contact element is loaded.

image

- Relationships are not populated in the edit element form.

image

- If second contact element form is edited and saved, it gets stored as "Generic Element"

image

After
- Relationships are correctly displayed.

image

- All contact elements on the webform are loaded.

image

Comments
@KarinG Sharing my inprogress work (in case you need). The remaining fix needed here is to populate the second contact element with the correct related cid.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -267,7 +267,7 @@ function _webform_edit_civicrm_contact($component) {
       '#options' => array('' => '- ' . t('None') . ' -'),
       '#default_value' => wf_crm_aval($component['extra']['filters'], 'relationship:contact'),
     );
-    $form['filters']['relationship']['type'] = array(
+    $form['filters']['relationship']['types'] = array(
       '#type' => 'select',
       '#multiple' => TRUE,
       '#title' => t('Specify Relationship(s)'),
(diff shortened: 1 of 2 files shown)


KarinG:
Awesome! Thank you - exciting.

jitendrapurohit:
Just pushed a fix that takes care of loading the relationships too. 

image

KarinG:
I'm in back to back meetings today through 8pm - but will promise to try/confirm first tomorrow morning.

jitendrapurohit:
Sure, this is pending initial review from my end too. So keeping wip in the title for now.

Thanks :)

KarinG:
Ok looking now!

Made a Contact in CiviCRM -> added Relationship Child of Parent (myself);

Logged in as myself:
 
image

Config for Parent: Default Value -> Set Default contact from: 
Current User

Config for Child:
image

Works!
image

Thank you for working on this @jitendrapurohit 
When you confirm it passes testing on your end as well - I'll merge it.

jitendrapurohit:
Thanks @KarinG There are still some fixes needed on contact element form regarding populating filter options when > 2 contacts elements are present, filter section missing, etc. But I think I can create a new PR for that. 

This PR fixes the loading of multiple contact elements on the webform main page and assigning default cid on it. I'll keep it as it is and create a new PR for more fixes related to contact component.

As it passes manual testing of what it provides and does not break anything else on the form, can you pls merge this as this is better in that out?

Thanks.

KarinG:
I completely agree 😀

PR #277, opened 2019-11-04
Fix contact element filters on webforms

jitendrapurohit:
Overview
Fix contact element filters on webforms.

Before
- Filter section missing.
- Default option breaks for multiple contacts.

Screenshot 2019-11-03 at 11 08 47 AM

After
- Filter section added and working on the webform

image

- Contact subtype, relationship, groups & tag filters work correctly on the webform.

image

Comments
ping @KarinG

### DIFF
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -63,12 +63,12 @@ class CivicrmContact extends WebformElementBase {
         'default_relationship' => '',
         'allow_url_autofill' => TRUE,
         'dupes_allowed' => FALSE,
-        'filters' => [
-          'contact_sub_type' => 0,
-          'group' => [],
-          'tag' => [],
-          'check_permissions' => 1,
-        ],
+        'filter_relationship_types' => [],
+        'filter_relationship_contact' => [],
+        'contact_sub_type' => '',
+        'group' => [],
+        'tag' => [],
+        'check_permissions' => 1,
         // Set for custom fields.
         'expose_list' => FALSE,
         'empty_option' => '',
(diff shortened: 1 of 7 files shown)


KarinG:
Awesome! I just gave this a spin and am confirming filters are now showing.

@mglaman - could you please have a look from a technical point of view?

image

src/Plugin/WebformElement/CivicrmContact.php
```diff
@@ -63,12 +63,12 @@ public function getDefaultProperties() {
...
-          'group' => [],
-          'tag' => [],
-          'check_permissions' => 1,
-        ],
+        'filter_relationship_types' => [],
+        'filter_relationship_contact' => [],
+        'contact_sub_type' => '',
+        'group' => [],
+        'tag' => [],
+        'check_permissions' => 1,
```
mglaman, line 71: 👍

webform_civicrm.module
```diff
@@ -64,6 +64,12 @@ function webform_civicrm_menu() {
  * Implements hook_form_alter().
  */
 function webform_civicrm_form_alter(&$form, &$form_state, $form_id) {
+  if ($form_id == 'webform_edit_form' && !empty($form['webform_ui_elements'])
+    && !empty((int) preg_grep('/^civicrm_/', array_keys($form['webform_ui_elements'])))) {
+    \Drupal::service('civicrm')->initialize();
+    \CRM_Core_Resources::singleton()->addCoreResources();
+  }
+
```
mglaman: 🤔 does this run on the Build tab? or is this for the component edit form?  Do we need the CiviCRM resources on that page? Or does it fix weird AJAX insertion of assets when the offcanvas form opens?

mglaman: Would it be "easier/better" to check if the webform has the CiviCRM handler versus preg_grep over each array key?

mglaman (commented):
This looks great for clean up of the content element.

I just had one nit/concern on the form alter.

webform_civicrm.module
jitendrapurohit: yes, the above loads civicrm resources on the build tab so that contact element works fine when opened on the same page via AJAX. I've replaced the above condition to use handler check but it seems too large. Do you have any short suggestions?

jitendrapurohit:
The update also has a fix for the filters when there are multiple spaces in the webform title Eg `Leave Request form`. The `_` in the webformId is not completely replaced with `-` and is now fixed in the latest commit.

webform_civicrm.module
mglaman: Looks good to me! I know loading assets over AJAX doesn't always work right, especially I would imagine it does not for CiviCRM resources.

mglaman (approved):

KarinG:
Thanks guys! Will re-pull in the latest commit by Jitendra for a final test - today;

KarinG:
Final tested!

PR #278, opened 2019-11-20
#3095675 Show Line Item in same order as form field are configured

sunilpawar:
Overview
https://www.drupal.org/project/webform_civicrm/issues/3095675

Before
Line Item table does not follow weight flag of field

After
Line Item table does follow weight flag of field

Technical Details
Line Item table follow weight flag of field

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -18,6 +18,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   private $totalContribution = 0;
   private $contributionIsIncomplete = FALSE;
   private $contributionIsPayLater = FALSE;
+  private $formFieldOrder = array();
 
   // During validation this contains an array of known contact ids and the placeholder 0 for valid contacts
   // During submission processing this contains an array of known contact ids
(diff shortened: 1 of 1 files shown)


KarinG:
Hi @sunilpawar - thanks for this. Will definitively take a look in near future.

KarinG:
@sunilpawar - I just tested this and ran into some issues. Here's my Webform:

image

And these are my Line Items afterwards with this PR 278:
image

Please note -> not only is the order not the same as on the Webform (which is what this PR is attempting), _**note that Line Item Amount 2 is missing**_. 

If I git stash -> to before 278.diff -> Line Item Amount 2 is there:
image

Note I've not yet tested multiple Participants or multiple Memberships. 

Conclusion: this would certainly be nice but in order to make this mergeable (after fixing the ordering and after handling multiple line items) please show some testing with (in addition to what I just did) multiple Memberships and multiple Participants as well.

KarinG:
PS - sorry for the delay on reviewing this. I'm finally catching up.

sunilpawar:
@KarinG  thanks for reviewing PR, i will check multiple scenario and get back to you with additional fix.

colemanw:
Looking at the code here, I have a suggestion.
Although it's very clever to index the array by component weight and then `ksort()` at the end, it results in overall more complex code, as the `$formFieldOrder` variable has to be used everywhere a line item is added to the array. Seems to make the code harder to read & understand.

I would suggest that instead, you index the line-item array by `$fid` and then add a utility function to do the reordering at the end. That function would look up component weights by `$fid` and then `ksort`. That function should also involve a fallback in case the component is undefined.

To prevent items disappearing, you could also suffix the array keys like `[$fid . '-a']` and `[$fid . '-b']`. The utility function would just need to be smart enough to handle that.

PR #279, opened 2019-12-01
Flatten elements now that Contribution config injects nested elements & Make recurring Contributions work in D8WFC

KarinG:
Before
To reproduce: Settings -> CiviCRM: Enable CiviCRM Processing; add Contact: First Name, Last Name + Email; add Contribution: select Contribution Page; Contribution Amount and Payment processor. Save Settings. All is well - Elements are created:

image

Go back to Settings -> CiviCRM -> and hit Save Settings

This produces a Fatal error -> as `civicrm_1_contribution_1_contribution_contribution_page_id` is a nested Element. 

Error: `TypeError: Argument 1 passed to Drupal\webform\Utility\WebformArrayHelper::removePrefix() must be of the type array, null given, called in /Applications/MAMP/htdocs/drupal853.local/modules/webform_civicrm/includes/wf_crm_admin_form.inc on line 1743`

After
No error. 

Technical Details
use `$elements = $this->webform->getElementsDecodedAndFlattened();`
instead of: `$elements = $this->webform->getElementsDecoded();`

Also
Fixed default value for instalments and frequency of recurring contributions fields.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1738,7 +1738,7 @@ class wf_crm_admin_form {
             }
           }
           elseif ($field['type'] === 'hidden' && !empty($field['expose_list'])) {
-            $elements = $this->webform->getElementsDecoded();
+            $elements = $this->webform->getElementsDecodedAndFlattened();
             $component = $elements[$enabled[$key]];
             $component = WebformArrayHelper::removePrefix($component);
             $component['value'] = $val;
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -659,21 +659,17 @@ class Fields implements FieldsInterface {
         $fields['contribution_installments'] = array(
           'name' => t('Number of Installments'),
           'type' => 'number',
-          'value' => '1',
-          'extra' => array(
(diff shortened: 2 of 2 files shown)


includes/wf_crm_admin_form.inc
```diff
@@ -1738,7 +1738,7 @@ class wf_crm_admin_form {
             }
           }
           elseif ($field['type'] === 'hidden' && !empty($field['expose_list'])) {
-            $elements = $this->webform->getElementsDecoded();
+            $elements = $this->webform->getElementsDecodedAndFlattened();
```
mglaman, line 1741: 👍 This change works for me! 

The basics of it are, from within the Webform class

```
// initElements()
$elements = Yaml::decode($this->elements);
$this->elementsDecoded = $elements; // returned by getElementsDecoded, full nested YAML basically

// initElementsRecursive
$this->elementsDecodedAndFlattened[$key] = WebformElementHelper::getProperties($element);
```

So flattened makes it easier to work with and we don't need to worry about nesting. This would probably make things easier in a few areas as well.

mglaman (approved):

PR #280, opened 2019-12-09
Changing Decimal numbers calculation to be more accurate

omarabuhussein:
Note 

**Hence that the work below is not tested by our QA team yet and I am opening this PR  as a try to get some concept approval first before proceeding**

Before
1- Create a membership type with fee = 55.68
2- Configure a webform with 12 installments and the membership defined above
3- Submit the webform
4- A contribution will be created with the amount = 4.75 and a recur contribution will be created with amount = 4.63 . both are wrong since : 

55.68/12 = 4.64

which should be the amount of the both the created contribution and the recur contribution.

2019-12-10 01_12_27-ads2@ad com ads2@ad com _ m7civi519

2019-12-10 01_12_35-ads2@ad com ads2@ad com _ m7civi519

The issue above is caused by the following lines : 

```
 $contributionRecurAmount = floor(($contributionParams['total_amount'] / $numInstallments) * 100) / 100;
$contributionFirstAmount = $contributionParams['total_amount'] - $contributionRecurAmount * ($numInstallments - 1);
```

The first line result is : 

$contributionRecurAmount = floor((55.68/ 12) * 100) / 100 = 4.63 

and the 2nd line : 
$contributionFirstAmount = 55.68 - 4.63  * (12-1) =  55.68 - 50.93 = 4.75

The main issue is caused by the wrong 4.63 which should have been 4.47, but since floating point numbers are used and floating points operations are not accurate  in general, 4.63 resulted instead.

CiviCRM core has many issues in general when it come to money calculation, one of them that it is  stores all money in database as decimal(10, 2) which make it hard to get correct results in some cases specially when dealing with payment plans with more than one installment, but at least accurate method of calculation should still be performed on money in case the core get changed in the future, and floating point arithmetic is  not accurate by design and should never be used when doing money calculations.

After
There are many options to resolve the floating point arithmetic issues, I decided to abstract the work inside a new class called **wf_crm_money_operations**  which contains methods for money addition, subtraction, division and multiplying. 

Technically this new class use bcmath PHP extension to do these operations and I replaced all the money calculations inside the module to use the methods in this new class.

2019-12-10 01_40_27-qwos02@ad com qwos02@ad com _ m7civi519

2019-12-10 01_40_38-qwos02@ad com qwos02@ad com _ m7civi519

### DIFF
--- a/includes/wf_crm_money_operations.inc
+++ b/includes/wf_crm_money_operations.inc
@@ -1,0 +1,44 @@
+<?php
+
+class wf_crm_money_operations {
+
+  /**
+   * The digits after decimal precision we are
+   * using, it is big enough to make sure
+   * rounding will work fine.
+   */
+  const CALCULATION_PRECISION_SCALE = 8;
+
+  /**
+   * CiviCRM stores all financial amounts
+   * inside database as decimal(10,2), which
+   * means only two digits are allowed after
+   * the decimal and thus we are choosing the
+   * same number for rounding the final amount.
+   */
+  const ROUNDING_DECIMAL_SCALE = 2;
+
+  public static function add($leftOperand, $rightOperand) {
+    $result = bcadd($leftOperand, $rightOperand, self::CALCULATION_PRECISION_SCALE);
+    return self::round($result);
+  }
+
+  public static function subtract($leftOperand, $rightOperand) {
+    $result = bcsub($leftOperand, $rightOperand, self::CALCULATION_PRECISION_SCALE);
+    return self::round($result);
+  }
+
+  public static function multiply($leftOperand, $rightOperand) {
+    $result = bcmul($leftOperand, $rightOperand, self::CALCULATION_PRECISION_SCALE);
+    return self::round($result);
+  }
+
+  public static function divide($leftOperand, $rightOperand) {
+    $result = bcdiv($leftOperand, $rightOperand, self::CALCULATION_PRECISION_SCALE);
+    return self::round($result);
+  }
+
+  public static function round($amount) {
+    return round($amount, self::ROUNDING_DECIMAL_SCALE, PHP_ROUND_HALF_UP);
+  }
+}
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -6,6 +6,7 @@
  */
 
 module_load_include('inc', 'webform_civicrm', 'includes/utils');
+module_load_include('inc', 'webform_civicrm', 'includes/wf_crm_money_operations');
 
 /**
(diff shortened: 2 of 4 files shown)


KarinG:
More accurate -> ok; I prefer smallest possible fix as it need to get ported to D8WFC as well.

I don’t like the idea of requiring a PHP extension that CiviCRM Core does not require.

omarabuhussein:
> More accurate -> ok; I prefer smallest possible fix as it need to get ported to D8WFC as well.

I know that your prefer smallest fix possible :) , and I like to think that for this to be done the right way then this is indeed the smallest fix possible since it only replace the mathematical operations with the methods defined in the new class but happy to hear any suggestion.

> I don’t like the idea of requiring a PHP extension that CiviCRM Core does not require.

I agree, but while the bcmath extension functions are not used in CiviCRM core, it is actually required according to this document : https://docs.civicrm.org/sysadmin/en/latest/requirements/

Edit

Actually I can see that **[bccomp](https://www.php.net/manual/en/function.bccomp.php)** is used here : https://github.com/civicrm/civicrm-core/blob/master/CRM/Financial/BAO/Payment.php#L374

KarinG:
This module is used by over 3.3k sites and has no unit tests - so yes I require the smallest possible edits and require the assurance that larger code changes have been live on a variety of sites prior to being merged. We're also actively working on D8WFC so PRs with D8 ports are preferred (and will be required once we release D8WFC officially). 

Your issue is with the instalment calculations -> just work on that. 

Note that what you are doing right now actually would break things in that you would be charging too much for many cases and too little for many others. 

The old method that I had implemented wasn't perfect but it was accurate in that it collected the correct total amount summed over 12 instalments. In many more cases than not the contributionFirstAmount simply can not be equal to the contributionRecurAmount. Take $100 e.g. over 12 months. $8.33/month ->as we can only transact two decimals. 8.33 x 12 = 99.96 -> $0.04 short.

omarabuhussein:
> The old method that I had implemented wasn't perfect but it was accurate in that it collected the correct total amount summed over 12 installments. In many more cases than not the contributionFirstAmount simply can not be equal to the contributionRecurAmount. Take $100 e.g. over 12 months. $8.33/month ->as we can only transact two decimals. 8.33 x 12 = 99.96 -> $0.04 short.

I see, I think the methodology we are following at the time being  is a bit different  so I will close the PR.

KarinG:
Any PR that proposes any changes in Math must ensure that the total $ to be collected across all instalments associated with the recurring series continues to add up to the total amount to be collected.

mattwire:
@omarabuhussein @KarinG On the subject of bcmath, as @omarabuhussein it is required by civicrm core.

PR #281, opened 2019-12-11
Removed outdated text

laryn:
If I understand correctly IPN payments and recurring payments are both supported now, right?


colemanw:
Thanks.

PR #283, opened 2019-12-20
Fix error after contact merge

mfrausto-mdw:
Overview
Based on https://github.com/colemanw/webform_civicrm/pull/210 Replaced deprecated code and using webform handlers.

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -8,6 +8,7 @@
  */
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\webform\Entity\WebformSubmission;
 use Drupal\webform_civicrm\Utils;
 
 /**
@@ -499,10 +500,53 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
       ->condition('contact_id', '%-' . $old_id . '-%', 'LIKE')
       ->execute();
     // Update contact reference field data
-    db_query("UPDATE {webform_submitted_data} d, {webform_component} c SET d.data = :new
-      WHERE d.data = :old AND d.cid = c.cid AND d.nid = c.nid AND c.type = 'civicrm_contact'",
-      array(':new' => $new_id, ':old' => $old_id)
-    );
+    $webforms = \Drupal\webform\Entity\Webform::loadMultiple();
+    foreach ($webforms as $webform) {
+      $elements = $webform->getElementsInitializedAndFlattened();
+      foreach ($elements as $key => $value) {
+        $handler_collection = $webform->getHandlers('webform_civicrm');
+        if (!$handler_collection->has('webform_civicrm')) {
+          continue;
+        }
+        $submissions = \Drupal::entityTypeManager()
+            ->getListBuilder('webform_submission')
+            ->getStorage()
+            ->loadByProperties([
+                'webform_id' => $webform->id(),
+            ]);
+        foreach ($submissions as $submission) {
+          $sid = $submission->get('sid')->value;
+          //Load submission and update old contact id with the new id.
+          /** @var Drupal\webform\Entity\WebformSubmission $submission */
+          $submission = \Drupal\webform\Entity\WebformSubmission::load($sid);
+          $suid = $submission->getOwnerId();
+
+          \Drupal::service('civicrm')->initialize();
+          try {
+            $old_user = civicrm_api3('User', 'getsingle', [
+                'contact_id' => $old_id,
+            ]);
+            $new_user = civicrm_api3('User', 'getsingle', [
+                'contact_id' => $new_id,
+            ]);
+            if (!empty($old_user) && !empty($new_user)) {
+              $old_uid = $old_user['id'];
+              $new_uid = $new_user['id'];
+
+              if ($suid == $old_uid) {
+                //Set $submission->summited_by to the new user id
+                $submission->setOwnerId($new_uid);
+                $submission->save();
+              }
+            }
+          }
+          catch (CiviCRM_API3_Exception $e) {
+            $errorMessage = $e->getMessage();
+            \Drupal::logger('webform_civicrm')->error($errorMessage);
+          }
+        }
+      }
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


KarinG:
@mfrausto-mdw - sorry for the late reply on this. I have tested this and will make a couple of specific suggestions, but in general:

1. This removes a direct db query 👍 
2. This now allows merges to happen [no fatal errors] 👍 
3. The updating Drupal User ID in the submitted results is handy and what we have been doing in D7. This PR does that (see screenshot) 👍 
4. Note that _**we're not updating Contact ID**_ [and my comment/review are to remove any language that suggests we do] (see screenshot). There are pros and cons re: updating Contact ID as well. But I'm happy to leave things like this for now. The fact that there is some historical merge evidence is not a bad thing.

Once you've made the updates to the comments we can merge this. 

**Screenshot shows:**

- Submission Results before
- Merging records in CiviCRM admin interface
- Submission Results after: two submissions now belong to the same user: karin (but still two different Contact IDs)

image

webform_civicrm.module
```diff
@@ -499,10 +500,53 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+        $submissions = \Drupal::entityTypeManager()
+            ->getListBuilder('webform_submission')
+            ->getStorage()
+            ->loadByProperties([
+                'webform_id' => $webform->id(),
+            ]);
+        foreach ($submissions as $submission) {
+          $sid = $submission->get('sid')->value;
+          //Load submission and update old contact id with the new id.
+          /** @var Drupal\webform\Entity\WebformSubmission $submission */
```
KarinG: Change comment to: update old Drupal User ID with new Drupal User ID

webform_civicrm.module
```diff
@@ -499,10 +500,53 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+            $new_user = civicrm_api3('User', 'getsingle', [
+                'contact_id' => $new_id,
+            ]);
+            if (!empty($old_user) && !empty($new_user)) {
+              $old_uid = $old_user['id'];
+              $new_uid = $new_user['id'];
+
+              if ($suid == $old_uid) {
+                //Set $submission->summited_by to the new user id
+                $submission->setOwnerId($new_uid);
```
KarinG: This comment can be removed - it's clear what's happening

KarinG (changes requested):

webform_civicrm.module
mfrausto-mdw: Comment changed

webform_civicrm.module
mfrausto-mdw: Comment removed

mfrausto-mdw:
@KarinG Changes done, please let me know if you see any other improvement.  Thanks!

KarinG:
I ran some more tests and am now getting a fatal error:

`Message | Expected one UFMatch but found 0`

What I'm testing is to reproduce the scenario above: the user logs in as KarinAst (again) and creates another submission which results in another Contact in CiviCRM. Trying to merge them produces the error [and kills the merge];

webform_civicrm.module
```diff
@@ -499,10 +500,52 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
-      array(':new' => $new_id, ':old' => $old_id)
-    );
+    $webforms = \Drupal\webform\Entity\Webform::loadMultiple();
+    foreach ($webforms as $webform) {
+      $elements = $webform->getElementsInitializedAndFlattened();
+      foreach ($elements as $key => $value) {
+        $handler_collection = $webform->getHandlers('webform_civicrm');
+        if (!$handler_collection->has('webform_civicrm')) {
+          continue;
+        }
```
mglaman: This needs to be reworked for performance. The handler check needs to happen before fetching the elements. Otherwise, we're re-fetching the handler at each element.

webform_civicrm.module
```diff
@@ -499,10 +500,52 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+        $submissions = \Drupal::entityTypeManager()
+            ->getListBuilder('webform_submission')
+            ->getStorage()
+            ->loadByProperties([
+                'webform_id' => $webform->id(),
+            ]);
+        foreach ($submissions as $submission) {
+          $sid = $submission->get('sid')->value;
```
mglaman: Can't we just use `$submission->id()` ?

webform_civicrm.module
```diff
@@ -499,10 +500,52 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+          //Update old Drupal User ID with new Drupal User ID
+          /** @var Drupal\webform\Entity\WebformSubmission $submission */
+          $submission = \Drupal\webform\Entity\WebformSubmission::load($sid);
```
mglaman: We already have the submission? why are we reloading it?

webform_civicrm.module
```diff
@@ -499,10 +500,52 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+          $suid = $submission->getOwnerId();
+
+          \Drupal::service('civicrm')->initialize();
```
mglaman: Let's run this at the beginning instead of constantly. It is reentrant but it's burning unnecessary cycles.

webform_civicrm.module
```diff
@@ -499,10 +500,52 @@ function webform_civicrm_civicrm_merge($type, $data, $new_id = NULL, $old_id = N
...
+            ]);
+            $new_user = civicrm_api3('User', 'getsingle', [
+                'contact_id' => $new_id,
+            ]);
+            if (!empty($old_user) && !empty($new_user)) {
+              $old_uid = $old_user['id'];
+              $new_uid = $new_user['id'];
+
+              if ($suid == $old_uid) {
+                $submission->setOwnerId($new_uid);
```
mglaman: Is this the Drupal `uid` or the CiviCRM ID?

mglaman (changes requested):
This needs updating for performance concerns. I also have a few questions.

webform_civicrm.module
mfrausto-mdw: Is the CiviCRM contact ID

colemanw: A clearer variable name would be helpful. "uid" usually stands for "user id".

KarinG:
I'm so confused right now. What is it that needs to happen to Webform Submission results when Contacts get merged in CiviCRM?

KarinG:
Ok great - this makes more sense now! I'm happy with the functionality -> screenshot below: Results before and after merge (in CiviCRM). @mglaman - are you happy with the technical implementation?

image

mglaman:
Works for me!

PR #284, opened 2019-12-23
Autofill related contacts on same page

nishant-bhorodia:
Overview
I have a webform where 5 contacts are added with relations as below:
- Contact 2 (Organization) is an **Employer of** Contact 1 (individual)
- Contact 3 (Individual) is **Parent of** Contact 1 (Individual)
- Contact 5 (Organization) is an **Employer of** Contact 4 (Individual)

**How it works**

- Contact 2 gets auto-filled when Contact 1 is selected.
- Contact 3 is **not** auto filled with the parent of Contact 1.
- Contact 5 is **not** auto filled with employer of Contact 4. 

_The above scenarios work fine where there is a page break on the webform but not when on the same page._

**How it should work**

- Contact 2 gets auto-filled with the employer when Contact 1 is selected.
- Contact 3 gets auto-filled with the parent of Contact 1.
- Contact 5 is auto-filled with the employer of Contact 4.

Before
before

After
after

Technical Details

1. Condition here: https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_webform_ajax.inc#L137 only checks for $c == 1 which makes autofill work for only Contact Removing $c == 1 makes it work for all contact fields.
2. In function wf_crm_find_relations it was giving zero results if relationship type was anything other than employee/employer. It happened because the parameters were missing the contact id parameter. Used $cid to fetch relatioships for given contact id.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -253,7 +253,7 @@ function _webform_edit_civicrm_contact($component) {
       '#parents' => array('extra', 'default_custom_ref'),
     );
   }
-  
+
   $case_count = isset($data['case']['number_of_case']) ? $data['case']['number_of_case'] : 0;
   if ($case_count > 0) {
     $form['defaults']['default']['#options']['case_roles'] = t('Case roles');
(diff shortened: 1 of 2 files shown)


jamienovick:
@colemanw @KarinG would be great to get this bug fix reviewed if possible.

colemanw:
Thanks for the fix. I tested this out and it works :)

colemanw:
@KarinG I also cherry-picked these commits into the 8.x branch.

jamienovick:
@colemanw @KarinG thanks guys; much appreciated.

nishant-bhorodia:
Thanks @colemanw @KarinG @jamienovick  :)

colemanw:
FYI @jamienovick this went into the 7.x-5.0 release.

PR #286, opened 2020-01-17
Replace usage of `watchdog` in saveCustomValues

mglaman:
If `CRM_Core_BAO_CustomValueTable` fails inserting values, Webform CiviCRM silently fails and reports the error. This was using `watchdog` and not properly using the logger.

Discovered while debugging https://www.drupal.org/project/webform_civicrm/issues/3099346


KarinG:
Thanks @mglaman 👍

PR #287, opened 2020-01-21
Prevent add element from failing

Aschab:
Overview
A solution for>
https://www.drupal.org/project/webform_civicrm/issues/3104870

Before
For certain users civicrm would not be loaded properly and "add element"  would fail to load

After
Make sure that civicrm is loaded before trying to load civicrm elements


KarinG:
Awesome - thank you @Aschab - I've had a really difficult time to consistently reproduce this issue and at the moment I can't. But this makes a lot of sense.

PR #288, opened 2020-01-27
Take field disabled status into account when considering if a contact can be created

colemanw:
Overview
An "existing contact" field has an `allow_create` flag telling it whether a contact can be created. This considers whether the contact has a name/email field on the form. Now it also considers whether those fields are disabled.

Comments
This does not address an existing bug that the `allow_create` flag is not auto-updated when enabling/disabling/deleting a webform component from the webform ui. You must go to the civicrm tab and resave for changes to take effect :(


KarinG:
I've managed to get this working on new forms, but I can't seem to shake `allow_create` on existing forms.

PR #289, opened 2020-01-29
Issue#3108529 : Fix Pay later option for CiviCRM versions >= 5.21.0

omarabuhussein:
Before
The contribution will not be created when using the pay later payment method on CiviCRM versions >= 5.21.0 .

The issue is reported here : https://www.drupal.org/project/webform_civicrm/issues/3108529

After
Using the pay later option is working fine and the contribution is created normally.

## Technical Notes 

The issue was interoducded 3 months ago after this PR get merged to CiviCRM core : 

https://github.com/civicrm/civicrm-core/pull/15697

one of the changes in the above PR updates CRM_Core_Payment::getPaymentInstrumentID()  method to return 0 instead of null when no payment method is found for the specified payment processor, but CiviCRM API treat 0 values differently from null values , it will ignore null values where it will try to set values that equal 0 into the database (which is normal given that 0 represents a value where null represents the absence of any value), and given that there is a foreign key constraint on the payment method/instrument field on the contribution table, CiviCRM will fail to create the contribution.

I solved the issue here by skipping the code that tries to get  payment method Id if the payment processor Id is not set (and hence that the pay later option Id  = 0 so it will always will skip this part) .

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2077,7 +2077,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
     }
 
-    if (empty($params['payment_instrument_id'])) {
+    if (empty($params['payment_instrument_id']) && !empty($params['payment_processor_id'])) {
       $params['payment_instrument_id'] = $this->getPaymentInstrument($params['payment_processor_id']);
     }
 


KarinG:
Awesome 👍 Thank you Omar. I could not reproduce this issue on live client sites. Makes sense now.

lalgwebdev:
Just to confirm the comment left on Issue https://www.drupal.org/project/webform_civicrm/issues/3108529 this change does fix my problem.

KarinG:
Thank you for confirming that @lalgwebdev 👍

PR #290, opened 2020-02-09
Nick name is not a required field.

KarinG:
nick_name is not a required field when creating new contacts -> see CiviCRM core docs:

https://docs.civicrm.org/user/en/latest/common-workflows/importing-data-into-civicrm/

image


colemanw:
Makes sense.

KarinG:
Thank you!

colemanw:

PR #294, opened 2020-03-17
Fix custom group fieldsets to not get recreated after adding additional CiviCRM Core fields.

jitendrapurohit:
Overview
fix fieldsets which are recreated on the webform

Before
To reproduce -

- Add 1 contact + enable some custom fields on the webform.
- This creates a fieldset for the custom set too. Move the fields to main contact fieldset and delete the custom fieldset. Webform should look something like -

image

- Now, add one more field from the civicrm setting page. Eg job title, nickname, etc. After saving the webform, the previous custom fieldset is recreated on the webform.

image

This is irritating when more no. of custom fieldsets are present on the webform and all of them are recreated after any field is added from the civicrm settings page. 

After
no fieldsets are added if custom field is already present on the page.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1996,8 +1996,8 @@ class wf_crm_admin_form {
       $field['value'] = $settings[$field['form_key']];
     }
     // Create fieldsets for multivalued entities
-    if ($ent !== 'contribution' &&
-      ($ent !== 'participant' || wf_crm_aval($settings['data'], 'participant_reg_type') === 'separate')
+    if (empty($enabled[$field['form_key']]) && ($ent !== 'contribution' &&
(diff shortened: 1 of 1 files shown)


KarinG:
:-) I've always wondered about who likes cg Fieldsets that always pop back in when you add a new cg field as a feature (also in D7). We always (no exception) delete cg Fieldsets on all our webforms - and do layout boxes and css to make things look good. @colemanw - can you think of any reason we would get pushback for removing excessive cg Fieldsets? I assume it was added originally to help people keep their webforms organized?

KarinG:
@jitendrapurohit - sorry for the delay on this. I just tested this. Confirming that with this PR cg Fieldset is NOT re-added after adding a CiviCRM Core field 👍 Thank you!

PR #295, opened 2020-03-18
Fixed entitTypeManager() typo

roblog:
Overview
Fixed typo with call to entitTypeManager() method.

Before
Site crashes with fatal error.

After
Site doesn't crash.


KarinG:
Thank you @roblog 👍

PR #297, opened 2020-03-22
Add a flag to identify that AJAX endpoint is being called by drupal webform

mattwire:
Overview
Some payment processors need to do certain things differently (eg. Stripe) if they are being loaded via webform_civicrm. Currently you can't detect on the server side if the caller is a drupal webform or a civi form.
The payment block is loaded via the endpoint `civicrm/payment/form` which calls functions such as `buildForm()` on the `CRM_Core_Payment` class. This flag allows code there to detect that it is being called by a webform.

Before
Not able to identify drupal webform as caller

After
Extra flag that can be checked.

Technical Details

Comments
@colemanw @KarinG What's the deal with 8.x-5.x as this should apply to both?


KarinG:
Good idea! 👍

colemanw:

PR #298, opened 2020-03-23
Fix test payment processor ID on form submission when there are multiple processors available

mattwire:
Overview
To reproduce:
1. Configure a webform with multiple payment processors available in mode "test".
2. Submit the form - you will see that the wrong payment processor (the live ID) is selected and this causes some processors (eg. Stripe) to fail.

Before
Cannot submit payment in test mode for some payment processors.

After
Can submit payment in test mode when there are multiple payment processors on the form.

Technical Details
We introduced the function `fixPaymentProcessorID()` a while ago, but it is only triggered during the form constructor. For a webform with multiple payment processors there is no payment processor ID when the form is built, only once it is submitted.

Comments
@KarinG @eileenmcnaughton This should be a straightforward fix with limited scope for breakage?

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -45,6 +45,7 @@ abstract class wf_crm_webform_base {
   function __get($name) {
     switch ($name) {
       case 'payment_processor':
+        $this->fixPaymentProcessorID();
         $payment_processor_id = wf_crm_aval($this->data, 'contribution:1:contribution:1:payment_processor_id');
         if ($payment_processor_id && !$this->_payment_processor) {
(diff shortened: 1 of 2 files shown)


mattwire:
Hey @KarinG @colemanw ok to merge this? It is a small follow on from the work I did previously for test mode.

KarinG:
Sorry for the delay @mattwire - it has been crazy busy. I’ve got 4h blocked off on webform civicrm module tomorrow and will put this top of the list

KarinG:
Ok looking now!

KarinG:
Live tested. Merging. Thanks @mattwire 👍

PR #299, opened 2020-03-26
Fix issues with civicrm_option widget.

jitendrapurohit:
Overview
Fix multiple and empty option label for civicrm option widget.

Before
1.  No UI input to select empty option label if civicrm option is displayed as listbox. So if civicrm custom field is displayed as select element, the first option is selected by default.

image

2. https://www.drupal.org/project/webform_civicrm/issues/3116288 No option to display civicrm option widget to display as checkboxes.
 

After
1.  Empty option label is added on the form.

image

If value is entered in the above field, the same is selected on the webform display -

image

2. Multiple checkbox is now displayed on the component form which is selected by default. If listbox option is disabled for the checkbox custom field, the input field is rendered as checkbox on the webform

image


KarinG:
Thank you @jitendrapurohit 👍

PR #302, opened 2020-05-08
Return if entity variable is NULL.

KarinG:
Overview
Circa 5.24 CiviCRM Core added var type declarations to some core functions including `civicrm_api()` so the signature changed from `civicrm_api($entity, $action, $params)` to` civicrm_api(string $entity, string $action, array $params)`

Before
Fatal Error when trying to Add Element on Build tab

```
TypeError: Argument 1 passed to civicrm_api() must be of the type string, null given, called in /var/www/drupal/web/modules/contrib/webform_civicrm/includes/utils.inc on line 532 in civicrm_api() (line 21 of /var/www/drupal/vendor/civicrm/civicrm-core/api/api.php)
#0 /var/www/drupal/web/modules/contrib/webform_civicrm/includes/utils.inc(532): civicrm_api(NULL, 'getoptions', Array)
```

After
Happy dance!

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -525,6 +525,10 @@ function wf_crm_array2str($arr) {
  *   Result of API call
  */
 function wf_civicrm_api($entity, $operation, $params) {
+  if (is_null($entity)) {
+    return;
+  }
+
   $params += array(
     'check_permissions' => FALSE,
     'version' => 3
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -525,8 +525,8 @@ function wf_crm_array2str($arr) {
  *   Result of API call
  */
 function wf_civicrm_api($entity, $operation, $params) {
-  if (is_null($entity)) {
-    return;
+  if (!$entity) {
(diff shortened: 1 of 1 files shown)


includes/utils.inc
```diff
@@ -525,6 +525,10 @@ function wf_crm_array2str($arr) {
  *   Result of API call
  */
 function wf_civicrm_api($entity, $operation, $params) {
+  if (is_null($entity)) {
+    return;
+  }
```
colemanw: This is just a weird situation. But since any empty value for $entity is nonsensical, and this function promises to return an array, I'd suggest:
```suggestion
  if (!$entity) {
    return [];
  }
```

KarinG: Turns out it is not just a weird situation but also an intermittent situation -> Webform module is not always trying add CiviCRM Elements via that Add Element button on Build. First reported last Fall by @jitendrapurohit - this may be a permanent avoid the situation and that is a good solution 👍

PR #304, opened 2020-05-11
Another small rector pass; also add core_version_requirement; remove …

KarinG:
…old D7 info file.

Overview
This is a small bit of maintenance:

- another small rector pass; still running into `Fatal error: Uncaught Error: Call to undefined function module_load_include()` issue but picked up another deprecation
- also -> add core_version_requirement (same requirement as webform module)
- and removed the old D7 .info file

### DIFF
--- a/webform_civicrm.info.yml
+++ b/webform_civicrm.info.yml
@@ -1,7 +1,7 @@
 name: 'CiviCRM Webform'
 type: module
 description: 'CiviCRM and Webform integration'
-core: 8.x
+core_version_requirement: ^8.7.7
 package: 'Custom'
 dependencies:
   - webform
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1239,7 +1239,7 @@ class wf_crm_admin_form {
       if ($submit_limit > 0 && $submit_interval != 0) {
         $webform['submit_limit'] = $submit_limit;
         $webform['submit_interval'] = $submit_interval;
-        db_update('webform')
+        \Drupal::database()->update('webform')
           ->condition('nid', $this->node->nid)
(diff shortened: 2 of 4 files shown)


webform_civicrm.info.yml
```diff
@@ -1,7 +1,7 @@
 name: 'CiviCRM Webform'
 type: module
 description: 'CiviCRM and Webform integration'
-core: 8.x
+core_version_requirement: ^8.7.7
```
jitendrapurohit, line 4: `core` seems to be the required key?

I get this after applying the patch -

>Uncaught PHP Exception Drupal\Core\Extension\InfoParserException: "Missing required keys (core) in modules/contrib/webform_civicrm/webform_civicrm.info.yml" at /Applications/MAMP/htdocs/d8civicomposer/web/core/lib/Drupal/Core/Extension/InfoParserDynamic.php line 29

@KarinG Can you pls confirm?

KarinG: I suspect you're testing this on < 8.7.7? `core:` is required for Drupal Core versions before 8.7.7 ( they do not recognize the `core_version_requirement: key` )

So this PR may be a bit early and we can definitely wait. As we're moving to D9 and replacing more and more deprecated APIs we do need people on ^8.7.7 (pretty sure - looking at other modules).

jitendrapurohit: Ah ok. makes sense 👍

KarinG:
What do you think we should do in terms of timing? We obviously don't want to break any Fuzion sites :-)

jitendrapurohit:
I was testing this on my local build which had a slightly older version than 8.7.7. All FZ sites are updated to latest so this shouldn't be a problem for us :). Thanks. 

I've done some more testing(updated submission limit, tested users for the limit setting, etc) on one of our dev sites and confirmed that this isn't causing any issues. Yay.

PR #305, opened 2020-05-13
Fix saving #form_key when editing civicrm_contact webform element

dsnopek:
With Webform 8.x-5.11, if you edit the automatically added 'Existing Contact' element and save it (without any changes), the '#form_key' will get set to 0 (removing it's correct value, which is needed for that element to operate correctly).

This patch makes it so that saving will maintain the correct value. However, it doesn't fix elements where the #form_key has already gotten cleared out to 0.

### DIFF
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -34,7 +34,7 @@ class CivicrmContact extends WebformElementBase {
   public function getDefaultProperties() {
     return [
         'name' => '',
-        'form_key' => NULL,
+        'form_key' => '',
         'pid' => 0,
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you @dsnopek -> confirming that this fixes newly added Existing Contact elements to work as expected -> 

image

KarinG:
@jitendrapurohit - this is a critical issue I think -> so I'll push a new -dev and new -alpha to drupal.org

KarinG:
PS - if form_key has gone missing -> you can fix it by going to Build -> YAML -> Source -> note the #form_key:

change 
`#form_key: 0
`
to 

```
...   
...
```

PR #306, opened 2020-05-14
Make sure recurring amount always equals first amount; also calculate…

KarinG:
Make sure recurring amount always equals first amount; also calculate non_deductible_amount.

Now that it has been clarified that the amount field in the civicrm_contribution_recur table contains the amount to be transacted including any sales taxes we can finish the instalment bits for webform_civicrm module -> https://github.com/civicrm/civicrm-core/pull/17311

**Overview**
Logic for recurring contributions in core really expects them to be the same amount every month. This PR ensure that this is will now be the case. Also -> `non_deductible_amount` needs to be calculated properly. The use of `PHP_ROUND_HALF_UP` was previously proposed by @omarabuhussein.

**Before**
Logic: take the total amount -> split it up over x instalments, rounded to 2 decimals -> re-calculate the first amount to be total amount minus 11 instalments. That meant that the first amount could be different from the recurring amount.

**After**
Logic: much simpler, more predictable! Works much better with the expected recurring workflows in Core. Repeattransaction e.g. -> as the amount is the same month over month. 

**Downside**
Could end up with a few pennies more or less over a full year of rounded values. This however can be avoided by thinking ahead and pricing your annual fees/instalments accordingly :-)

**Example**
Professional Membership: $129 + 5% tax = 135.45

Before this PR:
$11.37 (first instalment) + $11.28 x 11 (instalments) = $135.45

After this PR:
$11.29 x 12 (instalments) = $135.48

The benefits of this integrating much better with existing recurring contribution workflows because the recurring amount is the same month over month far outweigh the $0.03 additional sales tax. Again, orgs can think ahead and adjust their pricing accordingly if they feel that is necessary. 

Comments
Live on one of our projects!

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1821,19 +1821,21 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     $numInstallments = wf_crm_aval($contributionParams, 'installments', NULL, TRUE);
     // if ($installments === 0) or $installments is not defined we treat as an open-ended recur
     $contributionFirstAmount = $contributionRecurAmount = $contributionParams['total_amount'];
-    $salesTaxFirstAmount = wf_crm_aval($contributionParams, 'tax_amount', NULL, TRUE);;
+    $salesTaxFirstAmount = wf_crm_aval($contributionParams, 'tax_amount', NULL, TRUE);
+    $nondeductibleFirstAmount = wf_crm_aval($contributionParams, 'non_deductible_amount', NULL, TRUE);
     if ($numInstallments > 0) {
       // DRU-2862396 - we must ensure to only present 2 decimals to CiviCRM:
-      $contributionRecurAmount = floor(($contributionParams['total_amount'] / $numInstallments) * 100) / 100;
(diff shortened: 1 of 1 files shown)


omarabuhussein:
hello @KarinG ,

Yeah we are using similar approach currently in our Membershipextras extension for the same reasons you mentioned in both the "After" and "Downsides" sections in your PR description, but we were actually planning to change that in our next Membershipextras release so it work similar to how the webform module currently work (in which the extra pennies are added to the first contribution). 

But now your PR here again made me rethink the plans we have in place, So from our side at Compucorp I think we still need to discuss it between ourselves internally first.

KarinG:
@omarabuhussein - they key issue was that $amount = the same -> is a strong assumption in the process of generating the next instalment/next Contribution. When sales taxes are involved this becomes critical. If you don't end up being able to template exactly off instalment #1 -> then you end up in core pathways that generate a new contribution and that leads to additional sales tax calculations. 

Apologies for not recognizing this earlier. At that time I had only dealt with membership instalments that were non-taxable and donation instalments. I am glad though that I remembered what you had done 👍 thank you!

Do you approve of merging this?

PR #309, opened 2020-05-18
D8port pr245

KarinG:
**Overview**
Backport of PR 245

**Before**
SQL

**After**
API

**Technical Details**
Had to comment out `wf_crm_get_empty_sets()` for now. Fields/Sets were completely restructured by Matt into a FieldsInterface so it's definitely not `Utils::wf_crm_get_fields('sets')` anymore - that blows up (reference to itself). 

**Comments**
Big PR - needs lots of testing.

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -81,9 +81,9 @@ class Fields implements FieldsInterface {
       if (in_array('CiviContribute', $components, TRUE)) {
         $sets['line_items'] = ['entity_type' => 'line_item', 'label' => t('Line Items')];
       }
-
-      $extra_sets = wf_crm_get_empty_sets();
-      $sets += $extra_sets;
+      // ToDo: figure out why this function blows up -> Error: "Maximum function nesting level of '2000' reached, aborting!"
+      // $extra_sets = wf_crm_get_empty_sets();
+      // $sets += $extra_sets;
       $this->sets = $sets;
     }
 
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -994,6 +994,8 @@ class Fields implements FieldsInterface {
           $fields[$id]['extra']['rows'] = $custom_field['note_rows'] ?? 4;
         }
       }
+      // The sets are modified in this function to include the custom sets.
+      $this->sets = $sets;
     }
     return $$var;
   }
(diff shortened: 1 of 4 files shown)


KarinG:
@colemanw - why is it that we need to find empty_sets?

colemanw:
Comment says `// add empty fieldsets for custom civicrm sets with no fields, if "add dynamically" is checked` so it's to support being able to add an empty custom group to the form, then have the fields populated later.

colemanw:
Strictly speaking, you could just skip over that part of the port and `wf_crm_get_empty_sets()` could stay as-is, as that query won't be affected by the schema changes in core.

src/Fields.php
```diff
@@ -516,11 +524,11 @@ protected function wf_crm_get_fields($var = 'fields') {
...
       }
-      $all_tagsets = wf_crm_apivalues('tag', 'get', [
-        'return' => ['id', 'name', 'used_for'],
+      $all_tagsets = wf_crm_apivalues('tag', 'get', array(
+        'return' => array('id', 'name', 'used_for'),
         'is_tagset' => 1,
-        'parent_id' => ['IS NULL' => 1],
-      ]);
+        'parent_id' => array('IS NULL' => 1),
+      ));
```
colemanw: Looks like this part of the patch is accidentally undoing short array syntax.
IMO we should autoformat both 7.x and 8.x branches to replace all arrays with short syntax. That should make porting a little simpler hopefully.

KarinG: Yes - sorry about that. I've (re)fixed my PhpStorm settings so that should not be happening again. I like the idea of autoformating both 7.x and 8.x branches to replace all arrays with short syntax.

KarinG:
Ok - I've reverted function `wf_crm_get_empty_sets() { ` back to SQL - like you mentioned that's not the critical bit here. I've also undone any accidental array syntax conversions. 

I think it's ready for (final) Review now.

KarinG:
@jitendrapurohit - could you please review/test this D8 port of PR245 -> it's substantial, involves custom fields so want to ensure we get it some more testing/eyes before we merge it.

jitendrapurohit:
I've applied this on one of our dev site setup and haven't seen any issues in a quick initial review. I'll perform a more detailed testing tomorrow and update with the status. Thanks for the PR @KarinG

jitendrapurohit:
tested with a complex webform involving custom fields, activity custom fields and relationships. Looks like this works fine in all the cases. Though I received some notice errors, but those seems to be also displaying without these changes. I'm going to merge this since it has been open from quite a while now. Thank you @KarinG 👍

PR #310, opened 2020-05-18
Switch findDuplicateContact to use api

colemanw:
Overview
Fixes https://www.drupal.org/project/webform_civicrm/issues/3137284 by switching code to use the api.

Before
Deduping was cobbled together and called CiviCRM internal functions, namely `CRM_Dedupe_Finder::dupesByParams()`

After
Uses the api for more consistent results. Note that the api ultimately calls this same internal function after it normalizes and formats the params.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -566,74 +566,52 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   }
 
   /**
-   * Search for an existing contact using default strict rule
+   * Search for an existing contact using configured deupe rule
    * @param array $contact
    * @return int
    */
   private function findDuplicateContact($contact) {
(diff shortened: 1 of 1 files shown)


colemanw:
This was tested and approved by the OP at https://www.drupal.org/project/webform_civicrm/issues/3137284. Does anyone else like @jitendrapurohit or @KarinG want to try it out before we merge?

jitendrapurohit:
I was able to replicate the issue mentioned in https://www.drupal.org/project/webform_civicrm/issues/3137284. 

Tested the patch with a webform involving multiple duplicate field criteria and confirm that the duplicate check is done correctly on them.

KarinG:
PS - Did you test with custom fields in the dedupe rule as well?

KarinG:
Just noting that I just confirmed this is all working on a custom field/dedupe as well 👍

PR #308, opened 2020-05-17
Replacing old D7 functions with D8 functions even when unreachable [r…

KarinG:
**Overview**
----------------------------------------
Small iterative step re: code cleanup and towards D9
My strategy will continue to be small iterations as I believe that's the safest way fwd. 

**Before**
----------------------------------------
Rector would choke on D7 functions -> like `variable_set` and `variable_get`
even when unreachable 

**After**
----------------------------------------
replaced all `variable_set` and `variable_get` with D8 `\Drupal::state()->set` and `\Drupal::state()->get`

**Technical Details**
----------------------------------------
Straight up replacement [for now] - not attempting to port the actual functionality.

PR #312, opened 2020-05-18
Replace old D7 functions and deprecated node_load (for D9 readiness)

KarinG:
**Overview**
Replaces PR-> https://github.com/colemanw/webform_civicrm/pull/308/files

Small iterative step re: code cleanup and towards D9
My strategy will continue to be small iterations as I believe that's the safest way fwd.
Replace deprecated node_load - no refactoring

**Before**

- Rector would choke on D7 functions -> like `variable_set` and `variable_get` even when unreachable
- `node_load()`

**After**
- replaced all `variable_set` and `variable_get` with D8 `\Drupal::state()->set` and `\Drupal::state()->get`
- `use Drupal\node\Entity\Node;` and `Node::load()`

**Technical Details**
Straight up replacement - not attempting to port the actual functionality. Since webforms are no longer nodes in D8 - much of this can be refactored - but this is just a drop-in replacement

Used: https://www.drupal.org/docs/8/converting-drupal-7-modules-to-drupal-8/step-5-how-to-upgrade-d7-variables-to-d8s-state

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -8,6 +8,8 @@
 use Drupal\Component\Utility\Html;
 use Drupal\webform_civicrm\Plugin\WebformElement\CivicrmContact;
 use Drupal\webform_civicrm\Utils;
+use Drupal\node\Entity\Node;
+
 
 module_load_include('inc', 'webform_civicrm', 'includes/utils');
 module_load_include('inc', 'webform_civicrm', 'includes/wf_crm_admin_help');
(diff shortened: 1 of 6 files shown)


KarinG:
PS - update on getting ready for D9: we're down another 20 errors - according to the Drupal 9 Deprecation Status check on Acquia -> https://dev.acquia.com/drupal9/deprecation_status?name=webform_civicrm&result=&error=&type=&group=&plan=&patch=&op=Submit

image

KarinG:
@jitendrapurohit - if you could also please try this small D9-readiness patch that would be great. It only replaces `node_load()` and `variable_get` and `variable_set` - should be a safe iteration towards D9.

jitendrapurohit:
Agree with the update. Tested this and confirmed the updated usage of the function works fine.

PR #315, opened 2020-06-13
Change to use 'name' instead of 'label' as Membership Status Id

ghost:
Overview
This corrects a malfunction in renewing a Membership during the Grace period if 'Grace' has been re-labelled as e.g. 'Overdue'.

D7 or D8?
This fault was found on D7, but relates to the CiviCRM APIv3 rather than any Drupal facility.

Before
If Grace is re-labelled as 'Overdue', and a renewal payment is made using the Pay Later facility, then the payment is processed correctly, but the Membership end date, and status, is left unchanged.
There is an API error message in the error log.

After
The renewal is processed correctly.

Technical Details
The CiviCRM Membership API uses the built-in 'name' of a Membership Status as the Id, not the 'label' which may be changed by the site administrator.  This change brings webform_civicrm in line with this usage.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -597,7 +597,7 @@ abstract class wf_crm_webform_base {
     $active = $expired = [];
     foreach ($existing as $membership) {
       $membership['is_active'] = $status_types[$membership['status_id']]['is_current_member'];
-      $membership['status'] = $status_types[$membership['status_id']]['label'];
+      $membership['status'] = $status_types[$membership['status_id']]['name'];
       $list = $membership['is_active'] ? 'active' : 'expired';
       array_unshift($$list, $membership);
     }


KarinG:
Reviewing this now. I can not reproduce the issue (D7 / CiviCRM 5.28.3 / 7.x-5.x webform-civicrm):

1. Membership Status Rules:
image

2. Gave my contact a Membership -> you can see the Status is Grace Period - on this renewal form:
image

3. After submitting the renewal - Membership is +1y and Membership status is updated from Grace Period to Current.
image

4. No errors 

If you can spot what's different between our scenarios - please let me know.

KarinG:
Ah got it - we only hit this if it's Pay Later. 

`The CiviCRM "membership create" API returned the error: "'Grace Period' is not a valid option for field status_id" when called by function "preSave" on line 201 of wf_crm_webform_postprocess.inc with parameters: "Array ( [num_terms] => 1 [membership_type_id] => 6 [status_override_end_date] => [id] => 243 [status_id] => Grace Period [end_date] => 2020-09-01 [contact_id] => 2 [skipStatusCal] => 1 [check_permissions] => [version] => 3 ) "`

KarinG:
@colemanw - should we make this change or should the API be able to handle both `name` and `label`?

PR #317, opened 2020-06-27
Prevent creation of blank Billing Address or Billing Email

ghost:
Overview
Checks the content of a Billing Address and a Billing Email before creating them, and does not do so if they are empty.

D7 or D8?
This PR relates to D7WFC.  I have not looked at the equivalent for D8.

Before
The base Payment Processor creates a blank Billing Address, as do Immediate payment derivatives such as Cash or Cheque.  If no Email address is given (e.g. in an offline action by an administrator) there is also a blank Billing Email.  The recent provision by the Stripe PP for running without the Billing Address also creates a blank one.

After
Blank Billing Address and Email are not created.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1757,7 +1757,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     if ($cid) {
       $address['contact_id'] = $email['contact_id'] = $this->ent['contact'][1]['id'] = $cid;
       // Ignore Address if all fields NULL, apart from Contact Id and Location Type Id.
-      if(count(array_filter($address,function($v){return $v !== null;})) > 2) {
+      if (count(array_filter($address)) > 2) {
         wf_civicrm_api('address', 'create', $address);
       }
       // Ditto for Email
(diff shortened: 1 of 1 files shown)


MegaphoneJon:
@maynardsmith If a contact doesn't already exist, then `$address['contact_id']` won't be defined.  So the `2` in `if (count(array_filter($address)) > 2) {` isn't going to fix the problem when a new contact is making a payment.

It's not pretty or clever, but `if ($address['street_address'] || $address['city'] || $address['country_id'] || $address['state_province_id'] || $address['postal_code']) {` is a more correct approach.

ghost:
@MegaphoneJon  Ok, I'm quite happy with an alternative solution, especially if it plugs a hole.

MegaphoneJon:
@maynardsmith to be clear - my PR isn't an alternative, it's a D8+ version.  I recommend you update yours to match, since I imagine the D8+ version will get looked at first, and if the code is basically identical this'll probably get merged at the same time.

Also if you make your change I'll vouch for the code as having been reviewed.

ghost:
Belated Thanks.  We are actually running on D7 with a rule which tidies blank addresses up afterwards, so now there is a fix for D8 I am withdrawing this PR. as outdated.

PR #292, opened 2020-02-10
Allow processor to determine whether email address is required

aydun:
See https://lab.civicrm.org/dev/core/issues/1584

PR #318, opened 2020-06-27
Prevent creation of duplicate Contacts

ghost:
Overview
In some circumstances, when creating a Contact and taking payment, the webform creates two Contacts, one associated with the Contribution, and the other having the remainder of the attributes that were specified.

D7 or D8?
This PR is for D7WFC.  I have not looked at the equivalent for D8.  

Before
Occurs when a Webform is used to create a Contact and make a Contribution to pay for a Membership, using an Immediate payment Payment Processor, such as Cash or Cheque, without giving an Email Address (e.g. as an offline administrator action).  It seems likely that it could also apply to other types of payment.

The effect is that two Contacts are created, the first being attached to the Contribution, and the second having the Membership.  In more complicated cases the second is also connected to Relationships being set up at the same time, etc.

If an Email Address is specified on the webform, in addition to First and Last Names, then only one Contact is created, correctly.

After
A single Contact is created, correctly.

Technical Details
In summary:
The first thing the webform does is to create a ‘Billing Contact’ in order to generate a Contact Id, to be used in creating the Contribution.  A little while later this Contact is updated with the further details, such as adding the Membership.

The problem is that it fails to preserve the Contact Id properly between the two steps, so the second step creates a second Contact.  It works if an Email is present because it also tries applying the de-duplication rule, which makes the connection with the first Contact instead.

In more detail:
•	The Billing Contact is created in the validate() procedure, Line 96, which calls createBillingContact() at Line 1715.  

•	createBillingContact() checks to see if it can find an existing Contact (i.e. if this submission is an update, not a create) and if not creates a new Contact at Line 1742.  The Contact Id is then stored in $this->ent['contact'][1]['id'] at Line 1758.

•	The update with further details takes place in the preSave() procedure, Line 113, the key point being in the call to getExistingContactIds() at Line 119.

•	Up to this call the previously stored Contact Id is still present in $this->ent['contact'][1]['id'], after the return this value is set to zero.

•	getExistingContactIds() scans the ‘Existing Contact’ fields in the webform and stores the Contact Id’s they contain in $this->ent['contact'][xxx]['id'].  The problem is that it doesn’t allow for the fact that a value may already be present, and in this scenario overwrites the previously stored Id with zero, because the Existing Contact field on the form is empty. (Line 526.)

•	In most cases this does not cause a problem, because the preSave() routine goes on to check empty Contact Ids by invoking the de-duplication rule at Line 136.  The Cash PP does not require an Email address, so the usual de-duplication rule fails.

Comments
The proposed solution is to check whether $this->ent['contact'][xxx]['id']  already contains a value before overwriting it.  This could result in overwriting one valid Id with a different one, but I can’t see how that should arise, or what the resolution would be if it did.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -523,7 +523,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       if (substr($field_key, -8) == 'existing') {
         list(, $c, ) = explode('_', $field_key, 3);
         $cid = wf_crm_aval($this->submissionValue($fid), 0);
-        $this->ent['contact'][$c]['id'] = $this->verifyExistingContact($cid, $field_key);
+        // Do not overwrite existing Id.  Otherwise can overwrite a valid value with zero.
+        if (!$this->ent['contact'][$c]['id']) {
+          $this->ent['contact'][$c]['id'] = $this->verifyExistingContact($cid, $field_key);
+        }
         if ($this->ent['contact'][$c]['id']) {
           $this->existing_contacts[$c] = $cid;
         }


colemanw:
Is someone able to test this?

lalgwebdev:
This PR also fixes the variant of this scenario described in the issue https://www.drupal.org/project/webform_civicrm/issues/3108332

KarinG:
If one tries to add a Contribution to a Webform -> the form requires an Email field for Contact 1 to process transactions.

https://github.com/colemanw/webform_civicrm/blob/3da7c660406217a23d759dcf2772b7221aafec11/js/webform_civicrm_admin.js#L520

image

I don't know the historical reason for it - and would be open to removing that requirement - but there must have been a good reason why it has historically enforced email addresses for forms processing transactions. @colemanw - any bells ringing? :-)

lalgwebdev:
@KarinG To clarify our requirement.  I am happy that the presence of the field is mandatory, since most  members have email addresses we wish to record.  A fair number however do not and we need the system to work correctly if an email is not provided.

KarinG:
What dedupe rule are you using for this Contact on your webform?

lalgwebdev:
@KarinG We are using "First Name AND Last Name AND (Email OR Postcode)".  The Address is however inherited from the Household Contact, and I think is not available at the point the Contact is created.  Which on reflection makes me think it's probably not doing anything for us anyway.

KarinG:
Ok, looking into this in detail now!

**Set up:**
I've created a new dedupe rule -> First Name + Last Name
Webform CiviCRM requires that Email is put on the form -> so the field is there - we just won't make it required on the form so that we can test what happens when we leave it blank.

**1. Credit Card ->** not allowed to transact at all without filling out the Email address field:
If using a Payment Processor (Credit Card) -> 
image

**2. Cash ->** -> setting Payment Processor -> to **_Pay Later_**?
image
image
image

So I do not get two Contacts if I leave Email blank -> both Contribution and Membership are attached to the same Contact record. 
image

I think this is an edge case and though I'd like to help you make this work - I'm worried about breaking something that is working as it should. It's an edge case to be using Webform CiviCRM for Cash/Cheque -> as it does not have the fields for the details for payment methods, like cheque number -> all it can do (using Pay Later) is create a Pending (Pay Later) Contribution -> which still needs to be completed afterwards in CiviCRM backend. So orgs that are collecting Cash/Cheque payments are typically doing so in CiviCRM itself. 

image

ghost:
@KarinG  Thanks for this.  We are not using Pay Later in this situation, but have a custom Payment Processor provided for us by @aydun which does record the Cash or Cheque payment (though we do not record cheque numbers). Given a recent transfer from a purely offline system about two thirds of our membership payments are currently done this way rather than online, and use of the Webform to collect all the information in one operation is important to us.  

I attach the Cash Payment Processor files for your information.

[cashpp.zip](https://github.com/colemanw/webform_civicrm/files/5288675/cashpp.zip)

KarinG:
Is there a GitHub repo for this Payment Processor Code?

I can see the use case - I’m a big fan of moving processes away from the CiviCRM UI and into Webform CiviCRM.

ghost:
https://github.com/lalgwebdev/cashpp

KarinG:
So that's forked from @aydun right? I can't find his repo though?

lalgwebdev:
I'm afraid I don't understand the details of how this all works.  I'll ask him to make contact.

aydun:
@KarinG see https://lab.civicrm.org/extensions/cashpp  See also https://github.com/colemanw/webform_civicrm/pull/292 where we made email address optional.

From a brief read of the issue, we may need another check somewhere making use of `$processor->supports('NoEmailProvided')`

KarinG:
Ok so I installed -> https://lab.civicrm.org/extensions/cashpp
(Enabled the Extension; Configured a Cash Payment Processor; enabled that Payment Processor on the Contribution Page that is linked to my Webform). And then I changed my Webform Config:
image


So I'm logged in -> and am Recording a $20 + GST Membership Payment for a new Contact: John Astridge:

image
image

Hitting Submit:
Everything works as expected: only 1 Contact is created -> and I can see a Completed Contribution and a New Membership

image

I suspect the issue may be with the Dedupe Rule you've configured?

ghost:
@KarinG What DeDupe Rule are you using?  Our rule does require the Email (or Postcode), because I found that just First Name & Last Name gave false positives when tested against our existing user base.

And is it reasonable to rely on the Dedupe Rule being configured in a particular way for this operation to work?

KarinG:
Dedupe rule -> see above^^ -> First and Last Name (1 + 1 = 2) and have configured that Rule as the Matching Rule on this specific webform.

You're absolutely correct in that just First and Last Name is not a great rule - but that's all you have for sure if you're not requiring Email on the Webform. 

So the operation (process Cash/Cheque payments via Webform CiviCRM module using @aydun 's Cash Payment Processor Extension) is working fine. It's just that your dedupe rule is configured such that you end end with two contacts because CiviCRM core is not finding a match when using your dedupe rule. 

There are scenarios where the Contribution and Membership need to land on different records so I am very hesitant to make any edits in Webform CiviCRM for something that is working as designed.

lalgwebdev:
Ok, thanks.  I need to think about that, and consequences, a bit.

aydun:
@KarinG I hear what you say about contributions and membership sometimes needing to land on different records but in those cases wouldn't the webform have separate contacts for each?  

If a webform has one contact, a membership and a contribution for that membership then (at leat to me!) the obvious expected behaviour is that the contribution and membership relate to the one specified contact.  Relying on the deduping rules seems unnecessary and error-prone - just pass the contact-id (whether created or matched) through to the contribution and membership creation.  The code analysis in the description above suggests this was the intention anyway.

KarinG:
I think relying on dedupe rules is a good thing - and it's not the dedupe rule that's error-prone - it's the misconfiguration of the dedupe rule (configured to do math on fields that are not required on the form as a result the rule finds no match and creates a new contact).

lalgwebdev:
I think my real problem is that the involvement of dedupe rules in this is totally non-obvious, so the thought of having to adapt the rule does not occur to one.  

Our scenario is actually a bit more complicated.  I have allowed for the fact that not all contacts have an email, by using a rule which says "First Name AND Last Name AND (Email OR Postcode)".  But that doesn't work either because the address is flowed down from a Household which is a second Contact, and the Address record isn't created until after the deduplication is attempted.

The only solution seems to be to use a poor rule (just the name) or introduce an extra field which is not needed by the business requirement.  In fact that's where we are at present with a Conditional and CiviRules to set a dummy email, and clear it again afterwards. Messy.

KarinG:
I'm happy to take another look but I need to be able to reproduce the issue and I still haven't been able to do so. 

**If you can screenshot me:**
Your exact custom dedupe rule (in CiviCRM -> showing the math -> length/weight/sum). 
The rule on your webform config. 
The fields are on your webform.
And anything else that is pertinent.

If you prefer to send them in private you can email this to frontdesk@semper-it.com 
A live link to a form would also be helpful.

ghost:
@KarinG Thank you.  I will collect and send details but I have another problem with my development system at the moment, and only have personal spare time for this, so it may be a day or so.

KarinG:
That's ok! I'm on personal time as well :-)

ghost:
@KarinG Please see email sent to the address given.

KarinG:
Thanks Tony - received it! I'll find some time to read it in detail (and log in to your dev/test) and will get back to you early next week (if not before).

lalgwebdev:
Hi Karin,
I should have said:  the dev2 site is running on CiviCRM 5.28.2 and Webform_Civicrm 5.1.  I have another test site which is on CiviCRM 5.29.1 and Webform_Civicrm 5.2 which shows the same symptoms – that’s where the screenshots of Pre/Post traces came from.

KarinG:
I've taken a look at your Test form -> and was able to made it work. Key is that the Rule that's configured for the webform -> First (20) + Last (20) + Email (5) + PostCode (5) and matching on 45 -> so if a form only contains First and Last we can never score 45. So if that's your rule you must present PostCode (if you want to skip email)  -> and doing that (see screenshot below): everything works fine. Even if I skip the Existing Contact - everything correctly lands in one Contact Record:

image

KarinG:
Taking a look at your Household scenario now.

KarinG:
Ah so the issue is that PostCode (which is collected on the Household) filters down to the Individual but this happens too late?

image

KarinG:
So `$this->processSharedAddresses()` occurs very late in `wf_crm_webform_postprocess.inc` - and indeed too late for the address to be included in the evaluation of the `$this->findDuplicateContact($contact);` That's a bigger issue that will also occur when Webform's are submitted without any Membership or Contribution details. Let me think about how best to solve this.

KarinG:
I touched base with @coleman and this is a chicken-egg situation -> you can't create/process the Shared Addresses until you've Created the Contacts and their Addresses. But if during the Creation of a Contact you have a dedupe Rule that contains an Address Element that is shared... So again this is not specific to Contributions or Membership.

You could add the Address to the Primary Contact (cid1). That will solve the problem with the field not existing prior to Webform CiviCRM invoking the dedupe rule to check the CiviCRM database to see if this Contact already exists (or not).

ghost:
Just a thought.  One could tighten up the conditions in my original PR by changing the insertion to something like:
```
// Do not overwrite existing Id with zero.
$existingCid = $this->verifyExistingContact($cid, $field_key);
if ($existingCid != 0 || !$this->ent['contact'][$c]['id']) {
  $this->ent['contact'][$c]['id'] = $existingCid;
}
```
This gives the id returned by 'verify' priority if it does return non-zero, which reflects the current behaviour.  It still fixes the problem of creating two contacts if dedupe fails.

ghost:
@KarinG Hi. Are you still using my Dev2 site, or can I do something else with it now?

KarinG:
Hi @maynardsmith - yes you can re-use the Dev2. 

I'm closing this as the issue really boil down to having an Individual Custom Dedupe rule with an Address Element. If this was my client I would not be patching Webform CiviCRM module -> I'd be looking at:

- add a custom field on Individual -> postcode
- add that new custom field to your Individual custom dedupe rule (or better yet make a new General rule)
- add that new custom field to your webform and configure that webform to be using the new General rule
- if you don't want admin to type postcode in twice - then you could add some javascript to copy the postcode from the Household to the Individual. The Individual will still inherit/share the complete Address after the form is submitted.

ghost:
@KarinG   Thank you.  I tried using a Custom Field and special Dedupe rule, but it didn't seem to work.  I have now tried again and it is working, so I must have made some mistake. 

I am happy that this works satisfactorily, but still feel that the involvement of the Dedupe rule is not at all obvious and could be better signposted for others somehow.

KarinG:
We have this help text/bubble ->

image

If you have additional text/wording please do let me know.

PR #319, opened 2020-06-27
Trim first_name and last_name.

KarinG:
Overview
Addresses https://civicrm.stackexchange.com/questions/35819/anyone-noticing-trailing-spaces-in-name-and-address-fields-submitted-via-profile

D7 or D8?
PR for D7 -> we should be able to cherry pick this for D8 after merge. 

Before
Contacts in database with e.g. first_name = "Just  "

After
No more trailing spaces.

Technical Details
This PR addresses the immediate issue. When we see a webform submission that contains "Just  " -> it typically also has other textfields like street_address that have trailing spaces - but that's not as critical and can be addressed later.


colemanw:
This PR might cause undefined index warnings if either of those fields are not on the webform (or not applicable because the contact is a Household or Organization).

Here's a more aggressive angle (hopefully not overly so) #321.

PR #320, opened 2020-06-29
RSE-1204: Fix civicrm file field show/hide behavior

i-grou:
Overview
This PR fixes broken show/hide behavior for civicrm files in webforms. It was working fine if field was initially visible, but has an issue if field is initially hidden (e.g. when conditional fields are used).
Issue is reproducable with bartik and with custom theme.

Before
before

After
after

Technical Details
Previously only visible field was hidden, now we will hide it anyway.


colemanw:
That makes sense, thanks. I also notice there's a test for `:visible` on line 99 as well. Wanna fix that one too?

i-grou:
Thanks @colemanw I've fixed it, could you review?

i-grou:
@colemanw are there any other questions?
What do we need to merge it?

colemanw:
Yep. Looks good to go. Thanks for the PR.

PR #321, opened 2020-06-29
Trim whitespace from user input

colemanw:
Overview
Make sure user input (first name, last name, etc. etc.) do not contain leading or trailing whitespace.

D7 or D8?
Needs port to D8

Before
Untrimmed strings from webform saved as-is

After
All user input trimmed

Comments
This might be over-aggressive but I can't think of any legitimate cases for whitepace at the beginning or end of a string in the db.


KarinG:
I like it! Will pull into client site for some life testing.

KarinG:
Merging after live testing!

PR #326, opened 2020-07-18
Add help tooltip - for Contact -Label

KarinG:
Overview
This PR uses the Webform way to create help tooltips -> e.g. 

image

D7 or D8?
Relevant to D8 only

Before
Help ? throwing JS errors -> 
```
Uncaught TypeError: D.settings is undefined
    <anonymous> http://d8civicrm.local/modules/contrib/webform_civicrm/js/webform_civicrm_admin.js?qdmz2h:583

```
After
When replaced -> no more JS error 

Technical Details
We can remove a lot of code after all ? help tooltips are replaced. I'm commenting out at this stage (rather than removing).

Comments
@jitendrapurohit - could you please have a look to see if technically this looks ok to you?

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -221,7 +221,7 @@ class wf_crm_admin_form {
     );
     $this->form['webform_civicrm'] = array('#type' => 'vertical_tabs');
     /*
-     * @todo evaluate this; inline JS is not allowed in D8.
+     * @todo evaluate this; inline JS is not allowed in D8 - KG - not needed anymore
     if (!\Drupal::state()->get('webform_civicrm_help_seen')) {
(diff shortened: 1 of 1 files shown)


colemanw:
Cool. Looks like the help function can be removed after all the help text is moved to `#description` attributes.

jitendrapurohit:
Works great. Have tested and replicated the js console error. The PR fixes the tooltip with the `#description` key 👍

jitendrapurohit:
Not sure if an additional `'#description_display' => 'tooltip'` will be required here. 

`'#description_display' => 'invisible'`  hides the `?` icon so probably it defaults to the `tooltip` value somewhere in the code.  

Since this is now working and tested, i think this can be merged.

Thanks @KarinG  :)

KarinG:
Thanks @jitendrapurohit ! 👍 

-- Karin

PR #328, opened 2020-07-25
Support computed elements across multiple pages

KarinG:
Overview
Refactors `submissionValue()` -> a more reliable way to retrieve form data between pages.

D8
D8 only

Before
Computed Twig element values don't make it onto the next page to be (in this example) included in the LineItem summary table.

image

After
JR assist -> It is more reliable to use the processed submission data. I did not realize this also works between pages -> but it does! Tested -> Computed Twig element value does make it into the LineItem summary table.

We've seen a couple of bug reports from people re: not retaining values across pages. I suspect this refactoring could very well help address these as well.

Comments
Could use some more testing! To reproduce: you must change the computed element value (in this example by changing the date of birth);

Technical

Form YAML for Computed Twig elements in example:

```
  age:
    '#type': computed_twig
    '#title': Age
    '#template': '{{ (''now''|date(''Y'')) - (data.civicrm_1_contact_1_contact_birth_date|date(''Y'')) }}'
    '#store': true
    '#ajax': true
  civicrm_1_participant_1_participant_fee_amount:
    '#type': computed_twig
    '#title': 'Participant Fee'
    '#template': '{% if data.age.__toString > 40 %}{{ 80 }}{% else %}{{ 99 }}{% endif %}'
    '#store': true
    '#ajax': true
    '#form_key': civicrm_1_participant_1_participant_fee_amount
    '#parent': civicrm_1_contact_1_fieldset_fieldset

```

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2470,8 +2470,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
 
   /**
    * Get or set a value from a webform submission
-   * During validation phase we use $this->crmValues
-   * During submission processing we use $this->submission
+   * Always use \Drupal\webform\WebformSubmissionInterface $webform_submission [more reliable according to JR]
    *
    * @param $fid
    *   Numeric webform component id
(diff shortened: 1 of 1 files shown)


petednz:
@jitendrapurohit if this needs testing can you pls add to your list
@KarinG i assume the typo in the title of this will not trickle through to anywhere that it might break things "mulitple"

KarinG:
Uh - shoot and no - there will be no long lasting effects.

sdoijad:
I can see that, this patch has 2 issues :
Save a Draft' functionality. it's throwing error when we click on the 'Save as Draft' button
    ( Error: Call to a member function getFormObject() on null in wf_crm_webform_postprocess->submissionValue() (line 2498 of /webform_civicrm/includes/wf_crm_webform_postprocess.inc. 
Issue #1 : - can be handle by conditional statement like :
 
```
if(isset($this->form_state) && !empty($this->form_state)) {
      $form_object = $this->form_state->getFormObject();
      /** @var \Drupal\webform\WebformSubmissionInterface $webform_submission */
      $webform_submission = $form_object->getEntity();
      $data = $webform_submission->getData();
    } else {
      if ($this->submission) {
        $data = $this->submission->getData();
      }
    }

```
Issue #2 :  
By this I am able to click on 'Save as Draft'. Now to test 'Save as Draft' , when I logout from system & again login. Then I am accessing the same form ( with same cid1 in url parameter ) , It should populated the selected 'Membership type' option or more specifically called Drafted Values.
All other fields ( first/last/email ) get populated but NOT the 'Membership Type' option.

KarinG:
I'm not able to reproduce the Save Draft error on my test form for this which is a Paid Event registration with Computed Twig elements. With this PR I can save Drafts (both on Page 1 and Page 2) and Computed Elements carry across multiple pages. 

I suspect your issue 2 is likely related to your issue 1 -> and appears to be specific to Membership. We can look at Membership issues in a separate PR but this PR does not cause or solve Membership issues you're experiencing.

KarinG:
PS - @jitendrapurohit - could you please have a look at this? I refactored based on Jake's recommendation here: https://www.drupal.org/project/webform/issues/3161380

jackrabbithanna:
Karin, we found this PR and applied the patch .. it solved a "Safe Draft" error we were experiencing.  +1

KarinG:
Awesome - thanks for looping back Mark!

KarinG:
Merging (tested by Skvare)!

PR #329, opened 2020-07-27
CATL-1465: Fix case contact loading

nishant-bhorodia:
Overview
When there are case contact fields on webform which do not have any contact associated with them in civi yet, then those fields currently get filled by the **case client**. 
Expected behaviour is that these fields remain empty since they do not have a contact for that case role.

D7 or D8
This PR is for D7

Before
Contact IDs are replaced by case_client ID if that case role doesn’t exist on the case yet
screenshot(1)

After
Contact IDs are not replaced by the case client and instead remain empty as they should be.
screenshot

Technical Details
`function findCaseClientForDefaultContact` is reponsible for finding case client for the case.
The above function was replacing the contact id,  for case role that doesn't exist, with case client ID, which should be 0 in case if the case role contact is not present for the case yet.
To fix that, we have added additional check for the case role and then fixing `$this->ent['contact'][$c]['id']` accordingly.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -271,7 +271,7 @@ abstract class wf_crm_webform_base {
         case 'case_roles':
           $to = $component['extra']['default_relationship_to'];
           $case = $component['extra']['default_for_case'];
-          if (empty($this->ent['contact'][$c]['id'])) {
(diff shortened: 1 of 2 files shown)


nishant-bhorodia:
Hey @colemanw 

Hope you are well!

This passed testing on our end, would it be possible to get this reviewed please?

Thanks!!

KarinG:
Hi! Could you please use the PR template in future?

PR #330, opened 2020-07-28
Tooltip refactor

jamie-tillman:
Overview
This is a an attempt to comprehensively incorporate old "civi-based" helptext into the admin form as Drupal #description array elements.  I found that the fieldsets that were dynamically build did not lend themselves to hardcoded #description elements, so I modified the old functions to return their values instead of printing them.  It seems to have taken effect globally, even on dynamically created fields and custom fields.

D7 or D8?
D8

Before
Previously, the fields all had relatively useless help icons:

image

After
Now, the old helptext shows on mouseover:

image

Technical Details
The main changes were the addition of a new `addHelpDescription()` method to the `wf_crm_admin_help` module and changing all of its "helptext" functions to return their text instead of directly printing it.  I was then able to make a single change in `wf_crm_admin_form` to switch over to using `addHelpDescription()` instead of `addHelp()`

Comments
There are a few tooltips with URLs in them in `<code>` elements.  These tend to overflow the tooltip container.  I'm hoping a solution can be found to correct that.  I was unable to figure it out.

There appears to be a potentially significant amount of dead code in this module that should be removed.  For example, I'm not sure what parts of `includes/contact_component.inc` are used, but parts of it appear to be reimplemented in `src/Plugin/WebformElement/CivicrmContact.php`

An analysis for dead code would probably help with future refactoring work, as it's very confusing to me as a newcomer to the module.

Due to this confusion, I've tried to make the smallest amount of code changes possible in this first attempt at refactoring these tooltips.  I think a cleaner approach could be identified once some clean-up and refactoring has been done.

### DIFF
--- a/includes/wf_crm_admin_help.inc
+++ b/includes/wf_crm_admin_help.inc
@@ -13,6 +13,19 @@ use Drupal\webform_civicrm\Utils;
  */
 class wf_crm_admin_help {
 
+  public static function intro() {
+    return '<p>' .
+      t('Create anything from a simple newsletter signup, to a complex multi-step registration system.') .
+      '</p><strong>' .
+      t('Getting Started:') .
+      '</strong><ul>' .
+      '<li>' . t('Enable fields for one or more contacts.') . '</li>' .
+      '<li>' . t('Arrange and configure those fields on the "Webform" tab.') . '</li>' .
+      '<li>' . t('Click the blue help icons to learn more.') . '</li>' .
+      '<li><a href="https://docs.civicrm.org/sysadmin/en/latest/integration/drupal/webform/" target="_blank">' . t('Read the instructions.') . '</a></li>' .
+      '</ul>';
+  }
+
   public static function contact_existing() {
     return '<p>' .
       t('Gives many options for how this contact can be autofilled or selected. From the Webform tab you can edit this field to configure:') .'</p><ul>' .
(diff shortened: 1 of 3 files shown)


includes/wf_crm_admin_help.inc
```diff
@@ -97,9 +99,9 @@ class wf_crm_admin_help {
 
 
   public static function contribution_contribution_page_id() {
-    print '<p>' .
+    return '<p>' .
       t("It is recommended to <a !link>create a new contribution page</a> solely for webform use. When configuring the page, most options will be irrelevant (such as profiles, premiums, widgets, recurring, etc.). Only the following need to be configured:",
-        array('!link' => 'target="_blank" href="' . url('civicrm/admin/contribute/add', array('query' => array('reset' => 1, 'action' => 'add'))) . '"')) .
+        array('!link' => 'target="_blank" href="' . Url::fromUri('internal:/civicrm/admin/contribute/add', array('query' => array('reset' => 1, 'action' => 'add')))->toString() . '"')) .
       '</p><ul>' .
```
KarinG, line 138: This is not a clickable link. Try this for help text with links in them:

```
      t('It is recommended to <a href=":link">create a new contribution page</a> solely for webform use. When configuring the page, most options will be irrelevant (such as profiles, premiums, widgets, recurring, etc.). Only the following need to be configured:',
        array(':link' => '/civicrm/admin/contribute/add')) .
```

KarinG:
OK reviewing this now!

**1. Patch applies cleanly to latest 8.x-5.x**
```
karins-MBP:webform_civicrm sysadmin$ cat 330.diff |patch -p1
patching file includes/contact_component.inc
patching file includes/wf_crm_admin_form.inc
patching file includes/wf_crm_admin_help.inc
```
**2. Visual check -> seeing very nice bright tooltips everywhere ->** 
image

**3. Code ->** I think this is a very clever way to get all them ported over quickly. 

**4. Details**
a) issue with: clickable links (see screenshot; I added a snippet during code review - that makes links clickable).
image

**also:**
b) We should remove the 3x #description I had added - so that we have everything in one place/one way of doing things. If you can get to that while refactoring the url bits - that would be great - then we contain it all in one PR. Otherwise I'm happy to do that after we get your work in. 

**5. Clean up** - I think for after we merge this? For cleanup commenting out with a ToDo note [a PR with just commented out code for others to review/test] is a great way without unintended consequences.

KarinG (changes requested):

KarinG (approved):

KarinG:
Great work @jamie-tillman - thank you! 

We'll have to put some thought into how we could do some CSS for the links (to make them more visible) - but that's for later! 

Also -> the overflow -> I think these are quite alright:
image

Merging this ahead of our upcoming beta1 release.

PR #331, opened 2020-07-29
Add CRM.payment.getTotalAmount() function (fix for PayPal / Stripe)

mattwire:
Overview
For javascript based payment processors (eg. Stripe, PayPal Checkout via Omnipay) it is difficult to retrieve the total amount to be billed from the form.  For Stripe the total amount is recalculated using an internal function - https://lab.civicrm.org/extensions/mjwshared/-/blob/master/js/crm.payment.js#L29 but this is not ideal because a change to webform could break Stripe and vice-versa. For PayPal Checkout getting the amount did not even work in many parts of CiviCRM.

From 0.8 of mjwshared I started shipping a "CRM.payment" javascript library which has been slowly evolving. From 0.9 I implemented an "override" which allows individual clients to implement (and maintain) their own `getTotalAmount()` function - https://lab.civicrm.org/extensions/mjwshared/-/blob/0.9/js/crm.payment.js#L160-163

This is just a start and I hope to further develop / improve the `CRM.payment` library and "interface" to make it easy for 3rd-parties to interact and process payments with CiviCRM.

The change specific to `webform_civicrm` here is that we separate the formatting and calculation and then expose the calculation function via `CRM.payment.getTotalAmount()`.

D7
To be forward-ported to D8 - @colemanw @KarinG I'm still not sure of the process for this?

Before
Difficult to get total amount from webform via javascript.

After
Easy to get total amount from webform via javascript. Maintained within webform_civicrm code so easy to make changes to webform_civicrm frontend without breaking payment processors.

Technical Details
See proposed implementation for omnipay (paypal_checkout). Stripe currently ships with code to get the amount from webform_civicrm but will update to defer to this once it's available via a release.

Comments

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -54,11 +54,18 @@ cj(function($) {
   });
   loadBillingBlock();
 
+  function getTotalAmount() {
+    var totalAmount = 0.0;
+    $('.line-item:visible', '#wf-crm-billing-items').each(function() {
+      totalAmount += parseFloat($(this).data('amount'));
+    });
+    return totalAmount;
+  }
+
   function tally() {
     var total = 0;
-    $('.line-item:visible', '#wf-crm-billing-items').each(function() {
-      total += parseFloat($(this).data('amount'));
-    });
+    total = getTotalAmount();
+
     $('td+td', '#wf-crm-billing-total').html(CRM.formatMoney(total));
     if (total > 0) {
       $('#billing-payment-block').show();
@@ -120,4 +127,18 @@ cj(function($) {
     });
 
   tally();
+
+  var payment = {
+    getTotalAmount: function() {
+      return getTotalAmount();
+    }
+  };
+
+  if (typeof CRM.payment === 'undefined') {
+    CRM.payment = payment;
+  }
+  else {
+    $.extend(CRM.payment, payment);
+  }
+
 });


KarinG:
Hi @mattwire - our 1stPay iATS Payment Processor is javascript based as well and I don’t recall any issues with total amount. The only issues we had to make them work with Webform CiviCRM were re: ids of state/provinces and countries. Sorry - I’m not understanding what exactly this will solve?

Re: D8WFC -> we have been cherrypicking a fair bit over.

KarinG:
You may just have to explain it - your brain is usually 5 steps ahead of where I am and I’m catching up! I do want to (as you know) do what we can to help payment processing with a variety of payment processors possible.

mattwire:
I *think* the difference here (looking at 1stPay code) is that you pre-auth the card via javascript and generate a token to use for payment - but you don't actually do anything with the amount until the form is submitted and then it is passed in the normal PHP / `doPayment()` way. With both Stripe and PayPal you actually pass the amount to the javascript flow and only submit a "token" with the form which is authorized for the specific amount that was on the form.

KarinG:
Ah yes that's it! So that's actually good - as that means if I review/test your PR and if things still work with both iATS legacy as well as iATS 1st pay - then we have pretty good coverage to ensure we're supporting a variety of types of payment processors. Checking now!

js/webform_civicrm_payment.js
```diff
@@ -54,11 +54,18 @@ cj(function($) {
...
   loadBillingBlock();
 
-  function tally() {
-    var total = 0;
+  function getTotalAmount() {
+    var totalFee = 0.0;
     $('.line-item:visible', '#wf-crm-billing-items').each(function() {
-      total += parseFloat($(this).data('amount'));
+      totalFee += parseFloat($(this).data('amount'));
     });
```
KarinG, line 61: nitpick: I think should probably be `totalAmount` (this could be a membership fee or an event fee - but it could also be a donation and certainly is not a credit card processing fee)

KarinG:
Ok reviewing/testing:

1. Patch applies cleanly to current 7.x-5.x
root@343d29f437fb:/var/www/html/sites/default/modules/webform_civicrm# cat 331.diff |patch -p1          
patching file js/webform_civicrm_payment.js
cleared all caches

2. Use test form with multiple line items: membership (w/ tax) + donation top-up:
image

3. iATS Legacy [i.e. credit card details on form] -> works great - amount transacted is correct ($58.50);

4. iATS 1st Pay [js pre-auth] -> works great - amount transacted is correct ($58.50);

5. Code -> I like `getTotalAmount` 👍 
`totalFee` -> nitpick: that I think should probably be `totalAmount ` (this could be a membership fee or an event fee - but it could also be a donation and certainly is not a credit card processing fee)

@colemanw - JS is your thing! Does that look ok to you? Functionally I'm giving it 👍

KarinG (changes requested):

js/webform_civicrm_payment.js
```diff
@@ -120,4 +127,18 @@ cj(function($) {
...
+    getTotalAmount: function() {
+      return getTotalAmount();
+    }
+  };
+
+  if (typeof CRM.payment === 'undefined') {
+    CRM.payment = payment;
+  }
+  else {
+    $.extend(CRM.payment, payment);
```
eileenmcnaughton, line 141: At this stage CRM.payment seems to be a feature of mjwshared but I think we should discuss this being a core js class (how would that look?)- this approach would allow for backward compatibility - ie does the core version have the class

But, how would migrating it from mjwshared to core look? Would there be a class if both were present?

js/webform_civicrm_payment.js
mattwire: @KarinG Resolved!

js/webform_civicrm_payment.js
mattwire: Currently `CRM.payment` in mjwshared has a bunch of functions - I'm not sure if all of them are "core-ready" or not but `getTotalAmount` is the most obvious / simple of them.  The `CRM.payment = .. else $.extend(CRM.payment` construct is nice from a transition / improvement point of view because it allows anyone to be the definer of CRM.payment, whether that be webform, CiviCRM core, mjwshared etc. The only issue becomes load order - ie. which one actually gets used - this line (https://lab.civicrm.org/extensions/mjwshared/-/blob/0.9/js/crm.payment.js#L160) handles that in mjwshared and defers to the existing `CRM.payment.getTotalAmount` function. I would expect a core implementation to do the same so that the "final client", ie webform or anything that loads CiviCRM payment processors (eg. if we had an integration with caldera forms in the future) would be able to define their own implementation of `CRM.payment.getTotalAmount()` - the spec simply being "Return a numeric formatted total amount that should be charged".

KarinG:
Thanks Matt!

PR #332, opened 2020-07-30
[WIP] - CATL-1562: Add start date to case role and record related case activity

nishant-bhorodia:
Overview
Currently when trying to add the case role to a case using webform
1. **Start date** is not recorded with the case role.
2. **Case role activity** is not recorded in CiviCRM against the case.
As part of this PR, we are trying to sort both points mentioned above.

Before
screenshot(5)

image

After
screenshot(6)

screenshot(7)

Technical Details
In `private function processCases()` the code to add case role already exists. Just modified it to do the following:
1. Add case role only if an active case role doesn't exist.
2. Add start date to the case role.
3. Also record related case role activity for the 'Assign Case Role' type.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1298,12 +1298,31 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
             if ($val && strpos($param, 'role_') === 0) {
               foreach ((array) $params['client_id'] as $client) {
                 foreach ((array) $val as $contactB) {
-                  wf_civicrm_api('relationship', 'create', [
+                  $parameters = [
                     'relationship_type_id' => substr($param, 5),
                     'contact_id_a' => $client,
                     'contact_id_b' => $contactB,
                     'case_id' => $result['id'],
-                  ]);
+                    'is_active' => TRUE,
+                  ];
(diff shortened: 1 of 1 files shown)


KarinG:
Hi! Could you please use the PR template in future?

jamienovick:
@nishant-bhorodia Also the above doesn't make it clear that this is being done to align to the way that CiviCRM core works when assigning a new case role. Can you make that clear please.

jamienovick:
@nishant-bhorodia I also think we need to go back to the drawing board on this. We want to align core and webform, but what we are doing is replicating core here. It would be much better if core and webform simply used the same single API call to create a case role and the API should then create the activity etc. Let's discuss before we go any further please.

nishant-bhorodia:
Thanks @jamienovick  @KarinG 
Holding fire on this.

KarinG:
With over 500 Drupal 8/CiviCRM installs and a D8WFC-beta1 to be out very soon (this weekend) - I’ll soon be updating the PR template that PRs are to be submitted for both D7WFC and D8WFC.

mattwire:
@nishant-bhorodia @jamienovick Please see #392 #393. Note that I think the case role activity should be "fixed" in CiviCRM core and not in webform.

PR #335, opened 2020-08-04
CATL-1564: Handle hidden (select) fields in submissions and emails

nishant-bhorodia:
Overview
If the widget type for a civicrm select options is changed to **hidden**, then when viewing the submissions, it starts to show as IDs instead of labels.
For example, if the **country** field's widget type is changed to **hidden** then in submissions instead of showing the country name (In our case, United Kingdom), it will start to show ids (in our case, 1226).

D7 or D8?
This PR is for D7

Before
**Submission**
screenshot(8)

**Email**
screenshot(10)

After
**Submission**
screenshot(9)

**Email**
screenshot(11)

Technical Details
1. Handling of hidden fields, in terms of submission/email rendering, is like textfields. Meaning, whatever value they have, it is rendered as it is in submissions or emails send by webform.
2. But in our case we are making it possible to hide select fields which as a result, started showing ids instead of labels since the structure for select list and hidden fields is different.
3. To fix this, used `function webform_civicrm_webform_submission_render_alter(&$renderable)`.
4. To check if the rendered field was select or not (before making it hidden), we are using `wf_crm_field_options` which returns options if the field has any (for select fields).
5. After detecting if the field was a select field, we are processing the field values and changing them to labels.

Comments

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -325,11 +325,39 @@ function webform_civicrm_webform_submission_load(&$submissions) {
 
 /**
  * Implements hook_webform_submission_render_alter().
- * Add display name to title while viewing a submission.
  */
-function webform_civicrm_webform_submission_render_alter(&$sub) {
-  if (!empty($sub['#submission']->civicrm['contact'][1]['display_name']) && empty($sub['#email']) && $sub['#format'] == 'html') {
-    drupal_set_title(t('Submission #!num by @name', ['!num' => $sub['#submission']->sid, '@name' => $sub['#submission']->civicrm['contact'][1]['display_name']]));
+function webform_civicrm_webform_submission_render_alter(&$renderable) {
+  // Add display name to title while viewing a submission.
+  if (!empty($renderable['#submission']->civicrm['contact'][1]['display_name']) && empty($renderable['#email']) && $renderable['#format'] == 'html') {
+    drupal_set_title(t('Submission #!num by @name', ['!num' => $renderable['#submission']->sid, '@name' => $renderable['#submission']->civicrm['contact'][1]['display_name']]));
+  }
+
+  // Show labels for hidden (select) fields when submissions are rendered.
+  foreach ($renderable['#node']->webform['components'] as $component) {
+    if ($component['type'] == 'hidden') {
+      $options = wf_crm_field_options($component, 'submission', $renderable['#node']->webform_civicrm['data']);
+      if (empty($options)) {
+        continue;
+      }
+      $parentkeys = webform_component_parent_keys($renderable['#node'], $component);
+      $value = &$renderable[array_shift($parentkeys)];
+      foreach ($parentkeys as $parentkey) {
+        $value = &$value[$parentkey];
+      }
+
+      // Handle multiple values.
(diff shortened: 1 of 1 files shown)


nishant-bhorodia:
Hey @KarinG @colemanw , 
This passed testing on our end, please can you review?
thanks!!

KarinG:
Hi @nishant-bhorodia - please see: my comment re: D8 here:
https://github.com/colemanw/webform_civicrm/pull/337#issuecomment-669929998

webform_civicrm.module
```diff
@@ -325,11 +325,39 @@ function webform_civicrm_webform_submission_load(&$submissions) {
...
+  if (!empty($renderable['#submission']->civicrm['contact'][1]['display_name']) && empty($renderable['#email']) && $renderable['#format'] == 'html') {
+    drupal_set_title(t('Submission #!num by @name', ['!num' => $renderable['#submission']->sid, '@name' => $renderable['#submission']->civicrm['contact'][1]['display_name']]));
+  }
+
+  // Show labels for hidden (select) fields when submissions are rendered.
+  foreach ($renderable['#node']->webform['components'] as $component) {
+    if ($component['type'] == 'hidden') {
+      $options = wf_crm_field_options($component, 'submission', $renderable['#node']->webform_civicrm['data']);
+      if (empty($options)) {
+        return;
```
colemanw: I think you mean 
```suggestion
        continue;
```

webform_civicrm.module
```diff
@@ -325,11 +325,39 @@ function webform_civicrm_webform_submission_load(&$submissions) {
...
+      }
+      $parentkeys = webform_component_parent_keys($renderable['#node'], $component);
+      $value = &$renderable[array_shift($parentkeys)];
+      foreach ($parentkeys as $parentkey) {
+        $value = &$value[$parentkey];
+      }
+
+      // Handle multiple values.
+      if (!empty($component['extra']['multiple'])) {
```
colemanw, line 349: Is this even possible? I don't think hidden elements support the `'multiple'` property.

webform_civicrm.module
nishant-bhorodia: Hello @colemanw 
Thank you for the review and suggestion, I will fix that.
Thanks!!

webform_civicrm.module
nishant-bhorodia: Thanks for the review @colemanw  :)

For CiviCRM select option fields, there is a setting called `multiple`. Even when the widget type is changed, the setting still stays.
But I see a failcase there, 
1. User submitted webform with hidden field that used to allow multiple values.
2. At some point the webform component is changed to single select (by chosing select option, unchecking multiple and making it hidden again)

In above case, the values won't convert to label again.
Do you think we can use the availability of `,` in hidden values to check if there is a single value or  multiple?

Thanks!

colemanw: I think this is ok as-is.

webform_civicrm.module
colemanw: Ok, waiting for your change here so we can get this merged.

nishant-bhorodia: Fixed @colemanw 
thanks :)

PR #336, opened 2020-08-05
[WIP] - Membership fee element

puresyntax71:
Adds membership fee element that depends on the selected membership type ID with the following configuration:


KarinG:
Hi @puresyntax71 - that's a lot of code! :-) Can you please break this down and explain what's not working in beta1, etc. It will make this easier to review/merge!

KarinG:
PS - did you check the example of how ParticipantFee can be calculated with native webform computed TWIG elements? https://docs.civicrm.org/webform-civicrm/en/latest/examples/eventcomtwig/

KarinG:
ping -> @puresyntax71 and @jackrabbithanna

KarinG:
We did a live demo yesterday showing how one can retype a participant fee -> into a computed twig element (a native Webform D8 element)! The same can be done for membership fee so I'm closing this as not necessary. Check out the examples I've referenced above - and if you have any questions - ping me on Mattermost (Drupal channel).

PR #337, opened 2020-08-06
CATL-1034: Allow updating case type of an existing case

nishant-bhorodia:
Overview
Updating case type for an existing case is not possible using webform civicrm. As part of this PR we are making it possible to be able to update case types for a case using webform.

D7 or D8?
This PR is for D7

Before
CATL-1034-before

After
CATL-1034-after

Also added help text:
image

Technical Details
1. In function `processCases` the field `case_type_id` was unset when the case is updated, because of which it doesn't update case type in CiviCRM.
Removed the key to prevent it from getting unset.
2. Added help text for **case type** field using `public static function case_case_type_id()`

Comments

### DIFF
--- a/includes/wf_crm_admin_help.inc
+++ b/includes/wf_crm_admin_help.inc
@@ -372,6 +372,18 @@ class wf_crm_admin_help {
       '</p>';
   }
 
+  public static function case_case_type_id() {
+    print '<p>' .
+      t('You can select the case type for your case here, or configure this to be "User Select".') .
+      '</p><p>' .
+      t('When "User Select" is chosen a new Drupal field component will be added where you can configure the behaviour of this field in more detail.') .
+      '<p></p>' .
+      t('Also note that setting this field will update the case type of any cases which are autoloaded by the form. ').
+      t('This may have undesirable consequences in some situations where, for example, certain custom fields no longer exist on ').
+      t('the new case type and could lead to data being inaccessible within the CiviCase user interface after the case type has changed.').
+      '</p>';
+  }
+
   public static function existing_case_status() {
     print '<p>' .
       t('If a matching case of the chosen type already exists for the client, it will be autofilled and updated.') .
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1278,7 +1278,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         else {
           $params['id'] = $this->ent['case'][$n]['id'];
           // These params aren't allowed in update mode
-          unset($params['creator_id'], $params['case_type_id']);
+          unset($params['creator_id']);
         }
         // Allow "automatic" status to pass-thru
         if (empty($params['status_id'])) {


nishant-bhorodia:
Can I ask why originally it was not allowed to update case types please?
Thanks!

KarinG:
@nishant-bhorodia : with D8 Webform CiviCRM in beta now, PRs are going to need to be ported to D8. Are you able to commit to that? When we get to D8WFC stable [this Fall] - I'll be asking that PRs for new features like this are to be against D8WFC and backported to D7WFC. 

I don't know much about the history of Case in Webform so we'll wait for @colemanw to answer your question.

KarinG:
PS - this of course would be a very simple port - but my point is bigger picture -> as all of Compucorp's PRs are still against D7 only at this time. Would really like to have your team on board on D8!

jamienovick:
Hi @KarinG @colemanw just to reply on this.

We're on a bit of a timeline for getting these PR's merged if at all possible since they fix open bugs for a major client. Appreciate you also want things on D8 but at the moment we don't have any infrastructure for working and testing Drupal 8.

Could we perhaps have a chat later in the week about ways or working with D7/8 PR's and agree approach so we know what we need to do?

Best

J

KarinG:
Hi Jamie - what would really help is if we have someone using Case who can confirm/test both in D7 and D8. This specific edit seems fine and would be easy for me to forward port to D8,  but I don't understand the functional implications.  We do not use Case at all with any clients ourselves. We do all Case management with Activities/Views & some custom glue to create Case Manager portals outside of CiviCRM, but in general allowing more flexibility to update I support. I'll ping Oliver Gibson to see if he can review/ok this change.

jamienovick:
Hi @KarinG of course. We have QA'd internally by a separate QA if that helps at all... ps. if there is any chance of getting this and the other 2 PRs: 
https://github.com/colemanw/webform_civicrm/pull/335
https://github.com/colemanw/webform_civicrm/pull/329
approved by Wed I would be eternally grateful... not sure if thats feasible?

gibsonoliver:
Hi Karin/Jamie, I think the use case is fine. But what happens if the original case type has associated custom fields which aren't associated with the new case type. I guess that will lead to case data loss. Maybe there could be a warning about this on the option help text?

jamienovick:
@nishant-bhorodia  see comment above - can we get this updated first thing tomorrow? Thanks

colemanw:
@jamienovick this never got updated with help text, which is what it's waiting for.

jamienovick:
@colemanw sorry this didn't get picked up.

Proposed text:

Add a blue (?) with the help text in bubble:

"You can select the case type for your case here, or configure this to be "user select". When "user select" is chosen a new Drupal field component will be added where you can configure the behaviour of this field in more detail. Also note that setting this field will update the case type of any cases which are autoloaded by the form. This may have undesirable consequences in some situations where, for example, certain custom fields no longer exist on the new case type and could lead to data being inaccessible within the CiviCase user interface after the case type has changed."

@nishant-bhorodia will submit a new PR very shortly.

nishant-bhorodia:
Thanks @colemanw @jamienovick 
I have added the help text and updated PR description as well with the screenshot.
Thanks!

colemanw:
Good to go, thanks.

PR #342, opened 2020-08-18
Use entity functions to save custom data

mattwire:
Overview
Currently custom data for entities is saved separately, after saving the main entity via the entity API3. Saving custom data does not use the API.

This is probably because we didn't used to use the API to save entities and you couldn't save the custom data at the same time. But API3 has supported saving custom data for a long time.  The biggest issue with saving separately is that it makes it very hard to interact and automate using hooks / civirules etc.

This PR changes that, removes a big chunk of custom code and switches saving of entity custom data to use the main entity API save function.

D7 or D8?
D7, I'll port to D8 once we've agreed and merged this.

Before
Custom data saved separately to entities. Not using API.

After
Minimal special handling for custom data. Saved with main entity API.

This means that, for example, you can trigger Civirules on Membership.create and you'll be able to "see" all the custom data that was saved with it.

Technical Details
There is a small block of code that manipulates the custom data values into the right parameter format for API3. I believe this works in all circumstances but it's possible there are edge cases I've not found (eg. saving files?).

Comments
@KarinG Per discussion. @colemanw


KarinG:
That looks great @mattwire - did you say you have this live on some sites already? Sorry I don't recall.

mattwire:
@KarinG Yes, it's on two live sites. They work exclusively with cases, activities and contacts.

KarinG:
Excellent 👍 I'll also push this live onto a site or two and will run a couple of scenarios to confirm, specifically: contributions and membership custom data.

colemanw:
I think the reason it used to be this way is because the api historically didn't support saving custom data and when it began to add support it was done in a piecemeal/inconsistent way. If apiv3 is now rock-solid in this area then I agree we can switch.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2296,6 +2281,19 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
+          if (!empty($component['extra']['items'])) {
+            $items = explode("\n", $component['extra']['items']);
+            $customValue = [];
+            foreach ($items as $item) {
+              list($itemName, $itemLabel) = explode('|', $item);
+              if (in_array($itemName, $val)) {
+                $customValue[$itemName] = 1;
+              }
+            }
+          }
```
colemanw: There's a function to do this: `wf_crm_webform_base::getExposedOptions()`

mattwire: Thanks @colemanw I've switched it over to that function

mattwire:
@KarinG @colemanw Just to let you know I've done a bit more testing and pushed an update which switches to the function that @colemanw pointed out.

KarinG:
Will check it out this weekend! Thanks @mattwire 👍

colemanw:
Code looks good to me now. Haven't tested.

KarinG:
Testing this now!

KarinG:
**Documenting testing:**

image
image
image

**Checking variety of custom fields:**

**Participant**
Session 1 - Arts (select)

**Contribution**
Question - Yes (radio)

**Contact**
Newsletter - No (radio)
File upload - ok (file)

**Membership** 
Online copy - Yes (radio)

**Activity**
Longitude - 1001 (text)
File upload - ok (file)

**Grant**
Cookie Preference - chocolate (select)

**Also:** double checked flattening (dragged 4 of these custom fields out of their fieldsets) - all good

**dblog: [unrelated]**
pr-342-custom-fields
Deprecated  function CRM_Core_BAO_Cache::getItem, use CRM_Core_BAO_Cache::getItem  is deprecated and will be removed from core soon, use Civi::cache()  facade or define cache group using hook_civicrm_container. Array (     [civi.tag] => deprecated )

KarinG:
I like the clean up! 👍 

I really like that this allows folks to create even more functionality by being able to use hooks/civirules to automate actions! 👍 👍 

I've tested a variety of custom fields (incl File) - and it's all working :-) so merging this!

mattwire:
Cool! Thankyou. I'll do a d8 port in the next week or so!

colemanw:
Because of the regressions in #364, I'm reverting this and releasing 7.x-5.3 immediately. Hopefully @mattwire can fix the two bugs and then we can un-revert.

PR #343, opened 2020-08-18
[WIP] - Decouple list of payment processors from contribution page.

mattwire:
Overview
Allow to select any active processor. Remove user_select on mode (live/test) and make sure that webform_civicrm always sends the correct (live or test) payment processor ID to CiviCRM.

This is work in progress per our discussion @KarinG.  `webform_civicrm` only actually uses about 4 parameters from a contribution_page and removing that dependency would allow for some pretty cool stuff like currency selection on the form etc.

D7 or D8?
D7, will port once finalised/agreed.

Before
Payment processor selection dependent on contribution page.

After
Payment processor selection not dependent on contribution page.

Technical Details
There are some specific complexities and issues here that we need some discussion on:
1. Test/Live processors. We *should* be passing the correct ID for the live/test processor when submitting the webform. However for multiple processors this isn't currently possible because the IDs are saved in the admin screen.
- Proposal: Remove "User select" for "Test mode/Live mode" and require it to be one or the other on the admin screen.

2. "User select" and multiple selections. These seem complex to implement at the moment and I'm trying to add some metadata to allow that. But it would be good to get some thoughts from the webform experts :-).
Using the (simpler) example of currency which is not in this PR:
a) You may select a single currency. This is easy and works with the admin UI we have.
b) You may select multiple currencies. This requires that a multiselect *without* "user select" is shown in the admin UI. Then a multi-select showing only the selected options is displayed on the form. I don't see a need to expose an "any" option in the admin UI though we could.
- You'll see in this PR I've added `'disable_user_select' => TRUE, 'multiple' => TRUE,` as admin metadata and that may be the right thing to do but I'd like agreement/alternatives. You'll also see it requires tweaks at various points in the code to make it work.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -75,6 +75,38 @@ function wf_crm_field_options($field, $context, $data) {
     elseif ($table == 'membership' && $name == 'num_terms') {
       $ret = drupal_map_assoc(range(1, 9));
     }
+    elseif ($table === 'contribution' && $name === 'payment_processor_id') {
+      // For the config form we display a list of all active (live) payment processors
+      // Saving will map the IDs to live or test.
+      // For the frontend we display the selected payment processors (with correct ID for live or test)
+      $params = wf_crm_aval($data, "$ent:$c:$table:$n", []);
+      if ($context === 'config_form') {
+        $paymentProcessors = wf_crm_apivalues('PaymentProcessor', 'get', ['is_test' => $params['is_test'], 'is_active' => 1]);
+        $paymentProcessors[0]['name'] = $field['exposed_empty_option'];
+        foreach ($paymentProcessors as $paymentProcessorID => $paymentProcessor) {
+          $ret[$paymentProcessorID] = $paymentProcessor['title'] ?? $paymentProcessor['name'];
+        }
+        return $ret;
+      }
+      else {
+        $selectedPaymentProcessors = wf_crm_aval($params, "payment_processor_id", []);
+        if (in_array(0, $selectedPaymentProcessors)) {
+          $ret[0] = $field['exposed_empty_option'];
+        }
+        $paymentProcessors = wf_crm_apivalues('PaymentProcessor', 'get', []);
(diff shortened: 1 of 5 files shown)


KarinG:
Kids are wondering why I'm dancing reading this :-)

KarinG:
Closed in favour of: https://github.com/colemanw/webform_civicrm/pull/366

KarinG:
The new Payment Processor Element -> on my iATS test site :-) They are all there!

image
 
**Ok the ToDo list:**
1) make a - User Select - option for Payment Processor
prepend or append "-test" (or something like that) behind all Test payment Processors Labels.
2) remove Payment Processor Mode element -> we don't need that anymore once 1) is a - User Select - and that also eliminates the problem of selecting the correct Payment Processor. 

1 and 2 will give Admins the ability to completely configure any Payment Processor - Live or Test and will put them onto the Webform in any Order and with any modified Label should they choose to do so.

**Requires some thought:**
3) we'll need to put some thought into putting Currency onto the form - as some payment processors should only be configured with one specific Currency... perhaps with a big warning help text?

**To discuss:**
4) Email receipt -> that's also a configuration that's currently taken from the Contribution Page that is selected. How shall we handle that? 

Any other variables?

mattwire:
> Requires some thought:
> 3) we'll need to put some thought into putting Currency onto the form - as some payment processors should only be configured with one specific Currency... perhaps with a big warning help text?

We don't handle this in core - ie. you can select the "wrong" currency and it will just fail somewhere during submission. I'd say just a suitable help text is fine. Most users configuring payment processors probably already know what currencies they are accepting.

> To discuss:
> 4) Email receipt -> that's also a configuration that's currently taken from the Contribution Page that is selected. How shall we handle that?

It's "just" another tickbox on the webform config I think?

KarinG:
Closing this in favour of:
D8 -> https://github.com/colemanw/webform_civicrm/pull/369
D7 -> WIP -> https://github.com/colemanw/webform_civicrm/pull/368

PR #346, opened 2020-08-25
Fix for Missing options for Contact Type, Relationship types etc.. on build Edit field section

sunilpawar:
Fix for Missing options for Contact Type, Relationship types etc.. on build Edit field section
Overview
When CiviCRM Webform field like Contact Sub Type, Relationship Types are exposed as `User Select` option, then under build tab ->  edit field -> field Options for these field missing. Modified any setting in Edit mode, loose all options by default available in view mode.

Before
https://www.drupal.org/project/webform_civicrm/issues/3152999
https://www.drupal.org/project/webform_civicrm/issues/3153274

After
Options are available.

### DIFF
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -50,13 +50,23 @@ class CivicrmSelectOptions extends FormElement {
     }
   }
 
-  protected static function getFieldOptions($form_key) {
+  protected static function getFieldOptions($form_key, $webform_id = NULL) {
     \Drupal::getContainer()->get('civicrm')->initialize();
     $field_options = \Drupal::service('webform_civicrm.field_options');
-    return $field_options->get(['form_key' => $form_key], 'create', []);
+    $data = [];
+    if (!empty($webform_id)) {
+      $webform = \Drupal::entityTypeManager()->getStorage('webform')->load($webform_id);
+      $handler_collection = $webform->getHandlers('webform_civicrm');
+      $handler = $handler_collection->get('webform_civicrm');
+      $settings = $handler->getConfiguration()['settings'];
+      $data = $settings['data'];
+    }
+    return $field_options->get(['form_key' => $form_key], 'create', $data);
   }
 
   public static function processSelectOptions(&$element, FormStateInterface $form_state, &$complete_form) {
+    $build_info = $form_state->getBuildInfo();
+    $webform_id = !empty($build_info['args'][0]) ? $build_info['args'][0]->id() : NULL;
     $element['#tree'] = TRUE;
     // Add validate callback that extracts the associative array of options.
     $element += ['#element_validate' => []];
@@ -109,7 +119,7 @@ class CivicrmSelectOptions extends FormElement {
 
     $current_options = $element['#default_value'];
     $weight = 0;
-    $field_options = static::getFieldOptions($element['#form_key']);
+    $field_options = static::getFieldOptions($element['#form_key'], $webform_id);
 
     // Sort the field options by the current options.
     if (!$element['#live_options']) {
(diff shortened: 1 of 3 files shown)


KarinG:
@sunilpawar - could you please break this PR up into smaller pieces - clearly explaining what/why and w/ screenshots? It will make this a lot easier to review/mergeable.

sunilpawar:
> @sunilpawar - could you please break this PR up into smaller pieces - clearly explaining what/why and w/ screenshots? It will make this a lot easier to review/mergeable.

original ticket : https://www.drupal.org/project/webform_civicrm/issues/3152999   , issue explained there. Will add more comment to explain changes.

KarinG:
You're referencing two tickets on d.o.; there is a comments that say M.H. added this. I'd encourage you to break this down if you want this reviewed quickly.

src/Plugin/WebformElement/CivicrmSelect.php
```diff
@@ -105,6 +105,14 @@ public function form(array $form, FormStateInterface $form_state) {
       '#prefix' => '<div id="webform-civicrm-options-wrapper">',
       '#suffix' => '</div>',
     ];
+    // M.H added this, provide empty option for CiviCRM Select
```
jackrabbithanna, line 108: This bit is separate from the rest of the code for this PR. This portion was to enable having an "Empty" option for elements that use this CivicrmSelect plugin, in the case that inspired this, Membership type. We needed a way to have an option in the select list that was '-Select-' .. so that when the form loads no membership type would be set as default, and we make the element required, forcing the user to make a choice for which membership type they want to purchase.

deepak-srivastava:
Options being checked by default doesn't seem to be working or addressed in the patch. For e.g If logged in contact already has a subtype (Staff), form field won't populate with subtype (Staff) checked.

src/Element/CivicrmSelectOptions.php
```diff
@@ -50,13 +50,23 @@ public static function valueCallback(&$element, $input, FormStateInterface $form
...
-    return $field_options->get(['form_key' => $form_key], 'create', []);
+    $data = [];
+    if (!empty($webform_id)) {
+      $webform = \Drupal::entityTypeManager()->getStorage('webform')->load($webform_id);
+      $handler_collection = $webform->getHandlers('webform_civicrm');
+      $handler = $handler_collection->get('webform_civicrm');
+      $settings = $handler->getConfiguration()['settings'];
+      $data = $settings['data'];
+    }
+    return $field_options->get(['form_key' => $form_key], 'create', $data);
```
sunilpawar, line 64: Main fix in PR is Passing missing `$data` element ,  if we look at D7 code : https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_admin_form.inc#L107
```
  private function initializeForm() {
    $this->settings = isset($this->node->webform_civicrm) ? $this->node->webform_civicrm : $this->defaultSettings();
    $this->data = &$this->settings['data'];
  }
```
same `$this->data` variable used everywhere to populate option value for any field.

We don't have that initialisation in  D8 in these options files , handler object is missing, What we doing is here , getting webform id and pull data from settings and pass it  `get` to populate the option as per form key.

KarinG:
@sunilpawar - again this PR is not mergeable as it stands. You can help by breaking this down into **one issue per PR** and clearly documenting before/after - is this a D8WFC parity issue; or a new feature; add before/after or D7WFC vs D8WFC screenshots in the PR. Of course you’ll need to also remove Mark’s initials in the breakout PR dealing with the empty option. Clearly show/compare D7WFC and how it handled the empty option.

KarinG:
We solve the Missing Options for Relationship Type -> here -> #475 - so I'm closing this PR.

chaplincircle:
If I've read this correctly,

https://www.drupal.org/project/webform_civicrm/issues/3152999

still remains unpatched?

It looks like it's still an issue in version 6.2.0.
@sunilpawar 
@KarinG

PR #349, opened 2020-08-31
When updating a Case with multiple contacts, unset contact ID fields to avoid API error

AsylumSeekersCentre:
When updating a Case, unset contact_id and client_id if specified and
there's more than one Contact on the Case. Otherwise we get a Case.create
error "Case is linked with more than one contact id".

Overview
The webform post-processing code would previously fail with an API error when updating a Case with multiple Contacts attached. This change prevents that error from occurring.

D7 or D8?
D7. I don't currently have a D8 Civi installation for testing.

Before
The webform would appear to submit correctly from the user's perspective, but the custom fields would not update. In the Drupal log I would find this message:
`The CiviCRM "case create" API returned the error: "Case is linked with more than one contact id. Provide the required params orig_contact_id to be replaced" when called by function "postSave" on line 219 of wf_crm_webform_postprocess.inc with parameters: "Array ( [client_id] => Array ( [0] => 34231 ) [medium_id] => 9 [id] => 20342 [check_permissions] => [version] => 3 ) "`

After
The custom fields are updated correctly and there are no errors in the Drupal log.

Technical Details
If a webform were designed to update the Contacts on a Case - where there were already multiple Contacts on that Case - I expect this change would prevent such a webform from doing its job. However, I believe that webform would also fail without this change (due to the error described above).

If there's only one Contact on the Case, no fields are unset, so it should not break any functionality which currently works.

Comments
I'll apply this change to our production server tonight. It has been tested locally but will see more intensive testing over the course of the next week.

I wasn't sure about the correct procedure for updating this old fork of the repository. If I shouldn't have created the extra branch for this commit or if there are any other issues, let me know and I can try again.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1268,6 +1268,18 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $params['id'] = $this->ent['case'][$n]['id'];
           // These params aren't allowed in update mode
           unset($params['creator_id']);
+
+          // If orig_contact_id is missing and there's more than one Contact
+          // on the Case, set orig_contact_id to the current ID. Otherwise
+          // we get a Case.create error "Case is linked with more than one 
+          // contact id".
+          if (empty($params['orig_contact_id']) && !empty($params['id'])) {
+            $caseContactCount = wf_civicrm_api('CaseContact', 'getcount', ['case_id' => $params['id']]);
+            if ($caseContactCount > 1 && !empty($params['client_id'])) {
+              $params['orig_contact_id'] = current((array) $params['client_id']);
+            }
+          }
+
         }
         // Allow "automatic" status to pass-thru
         if (empty($params['status_id'])) {


AsylumSeekersCentre:
This is an updated variation on #156 .

colemanw:
@AsylumSeekersCentre thanks for this PR. The patch looks clean. Haven't tested it but one note on code-review.
We are trying not to use internal functions from CiviCRM (such as `CRM_Case_BAO_Case::retrieveContactIdsByCaseId`) but instead go through the API.
I think the api equivalent for counting the number of clients in a case would be something like:
```php
wf_civicrm_api('CaseContact', 'getcount', ['case_id' => $params['id']]);
```

demeritcowboy:
Thanks for the PR. I was asked to give it a looksee and here's my comments:

* General standards
    * [`r-explain`] __PASS__
    * [`r-user`] __PASS__
    * [`r-doc`] __PASS__
    * [`r-run`] __PASS__
      * I'm not a heavy user of webform so really only tested this one path, and then also the same path with a "normal" setup where it only allows one case client. It worked.
      * I couldn't figure out how to set up a webform as you mention in your caveat, where you'd be adding/removing clients, so I didn't test that.
      * It might be a separate issue but since it didn't seem to function quite as in the help bubble beside "Update Existing Case" where it says it will preload and upload the existing matching case if you just pass in the contact id in the url, I couldn't test that fully. It sort of loads it but not for the custom field. Similarly if you just pass in the case id it either doesn't preload as expected or it's maybe just my unfamiliarity with it. So I only tested with both cid and case in the url.
* Developer standards
    * [`r-tech`] __Comment__
      * The core api itself suggests instead setting `orig_contact_id`, and this was also suggested in the drupal issue comments. I couldn't see anywhere in comments there or here or the previous PR what the advantage of unsetting client/contact vs setting orig_contact_id is. It seems like setting orig_contact_id might be preferred since then you wouldn't lose the client for whatever else it might be used for. So I also tested a variation of that and it seemed to also work.
         ```
         if (empty($params['orig_contact_id']) && !empty($params['id'])) {
           $caseContactCount = wf_civicrm_api('CaseContact', 'getcount', ['case_id' => $params['id']]);
           if ($caseContactCount > 1 && !empty($params['client_id'])) {
             $params['orig_contact_id'] = current((array) $params['client_id']);
           }
         }
         ```
    * [`r-code`] __Possible Issue__
      * As already noted by @colemanw can use `$caseContactCount = wf_civicrm_api('CaseContact', 'getcount', ['case_id' => $params['id']]);` and then compare `if ($caseContactCount > 1)` instead of the call to the internal function retrieveContactIdsByCaseId and then count(). I also tested this.
    * [`r-maint`] __?__
    * [`r-test`] __?__

KarinG:
Thanks @demeritcowboy for your review! @AsylumSeekersCentre -> if you can please make the the updates mentioned above then we can get this merged.

AsylumSeekersCentre:
I've applied this change and tested it on a non-production server. It works correctly in my tests. I've updated the comment above the change.

This is much closer to the change suggested in #156, but that variant didn't work when I tested recently. It looks like I was setting the orig_contact_id incorrectly, so thank you for showing me the correct way to do this.

I've updated this branch with a second commit, but please let me know if it would be better to make this pull request into a single commit somehow.

KarinG:
Excellent! Yes we would want you to squash that commit. It would probably look something like:

**On you local:**
```
git reset HEAD~2
git add wf_crm_webform_postprocess.inc
git commit -m "When updating a Case, unset contact_id and client_id if specified"
git push --force AsylumSeekersCentre fixUpdateCaseCustomField
```

There are ways to do this in fewer steps but these have always made sense to me :-)

AsylumSeekersCentre:
Great, I've followed the steps you suggested and it appears to have worked. Please let me know if there's anything else I should do.

KarinG:
Excellent 👍  I've got some time blocked off for tomorrow to test/merge D7 and D8 PRs. This one is top of the list (the 'easy' pile).

colemanw:
Thanks @KarinG and @demeritcowboy for reviewing and merging, and thanks @AsylumSeekersCentre for your responsiveness to feedback.
This is how we get stuff done!!!

AsylumSeekersCentre:
Thank you for your feedback and for merging the change. :)

KarinG:
@AsylumSeekersCentre - do check out our https://webform-civicrm.io/ -> it has monthly Newletters, monthly Lunch & Learn sessions (Free) and Office Hours options.

PR #350, opened 2020-09-01
Fix autoload of event from URL

jitendrapurohit:
Overview
Fix autoload of event from URL params.

D7 or D8?
D8WFC

Before
If the below setting is enabled on the event tab for the contact, the event cannot be loaded from the URL params eg c1event1=<event_id>

image

To replicate -

- Enable the above setting from `Event` tab -> `Registration option` section on the civicrm setting page.
- Try to load event id from the URL
- Event is not loaded on the field.

After
Event loaded from the URL Eg - `...?c1event1=2`

Technical Details
Handling for d8 element civicrm_options.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -531,6 +531,9 @@ abstract class wf_crm_webform_base {
       return $exposed;
     }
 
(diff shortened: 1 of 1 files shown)


KarinG:
@jitendrapurohit - thank you for this!
That's one more step to feature parity w/ D7WFC.

PR #351, opened 2020-09-05
Fix tag set retrieval

colemanw:
Overview
I think this will fix a potential infinite loop where tags that do not belong get fetched.


KarinG:
Yep this works fine!

colemanw:
Thanks @KarinG - I cherry-picked this into the 8x branch as well

PR #352, opened 2020-09-07
D7 : Do not change Membership type for Inherited membership record

sunilpawar:
Overview
Do not change Membership Type for Inherited Membership Record.

Steps to Recreate Issue

1. Assume you as individual have Inherited Membership record from your employer. ( you don't have other membership record).
2. Now, As individual signup for any membership type (excluding existing Inherited membership type, just to avoid confusion), when signup is completed, your existing Inherited Membership type get changed to New Membership type. If you click on View membership in Backend, you will see Error Message. 

D7 or D8?
This issue exist in D7 and D8 version of this module.

Before
Inherited Membership Type changes to other Membership record.

After
Inherited Membership Type record get skipped, Either New Membership is created or other membership record (which are not from  Inheritance) type get changed.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -586,6 +586,8 @@ abstract class wf_crm_webform_base {
       'contact_id' => $cid,
       // Limit to only enabled membership types
       'membership_type_id' => ['IN' => $membership_types],
+      // skip membership through Inheritance.
+      'owner_membership_id' => ['IS NULL' => 1],
     ]);
     if (!$existing) {
       return [];


KarinG:
@sunilpawar - thank you - and also thank you for the port.

I'm reproducing / testing this now.

KarinG:
"If you click on View membership in Backend, you will see Error Message." -> what Error message is that? 

Reproducing this ->

**1. Mom Purchasing a Family Membership ->**
image

**2. The Child successfully inherits it:**
image

**3. Child purchasing his own (and different) Membership Type ->** 
image

**4. The Child now has two Membership Types:**
image

In my example -> the two Membership Types belong to a different (sub)Organization. CiviCRM will want to try and upsell/downsell Membership belonging to the same Organizations. 

**Conclusion ->** what webform does is in line with what CiviCRM native does. So if you want the Individual to be able to purchase/hold their own Membership _**in addition to**_ the Membership they inherit from their Company -> then you must configure the two Membership Types to belong to two different organizations.

KarinG:
PS - if you can demonstrate that CiviCRM native (in backend eg) allows a Contact to have both and inherited Membership as well as their own Membership (even when both Membership types belong to the same Organization in the Membership Config) - then we can revisit this. My main goal here is to have consistent behaviour w/ Core when it comes to Membership re-tying.

sunilpawar:
1. Existing Membership Record through Inheritance..
Screenshot 2020-09-07 at 10 41 31 PM

2. Submitting form for different Membership type.

Screenshot 2020-09-07 at 10 45 04 PM

3. After Submitting Form: Inherited Membership type change to `General`
Screenshot 2020-09-07 at 10 46 47 PM


I am Not sure about why i got SQL Error first time only, the error got after clicking on View link in-front of Inherited Membership record after changing the membership type. (not reproducible now for me in 2nd attempt) 

> SELECT relationship_type_id,
>   CASE
>   WHEN  contact_id_a = 1 AND contact_id_b = 205 THEN 'b_a'
>   WHEN  contact_id_b = 1 AND contact_id_a = 205 THEN 'a_b'
> END AS 'relType'
>   FROM civicrm_relationship
>  WHERE relationship_type_id IN () [nativecode=1064 ** You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 7]"]
>   BACKTRACE :
>  PATH/drupal-7.72_testing/sites/all/modules/civicrm/CRM/Member/Form/MembershipView.php(214): CRM_Core_DAO::executeQuery("\nSELECT relationship_type_id,\n  CASE\n  WHEN  contact_id_a = 1 AND contact_...")
>  PATH/drupal-7.72_testing/sites/all/modules/civicrm/CRM/Core/Form.php(602): CRM_Member_Form_MembershipView->preProcess()
>  PATH/drupal-7.72_testing/sites/all/modules/civicrm/CRM/Core/QuickForm/Action/Display.php(76): CRM_Core_Form->buildForm()
> PATH/drupal-7.72_testing/sites/all/modules/civicrm/packages/HTML/QuickForm/Controller.php(203): CRM_Core_QuickForm_Action_Display->perform(Object(CRM_Member_Form_MembershipView), "display")

KarinG:
Do `Senier` and `General` belong to the same Membership organization [ in 
civicrm/admin/member/membershipType/add?action=add&reset=1 ]

sunilpawar:
> Do `Senier` and `General` belong to the same Membership organization [ in
> civicrm/admin/member/membershipType/add?action=add&reset=1 ]

yes, both belong to same Membership organisation

KarinG:
So let's try this CiviCRM native -> Public Contribution Pages

Two Memberships Types: Family - Magazines and Individual- Magazines - both are belonging to the same Organization:

**1. Mom takes out a Family - Magazines Membership** -> Justin correctly inherits it.
image

**2. Justin is now old enough and wants to get his own Individual- Magazines** 
So I'm putting these two Membership Types on a public contribution page:

image

**Justin's Membership got re-typed because it belongs to the same Membership Organization:**
image

**But I did get the SQL error on View ->** 
`$Fatal Error Details = Array ( [callback] => Array ( [0] => CRM_Core_Error [1] => handle ) [code] => -2 [message] => DB Error: syntax error [mode] => 16 [debug_info] => SELECT relationship_type_id, CASE WHEN contact_id_a = 2 AND contact_id_b = 188 THEN 'b_a' WHEN contact_id_b = 2 AND contact_id_a = 188 THEN 'a_b' END AS 'relType' FROM civicrm_relationship WHERE relationship_type_id IN () [nativecode=1064 ** You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 7] [type] => DB_Error [user_info] => SELECT relationship_type_id, CASE WHEN contact_id_a = 2 AND contact_id_b = 188 THEN 'b_a' WHEN contact_id_b = 2 AND contact_id_a = 188 THEN 'a_b' END AS 'relType' FROM civicrm_relationship WHERE relationship_type_id IN () [nativecode=1064 ** You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 7] [to_string] => [db_error: message="DB Error: syntax error" code=-2 mode=callback callback=CRM_Core_Error::handle prefix="" info=" SELECT relationship_type_id, CASE WHEN contact_id_a = 2 AND contact_id_b = 188 THEN 'b_a' WHEN contact_id_b = 2 AND contact_id_a = 188 THEN 'a_b' END AS 'relType' FROM civicrm_relationship WHERE relationship_type_id IN () [nativecode=1064 ** You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 7]"] ) `

So Webform CiviCRM behaves exactly the same as CiviCRM native Membership retyping does. 

However CiviCRM native clearly does not handle this scenario well. And the goal of matching functionality is not to match a bug. This is producing inconsistent data -> to 'inherit' a Membership of a different Type. 

Confirming this PR produces this result when the Child purchases their own Membership of the same organization:

image

Ok - I'm inclined to merge this. I think it's a good improvement to avoid inconsistent data. 

@sunilpawar you please put in an issue for CiviCRM Core? Consistent behaviour should be our goal.

PR #359, opened 2020-09-12
Webform 6.x compatibility

homotechsual:
Overview
_Make plugin definition compatible with 5.x and 9.x on Drupal 8 and 9._

D7 or D8?
_D8WFC backport not possible._

### DIFF
--- a/src/Plugin/WebformHandler/CivicrmWebformHandler.php
+++ b/src/Plugin/WebformHandler/CivicrmWebformHandler.php
@@ -43,30 +43,11 @@ class CivicrmWebformHandler extends WebformHandlerBase {
   protected $civicrm;
 
   /**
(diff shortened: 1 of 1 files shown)


jitendrapurohit:
Have tested the changes with webform 6.x. 

**Before** - Immediate problem observed on webform 6.x -

No Contact Tab appearing on the initial load of settings form.

Screenshot 2020-09-20 at 6 50 44 PM

**After** the patch - Contact tab loading fine.

Screenshot 2020-09-20 at 6 51 26 PM

Have also tested quick webform creation, submitting values, etc, and found the plugin to be usable with the latest changes.

Performed the same testing with latest 5.x webform. Works fine 👍. Thanks @MikeyMJCO

PR #361, opened 2020-09-23
D8: Issue with translations and creating groups

kewljuice:
Overview
Can't save webform with group(s) when the word webform is translated in the Drupal translation interface.

D7 or D8?
D8WFC

Before
Can't save webform with group(s).

After
Can save webform with group(s).

Technical Details
The $params['method'] gets sets to webform or its translation, in de database only 8 chars are allowed for the method field an once the word 'webform' gets translated (example) into dutch it becomes 'Webformulier' which is longer then 8 chars and an error is thrown.

Comments
Patch contains tempory fix, there could be a better solution!

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -906,7 +906,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
     switch ($data_type) {
       case 'group':
         $api = 'group_contact';
-        $params['method'] = t('Webform');
+        $params['method'] = substr(t('Webform'), 0, 8);
         break;
       case 'tag':
         $api = 'EntityTag';


KarinG:
Hoi Wouter! Thank you for the patch. 

I'm going to merge this in - happy to revisit this when there is a better solution but this way "Webformu" will not throw any more errors for you. 

Waar woon/werk jij? :-)

kewljuice:
Thanks!

There is indeed more discussion here, should 'webform' be translated here?
If yes, the field settings for 'method' should probably be adjusted in civicrm-core.

Ik woon in Stekene/België, en werk bij Calibrate (calibrate.be) in Antwerpen/België.

KarinG:
Yeah - agreed! 

Ik woon in Calgary, Canada nu maar ben geboren in Gendt (in Nederland) -> wat is de naam van de stripfiguur in je Github image? I herken die figuur van vroeger :-)

kewljuice:
De naam van die stripfiguur is Guust Flater :-)

KarinG:
Ja natuurlijk! 👍 Nu weet ik het weer. Het is alweer een tijdje geleden.

Leuk dat jullie ook Webform CiviCRM doen!

PR #362, opened 2020-09-26
Migrate help inc file to services

jitendrapurohit:
Overview
Update wf_crm_admin_help.inc file to use services.

D7 or D8?
D8

Before
.inc file to retrieve all help text for civi elements.

After
- New d8 service added `webform_civicrm.admin_help`
- Old code removed which depends on ajax notification popup and is no longer required.

Comments
ping @KarinG

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -11,13 +11,13 @@ use Drupal\webform_civicrm\Utils;
 use Drupal\node\Entity\Node;
 
 \Drupal::ModuleHandler()->loadInclude('webform_civicrm', 'inc', 'includes/utils');
-\Drupal::ModuleHandler()->loadInclude('webform_civicrm', 'inc', 'includes/wf_crm_admin_help');
 
 /**
  * Implements _webform_edit_component().
  */
 function _webform_edit_civicrm_contact($component) {
-  civicrm_initialize();
+  \Drupal::service('civicrm')->initialize();
+  $adminHelp = \Drupal::service('webform_civicrm.admin_help');
   $form = array();
   $node = Node::load($component['nid']);
   if (empty($node->webform_civicrm) && node_is_page($node)) {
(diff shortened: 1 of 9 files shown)


KarinG:
Great 👍 I’ll look at it today!

KarinG:
Ok - my notes:

**Applies cleanly:** 👍 
karins-macbook-pro:webform_civicrm sysadmin$ cat 362.diff |patch -p1 --dry-run
patching file includes/contact_component.inc
patching file includes/wf_crm_admin_form.inc
patching file includes/wf_crm_webform_ajax.inc
patching file includes/wf_crm_admin_help.inc
patching file src/AdminHelpInterface.php
patching file src/Form/WebformCiviCRMSettingsForm.php
patching file src/Plugin/WebformElement/CivicrmContact.php
patching file webform_civicrm.module
patching file webform_civicrm.services.yml

Clearing all caches

I'm getting an error when trying to go to an existing webform's civicrm tab 
Error: Class 'Drupal\webform_civicrm\AdminHelp' not found in Drupal\Component\DependencyInjection\Container->createService() (line 265 of /Applications/MAMP/htdocs/d8civicrm.local/web/core/lib/Drupal/Component/DependencyInjection/Container.php) 

wf_crm_admin_help.inc ->has:
class AdminHelp implements AdminHelpInterface {

Does wf_crm_admin_help.inc need to be moved to src/ ? Similar to FieldOptions.php -> so that would mean moving/renaming it to -> src/AdminHelp.php file - From working on d9readiness a few months ago I seem to recall that all classes in src/ gets autoloaded.

It's quite possible a couple more cache clears will also get this file loaded but if I think we can help by putting it in the right place and then we'll also be clearing out the /includes one at a time!

jitendrapurohit:
> Does wf_crm_admin_help.inc need to be moved to src/

I see it is already in src/ path. drush cr should have loaded all the files.  See https://github.com/colemanw/webform_civicrm/pull/362/files#diff-75bf0d81910d3dcda4372e6ecddd4926

image

It seems the step `cat 362.diff |patch -p1 --dry-run ` hasn't applied the patch correctly for the replaced files. Can you use`git apply 362.diff` to apply the same pls? @KarinG

KarinG:
My first attempt created the new file `AdminHelpInterface.php` - but did not not create `AdminHelp.php`

All good now 👍 and confirming help text tooltips looks good:

image

PR #363, opened 2020-09-27
Fix activities_assignee help.

KarinG:
Overview
Help tooltip for Activities -> Assignee -> not rendering properly. Described here: 
https://www.drupal.org/project/webform_civicrm/issues/3173099

D7 or D8?
D8WFC only - no port required

Before
Help text is printed at the top of the page

After
Help text is in the tooltip

image


jitendrapurohit:
Tested and confirmed the display of help text on the form. Thanks @KarinG

PR #364, opened 2020-09-28
Revert "Use entity functions to save custom data"

colemanw:
Reverts #342 due to 2 problems discovered:

1. Custom data was [not being saved for any contact reference fields](https://www.drupal.org/project/webform_civicrm/issues/3171420).
2. Multi-record custom groups (e.g. the kind that displays as tabs on the contact summary screen) were not being saved.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -669,6 +669,127 @@ abstract class wf_crm_webform_base {
   }
 
   /**
+   * Save custom data for an entity
+   *
+   * @param $entity
+   *   Array of values
+   * @param $entity_id
+   *   Numeric id of entity
+   * @param $entity_type
+   *   Type of crm entity, e.g. "Contact"
+   * @param $known
+   *   Is this a known record (as opposed to a contact matched via dedupe rules)?
+   *   We only allow saving blank fields for known contacts.
+   * @param $contact_index
+   *   Contact's index
+   */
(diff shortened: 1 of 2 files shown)


colemanw:
Because of the regressions, I'm reverting this and releasing 7.x-5.3 immediately. Hopefully @mattwire can fix the two bugs and then we can un-revert.

ghost:
Just as well I checked this before installing 7.x-5.3.  The change in PR#342 is necessary for the recommended alternative solution to my PR#318 to work though fortunately I don't think we hit either of the cases mentioned above.  I see there is now a revised version of PR#342 - presumably this will be in a 7.x-5.4 release?

colemanw:
@maynardsmith yes, #366 has been merged and will be part of 5.4 release. Should be out this week.

Tekspertise:
The issue of 
Multi-record custom groups (e.g. the kind that displays as tabs on the contact summary screen) were not being saved.
appears to be still unresolved in 7.x-5.4
If more than one record is entered in the webform for the custom multiple record fieldset, these are all visible in the Webform results but only the first entry record is visible in Civi.
Have tested on 4 sites, all with 7.x-5.4, using different data types, all with same result

pmosey:
Similar to Tekspertise and also on verison 7.x-5.4.
If two or more custom contact reference fields are changed in a webform. Only the first one appears to be saving in civi. I'll keep testing.

Created issue [here](https://www.drupal.org/project/webform_civicrm/issues/3196661).

PR #348, opened 2020-08-26
Use entity functions to save custom data

mattwire:
Overview
----------------------------------------
D8 port of https://github.com/colemanw/webform_civicrm/pull/342

Comments
----------------------------------------
Note I didn't test this on D8 yet.

PR #365, opened 2020-09-28
Revert "Use entity functions to save custom data"

colemanw:
Reverts colemanw/webform_civicrm#348 for the 8.x branch.
2 bugs were found in the [7.x version of this PR](https://github.com/colemanw/webform_civicrm/pull/342) which probably affect 8.x as well:

1. Custom data was [not being saved for any contact reference fields](https://www.drupal.org/project/webform_civicrm/issues/3171420).
2. Multi-record custom groups (e.g. the kind that displays as tabs on the contact summary screen) were not being saved.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -662,6 +662,130 @@ abstract class wf_crm_webform_base {
   }
 
   /**
+   * Save custom data for an entity
+   *
+   * @param $entity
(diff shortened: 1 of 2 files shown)


jitendrapurohit:
was able to replicate the regression and confirm that this PR saves the field correctly.

Normal Contact reference field used to work fine. It is on the case that failed to save after #348. Merging.

PR #366, opened 2020-10-02
Use entity functions to save custom data 

jitendrapurohit:
Overview
(Resubmission of #342) + fixes 2 regressions

- Custom data was not being saved for any contact reference fields. 

- Multi-record custom groups (e.g. the kind that displays as tabs on the contact summary screen) were not being saved.

Ping @KarinG @mattwire 

Have tested the final changes with different custom field types. Multi-Valued tested with create/edit and create-only mode.


KarinG:
Great @jitendrapurohit - I'll test this today! 
@mattwire - would really appreciate your eyes on this too - this was your idea :-)

mpaulso:
@KarinG and I have just tested this. We applied the patch and created a contact reference field and collected and saved multidata field. We confirmed these data were stored in CiviCRM.

image

image

image

KarinG:
@colemanw - we reviewed @jitendrapurohit PR - since you're the one who reverted the original one by @mattwire - could you please have a look and Merge this PR if think it's ready now?

jitendrapurohit:
just pushed a removal of an unnecessary line. It wasn't breaking anything but wasn't required either.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2233,8 +2219,10 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
    *   Leave blank - used internally to recurse through data
    * @internal param $depth int
    *   Leave blank - used internally to track recursion level
+   * @param $customOnly
+   *   TRUE if only custom contact reference needs to be filled.
    */
-  private function fillContactRefs($values = NULL, $depth = 0) {
+  private function fillContactRefs($values = NULL, $depth = 0, $customOnly = FALSE) {
```
colemanw: Adding a new param on the end makes this function awkward to use. Better to add it to the beginning.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2367,9 +2371,26 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
         // Only known contacts are allowed to empty a field
         if (($val !== '' && $val !== NULL && $val !== []) || !empty($this->existing_contacts[$c])) {
           $this->data[$ent][$c][$table][$n][$name] = $val;
+          if (substr($name, 0, 6) === 'custom') {
+            $multivaluesCreateMode = NULL;
+            // Handle 'Create mode' for multivalues custom fields of Contact entity.
+            if ($ent === 'contact' && is_numeric(substr($name, 7))) {
+              $createModeKey = 'civicrm_' . $c . '_contact_' . $n . '_' . $table . '_createmode';
+              $multivaluesCreateMode = isset($this->data['config']['create_mode'][$createModeKey]) ? (int) $this->data['config']['create_mode'][$createModeKey] : NULL;
+              $multiRecordGid = CRM_Core_BAO_CustomField::isMultiRecordField(substr($name, 7));
```
colemanw: This is a CiviCRM internal function. The goal of this module is to only use the API.
I think you can access this info with something like `wf_crm_get_fields('sets')['cg_' . substr($name, 7)]['max_instances']`

colemanw (changes requested):
Thanks for your work on this @jitendrapurohit.
I've made a couple suggestions.

jitendrapurohit:
Thanks @colemanw 👍 . Have updated the PR and retested the multivalued custom data submissions.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -125,7 +125,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
 
     $this->fillDataFromSubmission();
     //Fill Custom Contact reference fields.
-    $this->fillContactRefs(NULL, 0, TRUE);
+    $this->fillContactRefs(TRUE, NULL, 0);
```
colemanw: The reason I suggested adding the new param to the beginning is so you could omit default params.
```suggestion
    $this->fillContactRefs(TRUE);
```

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2367,9 +2371,26 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
...
         // Only known contacts are allowed to empty a field
         if (($val !== '' && $val !== NULL && $val !== []) || !empty($this->existing_contacts[$c])) {
           $this->data[$ent][$c][$table][$n][$name] = $val;
+          if (substr($name, 0, 6) === 'custom') {
+            $multivaluesCreateMode = NULL;
+            // Handle 'Create mode' for multivalues custom fields of Contact entity.
+            if ($ent === 'contact' && is_numeric(substr($name, 7))) {
+              $createModeKey = 'civicrm_' . $c . '_contact_' . $n . '_' . $table . '_createmode';
+              $multivaluesCreateMode = $this->data['config']['create_mode'][$createModeKey] ?? NULL;
+              $cgMaxInstance = $this->all_sets[$table]['max_instances'] ?? 1;
```
colemanw, line 2384: That's even better than my suggestion :+1: :+1: :+1:

includes/wf_crm_webform_postprocess.inc
jitendrapurohit: ah yes, sorry. Have updated.

colemanw:
Ok I'm happy with this now based on code review & testing.

KarinG:
Awesome guys! Thank you 👍 

@jitendrapurohit - could you please port this PR to D8?

jitendrapurohit:
@KarinG done https://github.com/colemanw/webform_civicrm/pull/371

PR #369, opened 2020-10-05
D8 - Decouple list of payment processors from contribution page

jitendrapurohit:
Overview
D8 port of #368

D7 or D8?
D8

Before
Payment processor list depends on the contribution pages.

After
- All Payment processor are listed on the contribution tab.
- User-select removed from Payment processor mode so admins have to select either live or test.
- `fixPaymentProcessorID ` isn't available in d8wfc so the removal code isn't required here.

Comments
ping @KarinG @mattwire

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1584,6 +1584,9 @@ class wf_crm_admin_form {
           (!empty($val) || (in_array($name, array('num_terms', 'is_active', 'is_test', 'payment_processor_id'))))) {
           // Don't add data for non-existent contacts
           if (!in_array($ent, array('contact', 'participant', 'membership')) || isset($this->data['contact'][$c])) {
+            if ($name == 'payment_processor_id' && !empty($val) && is_numeric($val)) {
+              $val = $this->getPaymentProcessorValue($val);
+            }
             $this->data[$ent][$c][$table][$n][$name] = $val;
           }
         }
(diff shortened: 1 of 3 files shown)


KarinG:
@jitendrapurohit - this is looking good thus far - I've installed it on a D8 and can see the payment processors. Let's keep this PR Open until you have had a chance to add the other remaining bits necessary before we can completely remove the dependency, like the email receipt bit for example.

KarinG:
@jitendrapurohit - would it help you if we merge this while you are still working on the mailing receipt bits?

jitendrapurohit:
@KarinG yes, this is independent of the email receipt addition code. So we can merge this PR if it has been approved from testing. I can rebase the other if needed.

KarinG:
Hi! Just tested this -> we still need to select a Contribution page to make this work. It also still uses the Financial Type that is on that Contribution. So we'll need Financial Type field to remove the dependency. 

Otherwise live contribution was succesful!

KarinG:
Oh and we'll want to include currency as well!

KarinG:
@jitendrapurohit - Merging this - leftover work will go into a seperate PR 👍

jitendrapurohit:
Thanks @KarinG

petednz:
how hard to push back to d7

KarinG:
@petednz - not impossible, but a bit scary. It's a major change - I am thinking it's easier to roll this out on D8 (fewer and typically quite knowledgable admins).

petednz:
ok. thought i would ask as am just doing a d7 form that would be easier :-). The foreign language one i mentions on the GTM

KarinG:
I hear you - I'm just a bit worried about such a major change to how things work for D7 admins. Maybe I should not be worried :-)

PR #370, opened 2020-10-06
Issue#3174133: Allow static widget to use values other than default value

nishant-bhorodia:
Overview
Currently if you add two contacts to the webform with configuration below:
1. Contact A: Individual (Widget: Autocomplete)
2. Contact B: Employer of contact A (Widget: Static)

With above configuration the below will happen:
1. Change contact A to another individual contact.
2. Contact B will change as per the contact chosen in step 1
3. Save the webform
4. See the submission, you will notice the contact B is still not changed even though it loaded fine in step 2.

This PR fixes that.

D7 or D8?
The PR is for D7.

Before
CATL-1711-before

After
CATL-1711-after

Technical Details
In `function wf_crm_fill_contact_value` the code sets the value for `$element['#value']` because of which webform skips this element and doesn't fetch it's value from `$form_state['input']` because of which the value for existing contact element remains unchanged when the `static` widget is used.
Removing that code as part of this PR.

### DIFF
--- a/includes/contact_component.inc
+++ b/includes/contact_component.inc
@@ -549,8 +549,6 @@ function _webform_render_civicrm_contact($component, $value = NULL, $filter = TR
 function wf_crm_fill_contact_value($node, $component, &$element, $ids = NULL) {
   $cid = wf_crm_aval($element, '#default_value', '');
   if ($element['#type'] == 'hidden') {
-    // User may not change this value for hidden fields
-    $element['#value'] = $cid;
     if (!$component['extra']['show_hidden_contact']) {
       return;
     }


tunbola:
@nishant-bhorodia , A few questions
Is the static widget equivalent to a hidden form element? Because I can see that the changes made to the code is for when the form element type is hidden.
Also do we have plans to put up a PR for D8 because I can remember someone mentioned on of the D7 PR's that we are to put up equivalent D8 PR's

nishant-bhorodia:
Hey @tunbola 
Yes, here **hidden** is for static widget. Please refer this: https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/contact_component.inc#L107

nishant-bhorodia:
Hey @colemanw @KarinG 
Hope you are doing well !!
This passed testing on our end, please would it be possible to review?
Thanks!!

jamienovick:
@KarinG @colemanw - Hey guys, sorry to push but we're coming up against it for another deadline. Any chance we can get this one reviewed and merged?! Pretty please with sugar on top 😉...

jamienovick:
@colemanw any chance of getting this merged? Otherwise we are going to have to create a v5.3 fork :( today...

Anything you can do would be much apreciated!

colemanw:
@jamienovick I've tested this and I think it's mergable.

KarinG:
@colemanw - can you please also port this to D8?

colemanw:
@KarinG done

nishant-bhorodia:
Thanks a lot @colemanw @KarinG @jamienovick  :)

PR #372, opened 2020-10-11
D8 - Make email receipt independent of the contribution page

jitendrapurohit:
Overview
Add Email Receipt config on the webform instead of fetching it from the contribution page.

D7 or D8?
D8

Before
Sending Email receipt is dependent on the contribution page selected on the webform.

After
New email receipt config added.

Screenshot 2020-10-11 at 7 06 17 PM

- If enabled, contribution receipt will be sent to the donor.
- CC and BCC can be set from the webform.
- Upgrade code added to update the existing webforms to have the email fields default to the contribution page selected.

Comments
ping @KarinG 

Currently, this requires an additional civicrm patch - https://github.com/civicrm/civicrm-core/commit/b83493e9e454233b5a13e2e95db923c4e8973da6 due to a bug in civi core which uses cc and bcc from the contribution page even if it is overridden in the sendconfirmation api params.

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1082,6 +1082,66 @@ class wf_crm_admin_form {
         $this->form['contribution']['sets']['lineitem'][$fs][$fid] = $this->addItem($fid, $field);
       }
     }
+
+    // Receipt
+    $n = wf_crm_aval($this->data, "receipt:number_number_of_receipt", 0);
+    $this->form['contribution']['sets']["receipt_1_number_of_receipt"] = [
+      '#type' => 'select',
+      '#title' => t('Enable Receipt?'),
+      '#default_value' => $n,
+      '#options' => [t('No'), t('Yes')],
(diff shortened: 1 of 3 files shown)


KarinG:
@jitendrapurohit -looks like https://github.com/colemanw/webform_civicrm/pull/372/commits/2ae1f94c1526000ecb71544e73ea44e9566ff964
is merged - so that would be available as of... CiviCRM Core 5.32 right? 

Does the `sendconfirmation` api allow for a little text area (Receipt Message) to be send along? (like people can do in the Contribution Page config).

jitendrapurohit:
@KarinG Pushed an update to add more fields so the final display looks like -

image

Have created a PR for the core fix at https://github.com/civicrm/civicrm-core/pull/18789

mattwire:
Core PR now merged. This is looking really good!

KarinG:
I just live tested this during Lunch & Learn Nov 05! And I got a receipt!
image

PR #373, opened 2020-10-15
Invoke Test mode at run time with a URL argument.

ghost:
There is an option to set the Payment Processor into Test mode when configuring the Contribution section of the Webform. 

This change adds the ability also to set Test mode at run time, with a URL argument of '?payment=test'. This enables one to run regression tests for alterations made to the live site without having to take it out of service to change the configuration parameter, with the inconvenience and risk of error involved in doing so.

Overview
Adds the URL Argument as an additional input to the Test mode decision in fixPaymentProcessorID().

D7 or D8?
Tested on D7.  I don't know if it applies to D8 or not.

Before
Can only Invoke Test mode by changing Webform configuration.

After
Can also invoke Test mode at Run time.

Comments
First mentioned in Issue https://www.drupal.org/project/webform_civicrm/issues/3176642

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -836,9 +836,14 @@ abstract class wf_crm_webform_base {
    *   instead of test.  So we "fix" the processor ID here to be the correct one for live/test.
    * An "improved" fix would involve changing the configuration saved by the user but this avoids that.
    *
+   * This way also allows one to invoke Test mode at run time with a URL argument.
+   *
    * @throws \CiviCRM_API3_Exception
    */
   protected function fixPaymentProcessorID() {
+    if (isset($_GET['payment']) && $_GET['payment'] == 'test') { 
+    	$this->data['contribution'][1]['contribution'][1]['is_test'] = 1;
+    }
     // Check for is_test and payment_processor_id (pay later = 0 is set to '' here. is_test has no meaning for pay later).
     if (!empty($this->data['contribution'][1]['contribution'][1]['is_test'])
         && !empty($this->data['contribution'][1]['contribution'][1]['payment_processor_id'])) {


KarinG:
In D8 -> Webform module itself has TEST functionality. With the click of one button (TEST) -> you get a random Beatles combination contact and all fields are filled (see screenshot): 

image
image

So 👍   I like the idea - but not via the URL -> we would be implementing this via the TEST button (if one hits TEST -> set payment processor to test mode). 

D7WFC is closing in on maintenance only with new features going into D8WFC (which is also D9WFC at this point).

ghost:
Presumably the Payment Test Mode would be separate from the random data generation you show - I am running test scripts which want a defined set of input data in order to be able to verify the results.

A URL argument works for me for several reasons: 
-  It is simpler to implement the patch than having to code extra buttons.
-  it is simpler to code the test script, just a change to the URL literal.
-  it means the test can be run without having additional permissions; it is trying to mimic the user behaviour as far as possible.
-  we are a good way off implementing D8, for various reasons.

It looks like I will have to continue with this as a private patch if the PR Is not accepted, since it is significantly easier and less disruptive to run tests that way.  Is there any hook or other interface which might provide an alternative route without hacking standard code?

mattwire:
@maynardsmith @KarinG I do like the idea of a URL parameter - that's how civicrm forms work and it's *really* useful to do a quick test on a site. Especially when you are not logged in.  I also like the idea of integrating with the D8 "test" function but I think that's slightly different (eg. you might actually want to do a live payment in test mode or vice versa).

A few thoughts:
1. Any change should be looked at once https://github.com/colemanw/webform_civicrm/pull/369 and the related D7 PR is merged because that changes the way payment processors are loaded.
2. If we add a URL parameter it may make sense to follow the existing CiviCRM naming convention (though it may not as we may be doing something slightly differently on webform (eg. on contribution pages all entities are created as test entities - do we also want that on webform).
3. There is not yet a setting in CiviCRM to enable/disable test mode on forms but I'd like there to be. As I feel that permanently being able to run forms in test mode via URL params could be prone to abuse (though I've never actually seen any evidence of this).

KarinG:
The plan is:

New work goes into D8WFC. D7WFC is heading towards maintenance only. Right now new features would go into D8WFC and will still be back ported where possible and if determine its safe to do so. 

**New work:**
- implementing Kernel tests [test PRs]
- leveraging the native Webform TEST button for more tests (look at it as a UI to the tests in Webform module itself)
- decoupling contribution pages -> I was initially thinking we would be making mode (test/live) a user select -> may decide to go back to that as that would let you put it on a (your test) form with a user select and then you can (like with any other webform field) -> use a query parameter to set the default -> and that would achieve your via URL objective.

lalgwebdev:
I understand that this will not appear in D7WFC.

I have also realised a security problem with the change proposed above, namely that anyone who knows about the test query string can complete transactions normally on  a live system but only having made a dummy payment.  For my own purposes (and anyone else following this) I have therefore made the test option subject to a custom Permission (created elsewhere by `hook_permission()`) as follows:
```
+    if (user_access('LALG use Stripe Test mode')) {
+    	if (isset($_GET['payment']) && $_GET['payment'] == 'test') { 
+    		$this->data['contribution'][1]['contribution'][1]['is_test'] = 1;
+    	}
+    }
```
Regarding a possible implementation on the D8WFC my objective is to be able to run tests on the live configuration without having to manually reconfigure it to do so.  Either an option under the TEST button (provided it did not also have to fill the form with data), or an embedded Select would be suitable provided the necessary security lock was available.  The use of the URL query string was just the easiest feature available on a transaction by transaction basis that occurred to me.

KarinG:
Would you like to create a PR for D8/9WFC that would make Payment Processor Mode a -user select-? That would give you options to quickly switch LIVE to TEST; 

PS - We only ever test with real payment processors in LIVE mode (just using fake VISA credentials). To me it's the only way to be absolutely sure everything lands in the correct place as data navigates through payment processor servers, and into CiviCRM. We've been burned too many times not seeing the data in all the places we would expect to see it. Our webform civicrm code has such a LIVE mode with fake VISA credentials in a Contribution Test! And so does our iATS Payments Extension. Real pathways, just fake monies.

lalgwebdev:
Ah!  Maybe that's a better way to go.  Are these your private fake credentials, or generally available?  And where does that leave one with the problem of some bad actor using the fake credentials on a Live site?

lalgwebdev:
In principle I can look at taking this forward in D8WFC, but we won't have a Live D8 site to worry about for some time, and in development I'm happy just to configure the PP in test mode the whole time, so I won't be doing anything for quite a while.

KarinG:
> Ah! Maybe that's a better way to go. Are these your private fake credentials, or generally available? And where does that leave one with the problem of some bad actor using the fake credentials on a Live site?

Generally available -> for iATS Payments. They have public TEST accounts - so you can use them even before signing up with them. I do have some private ones with them as well [less cluttered] for development work. I don't know if Stripe has any such option -> but I know in the past I've also taken a Client's PayPal Pro Sandbox credentials and have plugged those into the Live config of a CiviCRM Payment processor and that worked great 🎉

ghost:
Yes, thanks, Stripe does just the same, which is what I've been using.  It does need the PP putting into test mode, which is what I've been going around on.  As I said above, low priority for now though.

PR #375, opened 2020-10-19
Autocomplete Element filter cannot be configured with smart groups

kainuk:
Overview
Webform_civicrm has a special element type "Existing Contact". It can be used to lookup existing contacts in CiviCRM and use them in the form. It is possible to restrict the list of contacts by requiring that they have to belong to a certain group. However, in the list with groups, the smart groups are not shown.

More information at https://www.drupal.org/project/webform_civicrm/issues/3177748

D7

Before
webform_smartgroup
(1) The smart groups are not shown.

After
(1) The smart groups are shown.

Technical Details
The filter does work with smart groups


kainuk:
It was a blib. Could not reproduce it myself.

PR #377, opened 2020-10-26
Load cases by manager as well as client

colemanw:
Overview
Allows cases to be loaded not just by client but also by manager.

D7 or D8?
D7. Will port once accepted.

Before
When loading a case from the url, the module would check if _case client_ was present on the webform, and only load the case if true.

After
When loading a case from the url, the module will check if _case client_ **OR** _case manager_ is present on the webform, and only load the case if true.

Comments
-------------------
Removed some unnecessary hacks that had been added to work around the lack of this feature.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -271,10 +271,6 @@ abstract class wf_crm_webform_base {
         case 'case_roles':
           $to = $component['extra']['default_relationship_to'];
           $case = $component['extra']['default_for_case'];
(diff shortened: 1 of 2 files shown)


KarinG:
I've pinged Olly to please take a look.

gibsonoliver:
Hi, I can't see any problems with that. Sounds like a good idea which should have no effects on anyone's existing setup

Olly

KarinG:
Awesome - Thank you for confirming that Olly 👍

PR #378, opened 2020-10-27
CATL-1711: Allow static widget to use values other than default value

colemanw:
8.x port of #370


KarinG:
Thank you @colemanw 👍

PR #381, opened 2020-11-02
Verify a contact is created when the Webform is submitted

mglaman:
Overview
Extends the test to submit the webform and verify the contact over the CiviCRM API

Before
Form loaded with a notice, didn't test submit

After
Form loads, no notices, and submitted!

### DIFF
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -488,7 +488,7 @@ class CivicrmContact extends WebformElementBase {
     $cid = wf_crm_aval($element, '#default_value', '');
     $contactComponent = \Drupal::service('webform_civicrm.contact_component');
     if ($element['#type'] == 'hidden') {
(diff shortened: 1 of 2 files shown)


src/Plugin/WebformElement/CivicrmContact.php
```diff
@@ -488,7 +488,7 @@ public static function wf_crm_fill_contact_value(WebformInterface $node, array $
     $cid = wf_crm_aval($element, '#default_value', '');
     $contactComponent = \Drupal::service('webform_civicrm.contact_component');
     if ($element['#type'] == 'hidden') {
-      if (!$component['#show_hidden_contact']) {
+      if (!isset($component['#show_hidden_contact']) || !$component['#show_hidden_contact']) {
```
mglaman, line 491: This means something else is funky with setting up the form, but it removes the error.

KarinG:
Awseome 👍 Getting close! :-)

PR #382, opened 2020-11-02
D7 - [WIP] Add new setting to respect existing membership_type on updates

sluc23:
Overview
Following the discussion threads on [#64](https://github.com/colemanw/webform_civicrm/pull/64) and [#199](https://github.com/colemanw/webform_civicrm/pull/199), `CiviCRM` core and `webform_civicrm` don't allow to create and update in a public form (Drupal Webforms or Contribution Pages) two or more Memberships with different types belonging to same organization for a single contact separately.

The arguments raised by @KarinG in [#199](https://github.com/colemanw/webform_civicrm/pull/199) are correct, **but** are not taking into consideration that this can be achieved from the CiviCRM backend itself.. there's no limitation for an user in Civi to create 2 different memberships of different membership_types belonging to same Org, for a contact. And because of that we have loads of Organizations using the CiviCRM in that way.

But we have the limitation on public forms to follow that rule. And this is excluding a lot of use cases from different ORGs to have for instance: 2 different Webforms to create 2 different `membership_types`, the default behaviour nowadays for the second webform will update the first membership that the User has, regardless the `membership_type`, instead of expecting behaviour which would be to create the second membership.
So this **[WIP] PR** is similar than the one @jaapjansma submitted in [#199](https://github.com/colemanw/webform_civicrm/pull/199) to fill the gap of this.

Still in **[WIP]** state so we can reopen this discussion about this topic.

D7 or D8?
D7

Before
If the User filling the webform, exists and has a membership, the webform will update that membership (the latest one in case of having many), regardless the `membership_type` defined for the membership in the `webfrom_civicrm`

After
If the User filling the webform, exists and has a membership, the webform will update only an existing membership of the same `membership_type`. If it doesn't exists webform_civicrm will create it.

Technical Details
It add s a new `webform_civicrm` setting, by default disabled, so the normal behaviour it's not affected at all, as it is right now.

Comments
Comments and ideas welcome :)

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1185,6 +1185,12 @@ class wf_crm_admin_form {
       '#default_value' => $this->settings['new_contact_source'],
       '#description' => t('Optional "source" label for any new contact/participant/membership created by this webform.'),
     ];
+    $this->form['options']['existing_membership_type'] = [
+      '#type' => 'checkbox',
+      '#title' => t('Keep Existing Membership Type'),
+      '#default_value' => !empty($this->settings['existing_membership_type']),
+      '#description' => t('If enabled, when the webform includes a Membership, the Membership Type will be respected in order to update it (if User already has it, if not it will created).'),
+    ];
   }
 
   /**
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1123,18 +1123,21 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
       }
 
       // Search for existing membership to renew - must belong to same domain and organization
-      // But not necessarily the same membership type to allow for upsell
+      // Depending on existing_membership_type setting allow update different membership type for upsell or not
       if (!empty($params['num_terms'])) {
         $type = $types[$params['membership_type_id']];
         foreach ($existing as $mem) {
           $existing_type = $types[$mem['membership_type_id']];
           if ($type['domain_id'] == $existing_type['domain_id'] && $type['member_of_contact_id'] == $existing_type['member_of_contact_id']) {
-            $params['id'] = $mem['id'];
+            if (empty($this->settings['existing_membership_type'])) {
+              $params['id'] = $mem['id'];
+            }
             // If we have an exact match, look no further
             if ($mem['membership_type_id'] == $params['membership_type_id']) {
               $is_active = $mem['is_active'];
               $membershipStatus = $mem['status'];
               $membershipEndDate = $mem['end_date'];
+              $params['id'] = $mem['id'];
               break;
             }
           }
--- a/webform_civicrm.install
+++ b/webform_civicrm.install
@@ -107,6 +107,13 @@ function webform_civicrm_schema() {
         'default' => '',
         'description' => 'Source label for newly created contacts',
       ],
+      'existing_membership_type' => [
+        'description' => 'Update the Membership based on existing membership types.',
+        'type' => 'int',
+        'size' => 'tiny',
+        'not null' => TRUE,
+        'default' => 0,
+      ],
     ],
     'primary key' => ['nid'],
   ];
@@ -831,4 +838,18 @@ function webform_civicrm_update_7403() {
     'default' => 0,
   ];
   db_add_field('webform_civicrm_forms', 'create_new_relationship', $field);
+}
+
+/**
+ * Add field to respect membership_type or not on existing Membership updates.
+ */
+function webform_civicrm_update_7404() {
+  $field = [
+    'description' => 'Update the Membership based on existing membership types.',
+    'type' => 'int',
+    'size' => 'tiny',
+    'not null' => TRUE,
+    'default' => 0,
+  ];
+  db_add_field('webform_civicrm_forms', 'existing_membership_type', $field);
 }


sluc23:
probably @mattwire interested on this discussion as well

mattwire:
Yes, I'd like to see the restriction on "a single membership type per org" on public civicrm forms removed completely. There is a check in about 3 places in CiviCRM core.

But I would think for webform it would be nice to have a flexible setting that allows you to choose one of:
1. Update existing membership of same type.
2. Update existing membership of different type and change type.
3. Add new membership and don't touch existing.

sluc23:
enhanced implementation of this PR based on @mattwire suggestions here: -> #383

KarinG:
7.x-5.x is really maintenance only at this stage. New functionality should go into 8.x-5.x - especially if the functionality is changing how Webform CiviCRM has worked for years. Like the Membership changes like you're proposing here. I'd be happy to consider such changes for D8WFC: still only just over 100 sites using it - and most of them have an experienced site-admin involved so changes in the category 'how it has always worked' are much easier to communicate.

sluc23:
thanks @KarinG for your reply. We are still heavy users of `D7` + `webform_civicrm-7.x-5.x`. **2021** will be the year to start migrating all our sites from **D7** to **D8/9** (around 200) but still it will be a slow uphill road to achieve that.. meanwhile we are trying to enhance as much as possible our customer's experience with `webform_civicrm-7.x-5.x` for at least max of 2 more years, until [D7 EOL in Nov '22](https://www.drupal.org/psa-2019-02-25).

If, as you said, the module's supporters decision is to freeze inclusion of new features in `webform_civicrm-7.x-5.x`, it's good time to know that, to probably start using our own fork of the module, or if more people are interested on keep on maintaining a richer `7.x-5.x` version, probably we can join efforts in that direction..
 
We'd love to start contributing in `8.x` branch.. but to be realistic don't see that possible in the short-term..

KarinG:
We're also in the process of adding automated testing infrastructure for D8WFC - so even small edits/bug fixes - like your other PR we'll soon want to see those in 8.x-5.x first so that we can run the automated tests -> before we backport them to D7WFC.

If we can help to get you up to and running re: D8/D9 - let me know. We have most of our major clients still on D7 as well - but all new sites start on D8 and we're actively migrating projects from D7 to D8. The richness of D8 Webform CiviCRM module itself (like e.g. computed elements, or masking) are huge.

sluc23:
@KarinG I would like to start migrating this new feature to `webform_civicrm-6.x.x` for Drupal9. We have it working for more than a year now in our D7 [forked version](https://github.com/ixiam/webform_civicrm/pull/2).

Before starting this development, what's the best procedure to start discussing the functionality, proper approach, etc.. ? to have a consensus to move this PR forwards? Raise an issue in drupal.org queue? document it somewhere? another PR to start the discussion there?

tx!

KarinG:
Hi @sluc23, that great!

- important that the current existing default behaviour is respected and that your new checkbox adds functionality for those that would like to use it. 
- you can PR it directly to this Github repo `https://github.com/colemanw/webform_civicrm`
- since it's new functionality I would encourage a test to preserve it. Functional Tests are easy to write -> check out the examples in: `https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/MembershipSubmissionTest.php` - you can add a function e.g. `testSubmitRespectType` and you will be able to copy/paste lots from the other tests in this file.
- you don't need a local test-dev to write these tests -> feel free to PR to the github repo and let it do the work. You can also download assets (screenshots of your tests). If you do want to install a local test-dev (it is a lot faster when you can just run the one test and not the entire battery) then here's my write up for it -> https://civicrm.stackexchange.com/questions/38141/how-can-i-run-the-drupal-9-webform-civicrm-unit-tests-on-my-local

sluc23:
@KarinG according to your [tutorial](https://civicrm.stackexchange.com/questions/38141/how-can-i-run-the-drupal-9-webform-civicrm-unit-tests-on-my-local) , I first need to install CiviCRM and the execute the tests (step **E.**, phpunit is step **M.**).

I've managed to setup a local Unit Tests environment with docker containers (I don't use PHPStorm, so it's a variation from your tutorial, executing manually the tests), but when I executed the tests I've got:
```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest::testOnePageDonation
RuntimeException: The provided database connection in SIMPLETEST_DB contains CiviCRM tables, use a different database.
```

So my question is: Do the test must be executed in a Drupal site with CiviCRM uninstalled? 
  A. If that the case, is it your tutorial wrong in the steps sequence?
  B.  If not the case, why the Tests setup, checks for existing `civicrm_%` tables in the DB, to abort the test execution?
    ref [here](https://github.com/colemanw/webform_civicrm/blob/6.2.1/tests/src/FunctionalJavascript/CiviCrmTestBase.php#L70)

Am I missing any step ?

KarinG:
Awesome! 

That error message is really a stop - you are about to blow away your existing database. I had blown away my local database (that had configs) too many times :-) So simply drop all the tables (and triggers) in your database. And of course if it has configs that you'd like to keep -> make yourself a backup copy before you drop everything.

So yes that tutorial write up on SE predates this safety measure. I'll make some edits!

sluc23:
@KarinG this means that if I run the full set of tests as `phpunit web/modules/contrib/webform_civicrm`or any other combination where I run more than 1 consecutive test.. I have to drop all tables between the tests?

How's this implemented in GitHub CI? I don't see an easy way to locally implement a table deletion inside the `phpunit` execution or if not, I'll be forced to run only 1 test at a time...

KarinG:
There is a tear down in between so you only have to dump the tables if they exist (eg if a test failed on syntax error and it never got to that stage).

But I would not run the entire test suite on my local - let Github do the work! So I write a new test on my local - check that it works and does what I want it to do - and then I PR it to this repo and let Github automatically run all tests against all CiviCRM versions.

PR #383, opened 2020-11-02
[WIP] D7 - Add new Membership Type action setting

sluc23:
Overview
This is an enhanced implementation of of [#382 ](https://github.com/colemanw/webform_civicrm/pull/382) based on the suggestions of @mattwire  --> https://github.com/colemanw/webform_civicrm/pull/382#issuecomment-720423579

D7 or D8?
D7

Before
If the User filling the webform, exists and has a membership, the webform will update that membership (the latest one in case of having many), regardless the membership_type defined for the membership in the webfrom_civicrm

After
Now the Membership update/creation action will be based in the setting selected as:
1.  Update existing membership of different type and change type.
2.  Update existing membership of same type.
3.  Add new membership and don't touch existing.

Technical Details
Noticed I had to revert commit 32f8acab6cb976f5c28f7be812ab0e844aa7373e for CiviCRM ESR 5.27.7 compatibility

Comments
still [WIP]

### DIFF
--- a/includes/wf_crm_admin_form.inc
+++ b/includes/wf_crm_admin_form.inc
@@ -1185,6 +1185,12 @@ class wf_crm_admin_form {
       '#default_value' => $this->settings['new_contact_source'],
       '#description' => t('Optional "source" label for any new contact/participant/membership created by this webform.'),
     ];
+    $this->form['options']['existing_membership_type'] = [
+      '#type' => 'checkbox',
+      '#title' => t('Keep Existing Membership Type'),
+      '#default_value' => !empty($this->settings['existing_membership_type']),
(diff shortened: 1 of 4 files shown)


sluc23:
Please note that the default setting will be `1.  Update existing membership of different type and change type. `, which is the expected behaviour right now in `webform_civicrm`. So this PR won't introduce any change if the User does not change this setting value

mattwire:
@sluc23 Needs rebase. Can you submit a PR for 8.x-5.x branch first and then we can backport to 7.x once merged.

sluc23:
@mattwire I'm not able to submit to 8.x-5.x, since we haven't started to work D8 yet.. we are starting the transition from D7 to D8 in the following months, so it will take us some time to get into the loop of D8.

PR #384, opened 2020-11-02
Fetch domain because APIv3 does not do so reliably

colemanw:
Overview
APIv3 has inconsitent handling of the 'current_domain' keyword,
so we'll do it ourselves to ensure it works.

This fixes a bug noted in https://github.com/colemanw/webform_civicrm/commit/32f8acab6cb976f5c28f7be812ab0e844aa7373e


KarinG:
Thank you @colemanw 👍 
@sluc23 - can you please confirm this fixes the issue?

sluc23:
@KarinG tested and it works form me! thanks @colemanw

PR #387, opened 2020-11-05
Update Help Text - Recurring

KarinG:
Overview
Update the Help Text for Recurring Contributions.

D7 or D8?
D8 - to be ported

### DIFF
--- a/src/AdminHelp.php
+++ b/src/AdminHelp.php
@@ -154,7 +154,7 @@ class AdminHelp implements AdminHelpInterface {
     return '<p>' .
       t('Supported payment processors enabled on the contribution page are available here. "Pay Later" option allows the user to purchase events/memberships without entering a credit card.') .
       '</p><p>' .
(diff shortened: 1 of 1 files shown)


mglaman:
```
Behat\Mink\Exception\ElementNotFoundException: Form field with id|name|label|value "Enable CiviCRM Processing" not found.
```

🤔 this is a random failure I saw before, too. See this comment:

```
    $this->getSession()->getPage()->clickLink('CiviCRM');
    // @todo Randomly this fails saying that the checkbox does not exist.
    $this->assertSession()->waitForField('Enable CiviCRM Processing');
    $this->htmlOutput();
    $this->getSession()->getPage()->checkField('Enable CiviCRM Processing');
```

https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/tests/src/FunctionalJavascript/ContactSubmissionTest.php#L25

Maybe you can rerun it and see what happens

PR #388, opened 2020-11-08
Remove contribution page dependency from the webform

jitendrapurohit:
Overview
Adding a contribution section no longer requires enabling a contribution page.

D7 or D8?
D8

Before
The contribution section is dependent on civicrm contribution page for various fields - financial type, currency etc.

After
Contribution Page dropdown is removed. 

webform_contribution

Comments
Upgrade code included. Ping @KarinG

### DIFF
--- a/includes/wf_crm_admin_component.inc
+++ b/includes/wf_crm_admin_component.inc
@@ -93,7 +93,7 @@ class wf_crm_admin_component {
       $this->form['value']['#empty_value'] = '';
     }
 
-    elseif ($name == 'contribution_page_id') {
+    elseif ($name == 'enable_contribution') {
       $this->form['value']['#access'] = FALSE;
       $this->form['#prefix'] = '<p>' . t('This is a placeholder for a Contribution Page.') . '</p>';
     }
@@ -207,7 +207,7 @@ class wf_crm_admin_component {
   public function adjustPageBreak() {
     $components = $this->node->webform['components'];
     foreach ($components as $cid => $component) {
-      if ($component['form_key'] == 'civicrm_1_contribution_1_contribution_contribution_page_id') {
+      if ($component['form_key'] == 'civicrm_1_contribution_1_contribution_enable_contribution') {
         // Iterate up through parents
         while ($component['pid']) {
           $component = $components[$component['pid']];
@@ -573,7 +573,7 @@ class wf_crm_admin_component {
             $row['data'][3]['data'] = $val;
           }
           // Contribution page - link to civicrm config form instead of component edit form
-          if ($name == 'contribution_page_id') {
+          if ($name == 'enable_contribution') {
             $type = t('CiviCRM Billing Fields');
             $class .= ' contribution';
             $row['data'][6]['data'] = l(t('Configure'), 'civicrm/admin/contribute/settings', array(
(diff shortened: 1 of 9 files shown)


KarinG:
Awesome - test failure `Subcommand @composer compile  returned with error code 1 `unrelated.

KarinG:
Hi! @jitendrapurohit - testing this - my notes:

**1. Installing this on a site that has Webforms with Contributions set up -> will break these webforms.** We're in beta releases - so not requiring an upgrade path but this change breaks things. I'm not sure how to best handle this. Do we jump to 8.x-6.x for the next beta release? And make it clear that we jumped 5.x -> 6.x in release notes and that if you have Contributions in your Webforms configured you will need to go in and make edits?  

@colemanw @mattwire - how would you suggest we handle the transition? 

image

**2. Testing ->** re-configuring the webform quickly produces correct results - it works great! 👍 👍 
image

**3. Notices ->** can you please check to see if you get these? I'm getting a lot of these in the watchdog:
image

mattwire:
@KarinG What is it that breaks? And what's the fix - is it just opening the config and re-saving?

KarinG:
What breaks is that you can't process any contribution after installing the new code. I installed the new code and I could no longer make the contribution I made on Thursday during Lunch & Learn. 

The fix is to get into the Webform->CiviCRM tab and to configure things -> save it. So it's not just an opening and save.

jitendrapurohit:
Hi @KarinG Thanks for testing this quickly :). Breaking the existing forms should not be the case as there's already an upgrade code included in the PR. See - https://github.com/colemanw/webform_civicrm/pull/388/files#diff-3e39f4bed550c50fb87042565e62b09905cc381ef5b884c243a8f180f6e24d8c

I think what's required is to simply execute `drush up` after applying the PR. Do you mean you already did that it is still not working? Also, noting that the upgrade handling is done in the same function `webform_civicrm_update_8001()` as it was done for the email receipt PR - https://github.com/colemanw/webform_civicrm/pull/372.

KarinG:
Hi! @jitendrapurohit - yes I did remember to `drush updatedb` and it processed `webform_civicrm_update_8001()` - when I installed the previous 2 PRs on this topic. 

But after installing this PR:

I can not submit an existing Webform:
image

And the CiviCRM tab looks like -> 
image

KarinG:
So maybe it's the order in which I tested this - and `update_8001` won't (of course) re-run again...

jitendrapurohit:
yes, I think this will not be the case with others since update_8001 will run as a single upgrade package for both https://github.com/colemanw/webform_civicrm/pull/372 and this PR? So probably you should be able to roll back to 8000 and re-execute drush updb? Eg -

    drush ev "drupal_set_installed_schema_version('webform_civicrm', 8000)"

KarinG:
I’ll try that tomorrow 👍 I’ve put away my computer for the night and am chasing teenagers to bed.

jitendrapurohit:
Sure @KarinG. Have pushed the fix for e-notices you've mentioned in point3 at https://github.com/colemanw/webform_civicrm/pull/388#issuecomment-723642004.

KarinG:
Awesome - that worked 👍 

image

Ok merging this!

PR #389, opened 2020-11-10
Always allow case administrators to view cases

colemanw:
This permits logged-in users with 'access all cases and activities' permission to load cases based on an id in the url, regardless of what contacts are on the form.
Previously, cases could only be loaded if a case client or case manager was a known contact on the form.


KarinG:
That's an easy one! 👍 :-)

PR #390, opened 2020-11-11
Fix failing pipelines

mglaman:
For non-interactive installs, the `compile-mode` must be set to `all`.

Once that is fixed, we'll tackle the random failure on configuring CiviCRM for a Webform


mglaman:
🙄 go figure the PR passes without the random checkbox failure

mglaman:
@KarinG this fixes the tests failing due to the compile error on install, and it adds a safeguard for that random failure.

KarinG:
Happy dance! And all checks have passed on this PR :-)

mglaman:
Woo!

PR #391, opened 2020-11-13
Add Contribution configuration and test

mglaman:
Begins adding the test. Failing due to payment processor not showing up.

### DIFF
--- a/tests/src/FunctionalJavascript/ContributionPageTest.php
+++ b/tests/src/FunctionalJavascript/ContributionPageTest.php
@@ -107,8 +107,14 @@ final class ContributionPageTest extends WebformCivicrmTestBase {
     $this->getSession()->getPage()->fillField('Last Name', 'Pabst');
     $this->getSession()->getPage()->fillField('Email', 'fred@example.com');
     $this->getSession()->getPage()->pressButton('Next >');
-    // @todo submit contribution form.
+    // @todo need to figure out why the credit card form does not render.
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->getSession()->getPage()->fillField('Contribution Amount', '25.00');
+    $this->assertSession()->elementExists('css', '#wf-crm-billing-items');
+    $this->assertSession()->elementTextContains('css', '#wf-crm-billing-total', '25.00');
+    $this->getSession()->getPage()->pressButton('Submit');
     $this->htmlOutput();
+    $this->assertSession()->pageTextNotContains('Card Number field is required.');
   }
 
 }
--- a/tests/src/FunctionalJavascript/ContributionPageTest.php
+++ b/tests/src/FunctionalJavascript/ContributionPageTest.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Drupal\Tests\webform_civicrm\FunctionalJavascript;
+
+use Behat\Mink\Element\NodeElement;
+
+/**
+ * Tests submitting a Webform with a contribution page.
+ *
+ * @group webform_civicrm
+ */
+final class ContributionPageTest extends WebformCivicrmTestBase {
+
+  private function createPaymentProcessor() {
+    $params = [
+      'domain_id' => 1,
+      'name' => 'Dummy',
+      'payment_processor_type_id' => 'Dummy',
+      'is_test' => TRUE,
+      'is_active' => 1,
+      'is_default' => 1,
+      'user_name' => '',
+      'url_site' => 'http://dummy.com',
+      'url_recur' => 'http://dummy.com',
+      'billing_mode' => 1,
+      'sequential' => 1,
+      'payment_instrument_id' => 'Credit Card',
+    ];
+    $result = \wf_civicrm_api('payment_processor', 'create', $params);
+    $this->assertEquals(0, $result['is_error']);
+    $this->assertEquals(1, $result['count']);
+    return $result['values'][0];
+  }
+
+  public function testSubmitContribution() {
+    // $payment_processor['id'];
+    $payment_processor = $this->createPaymentProcessor();
+    $this->drupalLogin($this->adminUser);
+    $this->drupalGet($this->webform->toUrl('settings'));
+    $this->getSession()->getPage()->clickLink('CiviCRM');
+    // The label has a <div> in it which can cause weird failures here.
+    $this->assertSession()->waitForText('Enable CiviCRM Processing');
+    $this->getSession()->getPage()->checkField('nid');
+    $this->getSession()->getPage()->clickLink('Contribution');
+    $this->getSession()->getPage()->selectFieldOption('civicrm_1_contribution_1_contribution_enable_contribution', 1);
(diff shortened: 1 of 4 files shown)


mglaman:
The test is currently failing with:

```
Notice: Undefined index: civicrm_1_contribution_1_contribution_total_amount in wf_crm_webform_preprocess->displayLineItems() (line 565 of modules/contrib/webform_civicrm/includes/wf_crm_webform_preprocess.inc).
```

This is due to the way the webform elements are added. We create `contribution_pagebreak` as a wizard page, but the form is broken because the contact elements are not in a wizard page. This is part of the manual configuration process.

Instead of fixing that, the test should do the configuration. Then later one we can fix the bug.

Screen Shot 2020-11-15 at 10 40 40 PM

KarinG:
Great progress! 👍

mglaman:
webform_civicrm_contribution_form_payment

😞 need to figure out why the credit card form will not render.

KarinG:
Oh... is this in the Functional Test where it fails? Checking in UI itself:

My notes:

1. Latest code from Github - PR to refactor the Payment Processors has db updates!
2. Create Payment Processor Dummy in CiviCRM
http://d9civicrm.local/civicrm/admin/paymentProcessor?reset=1
3. Webform -> CiviCRM Tab -> Contribution? -> Enable Contribution checkbox Yes, then checkbox Contribution amount, Payment Processor -> select Dummy; select Financial Type -> Donation (anything really) -> select Currency -> USD$
4. Remember the Wizard pages.

image

mglaman:
@KarinG yeah, the form should be set up just fine. Going to have the test take more screenshots. I'll try running Chrome in windows/non-headless mode. I need to see why the credit card form for the dummy payment processor that is created will not load.

mglaman:
Here's some screenshots. It all looks well. I need to see if I can capture console logs.

webform_civicrm_contribution_settings
webform_civicrm_contribution_build
webform_civicrm_contribution_form_payment

I'll try adding a longer pause before checking for the credit card form

```
    // @todo need to figure out why the credit card form does not render.
    $this->getSession()->getPage()->fillField('Contribution Amount', '25.00');
    $this->createScreenshot($this->htmlOutputDirectory . '/webform_civicrm_contribution_form_payment.png');
```

Because the contribution table updates. The credit card form just didn't pop in.

mglaman:
The mount is in the markup, at least

```
<div class="crm-container crm-public" id="billing-payment-block"></div><input autocomplete="off" data-drupal-selector="form-35rynd-5jeyovradd-vq83-tvc-2irqgacpfmaq17mk" type="hidden" name="form_build_id" value="form-35rYND_5jeYOVrADD_VQ83_TVC-2IrQGACpFMAQ17mk" />
```

KarinG:
Ah yes -> so only when there is an amount -> the credit card fields will show up using javascript (so that takes time).

Alternatively to avoid this for testing -> perhaps set a default amount on that Contribution Amount form element? Then the Credit Card fields should be pretty much there...

mglaman:
Ah, ha! I found why (well, not the fix "why" but why it isn't loading)

In the `webform_civicrm_payment.js`:

The following is part of the setup.
```
    setting = drupalSettings.webform_civicrm,
    $processorFields = $('[name$="civicrm_1_contribution_1_contribution_payment_processor_id"]');
```

The billing block is loaded through

```
var type = getPaymentProcessor();
    if (type && type !== '0') {
```

The payment processor is fetch here:

```
  function getPaymentProcessor() {
    if (!$processorFields.length) {
      return setting.paymentProcessor;
    }
    return $processorFields.filter('select, :checked').val();
  }
```

When I run the following in the browser output, there are no fields:

```
document.querySelectorAll('[name$="civicrm_1_contribution_1_contribution_payment_processor_id"]');
```

That means `$processorFields.length` is empty, so it doesn't do anything. I also cannot find it in the markup. So! If we get that element to render during the test, it should be fixed.

mglaman:
The good news is this:

Here is the data from the Drupal settings:

```
{
    "processor_id_key": "processor_id",
    "contributionCallback": "/civicrm/payment/form?currency=USD&snippet=4&is_drupal_webform=1",
    "paymentProcessor": 1
}
```

That means wf_crm_webform_preprocess::addResources is working

```
      // Add payment processor - note we have to search in 2 places because $this->loadMultipageData hasn't been run. Maybe it should be?
      $fid = 'civicrm_1_contribution_1_contribution_payment_processor_id';
      if (!empty($this->enabled[$fid])) {
        $js_vars['paymentProcessor'] = $this->getData($fid);
        // @todo Why does it matter if its enabled or not? Nothing is submitted.
        // $js_vars['paymentProcessor'] = wf_crm_aval($this->form_state, 'storage:submitted:' . $this->enabled[$fid]);
      }
      else {
        $js_vars['paymentProcessor'] = $this->getData($fid);
      }
```

mglaman:
Yeah, in the test the `contribution_payment_processor_id` doesn't seem to be created. I just realized that `$js_vars['paymentProcessor'] = $this->getData($fid);` runs regardless if it is enabled or not. That logic is a carryover from D7 where multipage worked differently. It's like the element is configured and respected everywhere on the data side but not by the build.

tests/src/FunctionalJavascript/ContributionPageTest.php
```diff
@@ -0,0 +1,140 @@
...
+    $this->getSession()->getPage()->selectFieldOption('credit_card_exp_date[Y]', '2023');
+    $this->getSession()->getPage()->fillField('Billing First Name', 'Frederick');
+    $this->getSession()->getPage()->fillField('Billing Last Name', 'Pabst');
+    $this->getSession()->getPage()->fillField('Street Address', '123 Milwaukee Ave');
+    $this->getSession()->getPage()->fillField('City', 'Milwaukee');
+    $this->getSession()->getPage()->fillField('Country', '1228');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    // @todo find a better way to wait for select2.
+    sleep(1);
+    $this->getSession()->getPage()->fillField('State/Province', '1048');
```
mglaman: This doesn't work because Select2 hides the actual element. This needs to be changed to interact with Select2 properly.

mglaman:
@KarinG okay! it's finally ready for review.

mglaman:
Joy. Random failures, will investigate.

KarinG:
🙃

mglaman:
Okay, it's failing on the following because the Country field is also a Select2 element and stuck open

```
Exception: element click intercepted: Element <a href="javascript:void(0)" class="select2-choice select2-default" tabindex="-1">...</a> is not clickable at point (390, 769). Other element would receive the click: <div id="select2-drop-mask" class="select2-drop-mask" style=""></div>
```

mglaman:
And I see why: locally the Country select element isn't in Select2, but in the CI it is.

mglaman:
WOOOOOO GREEN!

KarinG:
I'm just going to screenshot this for posterity :-)

image

PR #392, opened 2020-11-16
Set case role (relationship) start date when creating case role, update existing active case role (relationship) if it already exists

mattwire:
Overview
Fix creating/updating case roles (relationships).

See related https://github.com/colemanw/webform_civicrm/pull/332

D7 or D8?
D8 (See #393 for D7).

Before
Error in relationship.create API call if a "duplicate" relationship exists (that could be active or inactive).

After
If an active relationship already exists with the same params update it (we probably don't actually need to update because I don't think there are any changes but this feels safest and means that hooks etc will continue to fire.

If an active relationship does not already exist (an inactive one might but we don't want to reactivate that because it preserves history) then we create a new relationship (case role) with the start_date set to today.

Previously we were not setting the start_date.

Technical Details
Check for existing relationships and add start_date to new ones.

Comments

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1350,16 +1350,31 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $this->handleEntityTags('case', $result['id'], $n, $params);
           // Store id
           $this->ent['case'][$n]['id'] = $result['id'];
+          $relationshipStartDate = date('Ymd');
           // Save case roles
           foreach ($params as $param => $val) {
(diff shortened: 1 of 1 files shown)


KarinG:
I've asked Olly (who I know uses case) to take a look at this PR.

gibsonoliver:
Hi All

I can't see any issues with this idea.
As long as the start date on an existing active relationship isn't updated.

Olly

mattwire:
@gibsonoliver That's correct, we only set start_date on new relationships

KarinG:
👍 Thank you both! Merging.

PR #393, opened 2020-11-16
7.x Set case role (relationship) start date when creating case role, update existing active case role (relationship) if it already exists

mattwire:
Overview
Port of #392

D7 or D8?
D7


KarinG:
@colemanw - I merged this for 8.x-5.x https://github.com/colemanw/webform_civicrm/pull/392 after checking with Olly for feedback on the case bit. I think this can go into 7.x-5.x as well.

PR #394, opened 2020-11-17
Fixes E-notice Undefined index

stesi561:
Overview
Fixes an error notice that was clogging up logs.

Before
Drupal watchdog logs filled up with annoying error notices.

After
Drupal watchdog logs fill up with other annoying error notices. But at least this one no longer causes pain.

Technical Details

Extracted part of the logic from the IF into a variable and test for existence of array keys before referencing them.

### DIFF
--- a/includes/wf_crm_webform_base.inc
+++ b/includes/wf_crm_webform_base.inc
@@ -357,9 +357,10 @@ abstract class wf_crm_webform_base {
     foreach ($settingsArray[$ent] as $setting) {
       $valueFound = false;
       foreach($values as $key => $value){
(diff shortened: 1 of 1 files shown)


KarinG:
Hi Luke - have you checked what this looks like in the D8 branch?

stesi561:
It doesn't quite apply cleanly due to change between array syntax. But appart from that is the same in D8. Which potentially this should be using array() instead of [].

Do you want me to reformat this to use array() instead of [], and also create something for D8 as well?

KarinG:
Hi Luke - can you please first submit a PR for D8 -> we have some automated testing now! Once merged we'll backport it to D7.

PR #395, opened 2020-11-17
Case relationship not expired if it is closed via webform submission

jitendrapurohit:
Overview
Case relationship end date is not set when the case is closed.

Before
Case Relationship still active after case is closed.

After
End Date is saved for the Case Relationship if it is closed.


mattwire:
@jitendrapurohit This is related to and partly overlaps #392 #393. What does CiviCRM core do when you close a case?

jitendrapurohit:
Ah yes,  I can rebase the PR if needed. Civi core ends all the relationship when the case is closed. So this is only an issue with w_c.

PR #398, opened 2020-12-02
fix radio element line item calculation

jitendrapurohit:
Overview
Fixes https://www.drupal.org/project/webform_civicrm/issues/3183061

Before
Contribution Amount if displayed as radio, does not display the calculated amount in the line item section.

image

After
The calculated value is displayed correctly. Tested with Membership Fee, contribution amount and 2 line items amount.

image

Technical Details
`form-radios` is the class for the div which holds all the radio button. This change avoids the js call to be made from div and instead triggers the js function for the exact input radio element.

Comments
ping @KarinG

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -118,7 +118,7 @@ cj(function($) {
     updateLineItem(lineKey, amount, label);
   }
 
-  $('.civicrm-enabled.contribution-line-item')
+  $('.civicrm-enabled.contribution-line-item:not(".form-radios")')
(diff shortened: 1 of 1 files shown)


KarinG:
@colemanw - can you please have a quick look at this one-line javascript fix?

@jitendrapurohit 👍

colemanw:
LGTM

KarinG:
@jitendrapurohit - could you please have a look if this needs porting to 8.x-5.x?

KarinG:
Adding needs port label to ensure we remember to check if it requires a D8 port.

PR #400, opened 2020-12-04
D8: fix unported D7-style error handling on 'Existing Contact' widget

MegaphoneJon:
Overview
When "Existing Contact" triggers this warning message, it breaks validation of all future elements.  By fixing the error handling, validation of other elements works normally.

Before
Undefined call to watchdog.

After
Watchdog and on-screen message display correctly.

Technical Details
I don't know why `wf_crm_admin_access()` no longer exists.  I'm assuming there's a good reason, but it was a one-line function, so I simply put the D8 equivalent directly inline.

This is my first time looking at `webform_civicrm`'s code - if someone tells me that this is the correct approach, I can also replace the other remaining instances of this function.

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -162,19 +162,15 @@ class ContactComponent implements ContactComponentInterface {
         }
       }
       // If we get exactly $limit results, there are probably more - warn that the list is truncated
-      // TODO Error reporting is still Drupal7-ish
       if (wf_crm_aval($result, 'count') >= $limit) {
-        watchdog(
-          'webform_civicrm',
-          'Maximum contacts exceeded, list truncated on the webform "@title". The webform_civicrm "@field" field cannot display more than !limit contacts because it is a select list. Recommend switching to autocomplete widget in component settings.',
-          array('!limit' => $limit, '@field' => $component['name'], '@title' => $node->title),
-          WATCHDOG_WARNING,
-          l(t('Edit component'), "node/{$node->nid}/webform/components/{$component['cid']}")
+        \Drupal::logger('webform_civicrm')->warning(
+          'Maximum contacts exceeded, list truncated on the webform "@title". The webform_civicrm "@field" field cannot display more than @limit contacts because it is a select list. Recommend switching to autocomplete widget in component settings.',
+          array('@limit' => $limit, '@field' => $component['#title'], '@title' => $node->label()),
         );
-        if (wf_crm_admin_access($node) && node_is_page($node)) {
-          \Drupal::messenger()->addWarning('<strong>' . t('Maximum contacts exceeded, list truncated.') .'</strong><br>' .
-          t('The field "@field" cannot show more than !limit contacts because it is a select list. Recommend switching to autocomplete widget in <a !link>component settings</a>.',
-            array('!limit' => $limit, '@field' => $component['name'], '!link' => 'href="' . url("node/{$node->nid}/webform/components/{$component['cid']}") . '"')));
+        if ($node->access('update') && \Drupal::currentUser()->hasPermission('access CiviCRM')) {
+          $warning_message = \Drupal\Core\Render\Markup::create('<strong>' . t('Maximum contacts exceeded, list truncated.') .'</strong><br>' .
+          t('The field "@field" cannot show more than @limit contacts because it is a select list. Recommend switching to autocomplete widget.', ['@limit' => $limit, '@field' => $component['#title']]));
+          \Drupal::messenger()->addMessage($warning_message);
         }
       }
     }


KarinG:
Jon - thank you! I just grep-ed and I think there is one more instance. 

I wonder why the automated tests did not run on your PR - going to see if I can kickstart them.

MegaphoneJon:
I've since discovered that this code is being triggered when it shouldn't - it should only be triggered when the "Existing Contact" widget is a select list, which mine aren't.  However, it's no longer breaking the entire form - it's just posting a notification that's incorrect, which is a lower priority for me right now.

KarinG:
Sounds good - I understand 👍

KarinG:
@mglaman - is there a way for me to start Github Actions on this PR?

mglaman:
🤔  I don't get why it didn't automatically run. I'll check in a little bit

src/ContactComponent.php
```diff
@@ -162,19 +162,15 @@ function wf_crm_contact_search($node, $component, $params, $contacts, $str = NUL
...
+        \Drupal::logger('webform_civicrm')->warning(
+          'Maximum contacts exceeded, list truncated on the webform "@title". The webform_civicrm "@field" field cannot display more than @limit contacts because it is a select list. Recommend switching to autocomplete widget in component settings.',
+          array('@limit' => $limit, '@field' => $component['#title'], '@title' => $node->label()),
         );
-        if (wf_crm_admin_access($node) && node_is_page($node)) {
-          \Drupal::messenger()->addWarning('<strong>' . t('Maximum contacts exceeded, list truncated.') .'</strong><br>' .
-          t('The field "@field" cannot show more than !limit contacts because it is a select list. Recommend switching to autocomplete widget in <a !link>component settings</a>.',
-            array('!limit' => $limit, '@field' => $component['name'], '!link' => 'href="' . url("node/{$node->nid}/webform/components/{$component['cid']}") . '"')));
+        if ($node->access('update') && \Drupal::currentUser()->hasPermission('access CiviCRM')) {
+          $warning_message = \Drupal\Core\Render\Markup::create('<strong>' . t('Maximum contacts exceeded, list truncated.') .'</strong><br>' .
```
mglaman, line 171: I don't _think_ `Markup::create` is required for `<strong>`.

src/ContactComponent.php
```diff
@@ -162,19 +162,15 @@ function wf_crm_contact_search($node, $component, $params, $contacts, $str = NUL
...
-          'webform_civicrm',
-          'Maximum contacts exceeded, list truncated on the webform "@title". The webform_civicrm "@field" field cannot display more than !limit contacts because it is a select list. Recommend switching to autocomplete widget in component settings.',
-          array('!limit' => $limit, '@field' => $component['name'], '@title' => $node->title),
-          WATCHDOG_WARNING,
-          l(t('Edit component'), "node/{$node->nid}/webform/components/{$component['cid']}")
+        \Drupal::logger('webform_civicrm')->warning(
+          'Maximum contacts exceeded, list truncated on the webform "@title". The webform_civicrm "@field" field cannot display more than @limit contacts because it is a select list. Recommend switching to autocomplete widget in component settings.',
+          array('@limit' => $limit, '@field' => $component['#title'], '@title' => $node->label()),
         );
-        if (wf_crm_admin_access($node) && node_is_page($node)) {
```
mglaman, line 174: Verified admin_access is this

```
/**
 * Access callback to determine if user can see the CiviCRM tab of a webform.
 *
 * @param object $node
 * @return bool
 */
function wf_crm_admin_access($node) {
  return (node_access('update', $node) && user_access('access CiviCRM'));
}
```

I thought I left functions like this in `webform_civicrm.module` during the report so we wouldn't have breaks like this :(. Sorry!

mglaman (approved):
Looks good to me. Just a little nit on the maybe unneeded usage of `Markup::create`.

Markup just wraps a string to bypass XSS filtering – which isn't needed

```
* Objects that implement this interface will not be automatically XSS filtered
 * by the render system or automatically escaped by the theme engine.
```

mglaman:
@KarinG can you see if you can manually run the workflow from the Actions tab? Maybe?

I don't get why Actions didn't run – they didn't even run on the fork: https://github.com/MegaphoneJon/webform_civicrm/actions and they always do.

mglaman:
Yeah, I'm at a loss for why it didn't run. I just pushed to my fork and it executed: https://github.com/mglaman/webform_civicrm/actions/runs/410754057. 

We could see if a new commit triggers the workflow.

src/ContactComponent.php
MegaphoneJon: Hey Matt - if I don't use `Markup::create` the `<strong>` and `<br>` are rendered literally.  So I'm not sure if I should leave in `Markup::create` or you're suggesting there's a more lightweight `MarkupInterface` I should be using instead.

mglaman: Oh! Nevermind then :)

KarinG: Ok resolved then! 👍

PR #401, opened 2020-12-05
D8 - Add Pay Later Receipt message to receipts

jitendrapurohit:
Overview
An extension of #372. Allow webform to set pay later receipt message on the settings page.

Before
As contribution page dependency is now removed, there is no way to set custom text for the pay later contributions.

After
Added pay later receipt text on the settings page -

Screenshot 2020-12-05 at 3 40 27 PM

Comments
Needs https://github.com/civicrm/civicrm-core/pull/19129 PR in core to be applied for this to work as expected. @KarinG Seems like this was pending since we got https://github.com/colemanw/webform_civicrm/pull/372 merged?


KarinG:
Yes - excellent 👍 I'll add a comment to the core PR.

PR #399, opened 2020-12-03
Add build matrix for D8 and D9 testing

mglaman:
Adds a build matrix for ^8.9 && Webform 5.x and ^9.0 && Webform 6.x

PR #402, opened 2020-12-08
Remove civicrm-core patch for test support

mglaman:
I'm fairly certain with 5.32 we can now remove this patch.


KarinG:
Closing in favour of https://github.com/colemanw/webform_civicrm/pull/399

PR #403, opened 2020-12-08
Fix `civicrm_` table cleanup between tests

mglaman:
It turns out all `civicrm_` tables remained between tests. This ensures they are dropped after a test completes, so that they can be recreated during the next test.


KarinG:
Closing in favour of #399

PR #405, opened 2020-12-09
Use prefer-dist to prevent civicrm-core from running in development mode

mglaman:
🥳 This fixes failures when testing beta/RC CiviCRM releases.


KarinG:
That's awesome 🥳 - Confirms the issue was indeed what we arrived at yesterday.

PR #406, opened 2020-12-12
Activities fix for default_value + activity_date_time + Add test

KarinG:
Before
image

After
Can successfully create Activities -> 
image

Technical Details
In the process also fixed some error handling.

### DIFF
--- a/includes/utils.inc
+++ b/includes/utils.inc
@@ -547,14 +547,14 @@ function wf_civicrm_api($entity, $operation, $params) {
     }
     $stop = null;
     \Drupal::logger('webform_civicrm')->error(
-      'The CiviCRM ":function" API returned the error: ":msg" when called by function ":fn" on line :line of :file with parameters: ":params"',
+      'The CiviCRM "@function" API returned the error: "@msg" when called by function "@fn" on line @line of @file with parameters: "@params"',
       array(
-        ':function' => $entity . ' ' . $operation,
-        ':msg' => $result['error_message'],
-        ':fn' => $bt[$n+1]['function'],
-        ':line' => $bt[$n]['line'],
-        ':file' => array_pop($file),
-        ':params' => print_r($params, TRUE),
+        '@function' => $entity . ' ' . $operation,
+        '@msg' => $result['error_message'],
+        '@fn' => $bt[$n+1]['function'],
+        '@line' => $bt[$n]['line'],
+        '@file' => array_pop($file),
+        '@params' => print_r($params, TRUE),
       )
     );
   }
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2414,7 +2414,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $val = array_search($val, $options);
         }
         // Fudge together date and time fields
-        if ($field['type'] === 'time' && substr($name, -8) === 'timepart') {
+        if (($field['type'] === 'time' || $field['type'] === 'webform_time') && substr($name, -8) === 'timepart') {
           $name = str_replace('_timepart', '', $name);
           // Add date (default to today)
           $date = wf_crm_aval($this->data, "$ent:$c:$table:$n:$name", date('Ymd'));
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -428,23 +428,21 @@ class Fields implements FieldsInterface {
       $fields['activity_activity_date_time'] = array(
         'name' => t('Activity # Date'),
         'type' => 'date',
-        'value' => 'now',
+        'default_value' => 'now',
       );
       $fields['activity_activity_date_time_timepart'] = array(
         'name' => t('Activity # Time'),
-        'type' => 'webform_time',
-        'value' => 'now',
+        'type' => 'time',
+        'default_value' => 'now',
       );
       $fields['activity_duration'] = array(
         'name' => t('Activity # Duration'),
         'type' => 'number',
-        'extra' => array(
-          'field_suffix' => t('min.'),
-          'min' => 0,
-          'step' => 5,
-          'integer' => 1,
-        ),
-      );
+        'field_suffix' =>  t('min.'),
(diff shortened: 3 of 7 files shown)


includes/utils.inc
```diff
@@ -547,14 +547,14 @@ function wf_civicrm_api($entity, $operation, $params) {
...
-        ':params' => print_r($params, TRUE),
+        '@function' => $entity . ' ' . $operation,
+        '@msg' => $result['error_message'],
+        '@fn' => $bt[$n+1]['function'],
+        '@line' => $bt[$n]['line'],
+        '@file' => array_pop($file),
+        '@params' => print_r($params, TRUE),
       )
     );
   }
```
KarinG, line 560: Came across this by accident. Fixed in the process.

src/Fields.php
```diff
@@ -428,23 +428,21 @@ protected function wf_crm_get_fields($var = 'fields') {
...
-        'value' => 'now',
+        'default_value' => 'now',
       );
       $fields['activity_activity_date_time_timepart'] = array(
         'name' => t('Activity # Time'),
-        'type' => 'webform_time',
-        'value' => 'now',
+        'type' => 'time',
+        'default_value' => 'now',
       );
```
KarinG, line 438: Critical bits are 'default_value' is the proper key and type should be 'time' -> not 'webform_time'. When it's 'time' -> the webform civicrm will find it and glue it to the date to create a proper activity date time stamp.

src/Fields.php
```diff
@@ -428,23 +428,21 @@ protected function wf_crm_get_fields($var = 'fields') {
...
-          'step' => 5,
-          'integer' => 1,
-        ),
-      );
+        'field_suffix' =>  t('min.'),
+        /*ToDo Figure out why setting min does not work!*/
+        'min' => 0,
+        'step' => 5,
+       );
       $tag_entities = array('other', 'activity');
```
KarinG, line 447: I pulled these out of the extras to get them working. We'll 'field_suffix' and 'step' work - but for some reason 'min' does not. Not sure why that is.

src/Plugin/WebformElement/CivicrmContact.php
```diff
@@ -132,7 +132,7 @@ public function prepare(array &$element, WebformSubmissionInterface $webform_sub
         }
       }
     }
-    if (empty($cid) && $element['#type'] === 'hidden' && $element['none_prompt']) {
+    if (empty($cid) && $element['#type'] === 'hidden' && isset($element['none_prompt'])) {
       $element['#attributes']['data-civicrm-name'] = Xss::filter($element['none_prompt']);
     }
```
KarinG, line 137: Found this in the dblog -> 'index undefined' - fixed!

src/Fields.php
```diff
@@ -934,12 +932,15 @@ protected function wf_crm_get_fields($var = 'fields') {
...
+      $custom_fields = [];
+      if (count($custom_sets) > 0) {
+        $custom_fields = wf_crm_apivalues('CustomField', 'get', [
+          'is_active' => 1,
+          'custom_group_id' => ['IN' => array_keys($custom_sets)],
+          'html_type' => ['IN' => array_keys($custom_types)],
+          'options' => ['sort' => 'weight'],
+        ]);
+      }
       foreach ($custom_fields as $custom_field) {
```
KarinG, line 945: Ran into this in the dblog (this is a webform that has no custom fields on it -> so custom_group_id was empty) -> 

`The CiviCRM "CustomField get" API returned the error: "invalid criteria for IN" when called by function "wf_crm_get_fields" on line 937 of Fields.php with parameters: "Array ( [is_active] => 1 [custom_group_id] => Array ( [IN] => Array ( ) ) [html_type] => Array ( [IN] => Array ( [0] => Select [1] => Multi-Select [2] => Radio [3] => CheckBox [4] => Text [5] => TextArea [6] => RichTextEditor [7] => Select Date [8] => Link [9] => Select Country [10] => Multi-Select Country [11] => Select State/Province [12] => Multi-Select State/Province [13] => Autocomplete-Select [14] => File ) ) [options] => Array ( [sort] => weight [limit] => 0 ) [check_permissions] => [version] => 3 ) "`

Fixed by counting to ensure we have some before we go and retrieve them.

KarinG:
Going to try and add a test in a separate PR!

src/Fields.php
KarinG: Actually I was wrong on that!  Reverted back to webform_time.

includes/wf_crm_webform_postprocess.inc
```diff
@@ -2414,7 +2414,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $val = array_search($val, $options);
         }
         // Fudge together date and time fields
-        if ($field['type'] === 'time' && substr($name, -8) === 'timepart') {
+        if (($field['type'] === 'time' || $field['type'] === 'webform_time') && substr($name, -8) === 'timepart') {
```
KarinG, line 2417: Fixing the glue -> in D8 some elements are of type `webform_time` -> so make sure they also are found when date and time fields are glued together!

KarinG:
PR #407 reproduced the Error in a Test.

KarinG:
Github Action version of that validation error screenshot:
image

Ok so the defaults are population properly now: not sure why it failed...

image

Going for a re-run.

KarinG:
And we are green! ✅

I've commented out - for now:
```
    // ToDo -> figure out what Error message it is! The submission itself works well.
    // $this->assertPageNoErrorMessages();
```
Because the test continues -> the webform is submitted properly and the activity is stored in CiviCRM.

KarinG:
Doing some live testing -> given a URL with cid1 and aid -> it's not pulling in Activity Date. Need to look at that!

image

KarinG:
Ok loading Activity Date now:

image

jitendrapurohit:
Can replicate the bug with the activity date time field and confirm that the above change fixes the problem. Though it required to re-add the activity fields (date and time). 

Thanks for fixing the additional bugs in the DBlog @KarinG 👍

PR #407, opened 2020-12-12
Activity test

KarinG:
First bits for an Activity Test!


KarinG:
It fails to set the defaults for Activity Date and Activity Time -> and that makes sense as that's exactly what I fixed that in https://github.com/colemanw/webform_civicrm/pull/406 - still to be merged.

image

Going to add this to #406

PR #410, opened 2020-12-18
Record line_item amounts without rounding.

agileware-fj:
Overview
Removes the rounding that is applied to line item amounts during the contribution processing. This fixes some tax round errors that have already been addressed in CiviCRM core by avoiding premature rounding.
This is a port of #409.

Before
When a contribution is created from a webform submission, the line item total amount, unit price, and tax amount are rounded.  This can result in a 1c difference between the line item amount and the contribution total per line item.

Example:
- 1 x Membership with at $704.5454
- 10% tax is added ($70.45454)
- Contribution records this exactly, comes out to ~$775.00 after final rounding
- Line item records $704.55 + $70.46† comes to ~$775.01
- Contribution total and line item total *do not match*

† I think it's already rounded the total amount once, so I guess more work might be required here.

One place this causes issues is when you're recording a payment against a "pay later" contribution created in this way, it "Balanced owed" is the line item amount, not the contribution amount.  If you record the contribution amount, the contribution is marked as "Partially paid."

After
Amounts are not rounded when creating the line item.  As a result, the contribution amount and line item total / balance owed match.
Actual rounding is deferred to CiviCRM to keep precision for as long as possible.

Technical Details
Removes the calls to CRM_Utils_Money::format entirely. These were only used for rounding purposes as far as I could tell.

Comments
Agileware ref CIVIWF-5

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -2207,11 +2207,6 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
         }
       }
 
-      // Format (and round) monetary values before saving the line items
-      isset($item['unit_price']) ? $item['unit_price'] = CRM_Utils_Money::format($item['unit_price'], NULL, NULL, TRUE) : NULL;
-      isset($item['line_total']) ? $item['line_total'] = CRM_Utils_Money::format($item['line_total'], NULL, NULL, TRUE) : NULL;
-      isset($item['tax_amount']) ? $item['tax_amount'] = CRM_Utils_Money::format($item['tax_amount'], NULL, NULL, TRUE) : NULL;
-
       // Save the line_item
       $line_result = wf_civicrm_api('line_item', 'create', $item);
       $item['id'] = $line_result['id'];


agileware-fj:
Note: we don't have any Drupal 8 CiviCRM sites, so I'm relying on the testbots for this one at this point.

KarinG:
👍 I will look to add a test for line items to this PR.

KarinG:
I wonder where the 10% tax that is added ($70.45454) -> become $70.46? That's a round up. 

I don't have an easy way or time right now to add Sales Tax into the tests [but definitively on the ToDo list], but I did add Line Items -> and used your amounts:

```
    $this->getSession()->getPage()->fillField('Line Item Amount', '704.5454');
    $this->getSession()->getPage()->fillField('Line Item Amount 2', '70.45454');

```

With and without this PR - these line totals save as:

```
    $this->assertEquals('704.55', $api_result['values'][1]['line_total']);
    $this->assertEquals('70.45', $api_result['values'][2]['line_total']);
```

So something else must be rounding it after we send it over to CiviCRM using the API?

KarinG:
Line items tests added in https://github.com/colemanw/webform_civicrm/pull/411

agileware-fj:
@KarinG The $70.46 - I think the example might not be the best one, it's what I was testing at the time, but the problem holds true anyway.  When I re-run my calculations outside of the CRM, I have to round the total amount before calculating the tax amount, then round that again.

Perhaps use 45.4545 (should come out to ~$50 after 10% tax added) instead?  It's been a more reliable problem over time.

> `$this->assertEquals('704.55', $api_result['values'][1]['line_total']);`
> `$this->assertEquals('70.45', $api_result['values'][2]['line_total']);`

Returning true both before and after – I wonder if the API is rounding for the return?

KarinG:
Ah... "I wonder if the API is rounding for the return?" -> Then how would we be able to test it?

KarinG:
I'll rerun that test locally with the 45.4545 and will also stop it before it does the teardown so that I can check the line item table.

KarinG:
Actually the amounts won't matter make a difference as I don't have a tax_amount in the test yet. 

I'm going to check on the line item table though ->  with your PR -> I still get a rounded value so it's not the retrieval. There is some rounding somewhere else 🤔

image

KarinG:
Merging this - I have a Line Item test (with Sales Taxable items working). The test however can not ensure there is no rounding - because even after we remove the rounding (this PR) - > something else rounds the line item values.

KarinG:
image

tax_amount was $1.475 at the time I save this, but is $1.48 in the screenshot 👆

PR #415, opened 2020-12-21
Patch 6

ghost:
Overview
The Membership Renewal Activity requires the source_record_id field to be set to the id of the Membership concerned.

If this is not included, then the 'view' action link in any Activity search has a blank 'id=' query parameter and throws a fatal error.

Before
Membership Renewal Activities create by the Webform throw a fatal error when the 'view' link is clicked in an Activity search. E.g. the list of Activities displayed under the Activities tab of a Contact Summary screen.

After
Works correctly.


ghost:
Created incorrectly.

PR #417, opened 2020-12-22
Add Sales Tax testing to Contribution test

KarinG:
Overview
Adding Sales Tax test to the ContributionPage test.

Before
Testing Contribution Amount + Line Items

After
Testing Contribution Amount + Line Items (one with Sales Tax)

Technical Details
Copied over https://github.com/civicrm/civicrm-core/blob/master/tests/phpunit/CiviTest/CiviUnitTestCase.php#L3104

Because I think we would like to be able to pass in Sales Tax Rate (it's hardcoded in that CiviCRM function), so that we can calculate what the expected values should be.

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -594,7 +594,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
         $taxSettings = wf_crm_get_civi_setting('contribution_invoice_settings');
 
         if (($itemTaxRate !== 0) && ($taxSettings['tax_display_settings'] !== 'Do_not_show')) {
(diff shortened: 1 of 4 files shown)


KarinG:
Ok got the basics working -> too many taxes though! 

image

And found `!rate` issue -> will fix this in the process
`User error: Invalid placeholder (!rate) in string: includes !rate @tax in Drupal\Component\Render\FormattableMarkup::placeholderFormat() (line 241 of core/lib/Drupal/Component/Render/FormattableMarkup.php).`

KarinG:
Ok this is what I want:

image

KarinG:
And ✅ for 5/6 -> going to re-run the jobs.

KarinG:
✅ Happy dance!

KarinG:
Merging this as the only code fix here was a string replacement 🙂

PR #419, opened 2020-12-23
doPayment throws exceptions when payment fails. Catch and display to the user instead of an internal server error

mattwire:
Overview
I found this issue when submitting a payment via omnipay sagepay and failing postcode validation (which is only done once the call to doPayment has happened).  `doPayment` will throw an exception when something goes wrong and it should provide a user friendly error message that we can display to the user. Currently the exception is not trapped and the user sees "internal server error".

Before
User sees "internal server error" on payment failure.

After
User sees error message and is bounced back to form on payment failure.

Technical Details
described above.

Comments


KarinG:
Thanks @mattwire - and tests are ✅ (covering Contributions, Line Items and Sales Tax now).

PR #420, opened 2020-12-23
Remove doTransferCheckout

mattwire:
Overview
Remove deprecated payment function - which is called automatically by CiviCRM call if necessary anyway.

Before
Call to deprecated function if available.

After
Just call doPayment.

Technical Details

Comments
Follow up to #419 - I'd like to backport this one with both commits once approved/merged.


KarinG:
Thanks for this too Matt!

PR #421, opened 2020-12-23
Fix javascript error due to incorrect path.

KarinG:
Overview
Raised in: https://www.drupal.org/project/webform_civicrm/issues/3182288
by Makemeweb and fkohrt

The library is added here:
https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/webform_civicrm.libraries.yml#L31

Then we're removing it here :-)
https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/webform_civicrm.module#L33

And we re-add it using the `$user_framework_resource_url` - which is incorrect and we error.
https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/webform_civicrm.module#L34

Before
image

After
No JS errors

Technical Details
`$user_framework_resource_url` -> includes core/ whereas the packages reside next to core/ (not underneath it).

Alternatively -> we could also decide to remove this altogether and just let the admin handle a non-typical install by editing `webform_civicrm.libraries.yml` ?

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -32,8 +32,8 @@ function webform_civicrm_library_info_alter(array &$libraries, $extension) {
     $user_framework_resource_url = CRM_Core_Config::singleton()->resourceBase;
     // Unset -> https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/webform_civicrm.libraries.yml#L31
     unset($libraries['civicrm_contact']['js']['/libraries/civicrm/packages/jquery/plugins/jquery.tokeninput.js']);
-    if (substr($user_framework_resource_url,-5) === 'core/') {
-      $libraries['civicrm_contact']['js'][substr($user_framework_resource_url, -5) . 'packages/jquery/plugins/jquery.tokeninput.js'] = [
+    if (substr($user_framework_resource_url, -5) === 'core/') {
+      $libraries['civicrm_contact']['js'][substr($user_framework_resource_url, 0,-5) . 'packages/jquery/plugins/jquery.tokeninput.js'] = [
         'preprocess' => FALSE,
         'minified' => FALSE
       ];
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Hi Karin, we're working on some CiviCRM Core updates that would load CiviCRM Core css/js properly, via Drupal 8 Libraries API (not the module, the concept)
It's very tricky but we've got a promising proof of concept. More on that next year!

webform_civicrm.module
```diff
@@ -31,10 +31,12 @@ function webform_civicrm_library_info_alter(array &$libraries, $extension) {
     \Drupal::service('civicrm')->initialize();
     $user_framework_resource_url = CRM_Core_Config::singleton()->resourceBase;
     unset($libraries['civicrm_contact']['js']['/libraries/civicrm/packages/jquery/plugins/jquery.tokeninput.js']);
-    $libraries['civicrm_contact']['js'][$user_framework_resource_url . 'packages/jquery/plugins/jquery.tokeninput.js'] = [
-      'preprocess' => FALSE,
-      'minified' => FALSE
-    ];
+    if (substr($user_framework_resource_url, 0,-5) == 'core/') {
```
jitendrapurohit: Is there a typo here? Think we need to remove `0,` from the above to check if the last 5 char of the url matches `core/`

    substr($user_framework_resource_url, -5) === 'core/'

PHP 8 seems to have a straightforward fn for the same - https://www.php.net/manual/en/function.str-ends-with.php but seems we need to wait for that to use in code.

Also, should we have an else clause just in case the condition fails?

KarinG: Yes - typo! Fixed that.

Else -> not sure ... if we get there we'll have an unexpected `$user_framework_resource_url`

KarinG:
Actually - that produces:

image

KarinG:
Ok got it! Duh...

KarinG:
I've added a Todo item to remind us to circle back and review this. For now this PR fixes the JS error.

PR #424, opened 2021-01-02
Move includes/utils to services

jitendrapurohit:
Overview
Move utils.inc to services

Before
.inc file

After
- New d8 service added webform_civicrm.utils
- Direct function call replaced to use services.

Comments
ping @KarinG


KarinG:
Re: tests -> we're hitting this -> 

`Drupal\Tests\webform_civicrm\FunctionalJavascript\ActivitySubmissionTest::testSubmitWebform
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "url_generator".`

It's not immediately clear to me what this PR changes in `tests/src/FunctionalJavascript/ActivitySubmissionTest.php` that would cause this error but suspect it's in here. Why add a new Container in (all) tests?

```
    $utils = new Utils();
    $container = new ContainerBuilder();
    \Drupal::setContainer($container);
    $container->set('webform_civicrm.utils', $utils);
```

jitendrapurohit:
Ah yes thanks @KarinG, seems the service container initialization was only required on the utils test and not on the functional-javascript ones. Fixed. All green now :).

tests/src/FunctionalJavascript/ContactSubmissionTest.php
```diff
@@ -3,6 +3,7 @@
 namespace Drupal\Tests\webform_civicrm\FunctionalJavascript;
 
 use Drupal\Core\Url;
+use Drupal\webform_civicrm\Utils;
```
KarinG: Do we need the `use` here? I don't see it in `ActivitySubmissionTest.php` e.g.

KarinG:
Wow big PR! I looked at it noted some inconsistencies in the use of `use` 🙂 👆 If you could please have one more look to see if there can be some more tidying up that would be great.

tests/src/FunctionalJavascript/ContactSubmissionTest.php
jitendrapurohit: Nope, it wasn't required. I've removed it. Thanks for pointing that 👍

KarinG:
Ok merging this! That was a big one 👍

PR #425, opened 2021-01-03
8x port - Fetch domain because APIv3 does not do so reliably

jitendrapurohit:
Overview
Port the domain fix done in https://github.com/colemanw/webform_civicrm/pull/384

Before
Fatal error if membership is enabled on webform.

Failing test - tests/src/FunctionalJavascript/MembershipSubmissionTest.php.

After
No error.


KarinG:
👍 Thank you Jitendra!

PR #428, opened 2021-01-06
Add Dedupe + Contact (Sub) Type Test

KarinG:
Overview
Adding first test for Dedupe Matching Rule testing!

Found and Fixed a bug in `contact_sub_type` Submissions while at it. Subsequent submission of webform adds another `contact_sub_type` of the same:
image

Before
No Dedupe Matching Rule or Contact Sub Type testing.

After
First Dedupe Matching or Contact Sub Type testing.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -736,7 +736,7 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
             unset($exist[$k]);
           }
         }
-        $params[$name] = array_unique(array_merge($params[$name], $exist));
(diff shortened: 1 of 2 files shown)


KarinG:
Ok back to debugging the contact_sub_type issue -> having deja vu:

image

KarinG:
@jitendrapurohit -> I fixed the 'double' `contact_sub_type` -> so that we don't end up with 'student' and 'Student' when merging array_keys:

`$params[$name] = array_unique(array_map("strtolower", array_merge($params[$name], $exist)));`

KarinG:
We are ✅

jitendrapurohit:
Replicated the problem and confirmed the fix 👍

PR #429, opened 2021-01-09
Add Relationship Test.

KarinG:
Overview
Adding a First Test for Relationship and fixed a bug in the process.
This is also the First test that with multiple contacts. 

Before
No relationship test

After
Test! 🎉

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -357,7 +357,13 @@ class Utils implements UtilsInterface {
         $r['type_a'] = strtolower(wf_crm_aval($r, 'contact_type_a'));
         $r['type_b'] = strtolower(wf_crm_aval($r, 'contact_type_b'));
         $r['sub_type_a'] = wf_crm_aval($r, 'contact_sub_type_a');
+        if (!is_null($r['sub_type_a'])) {
+          $r['sub_type_a'] = strtolower($r['sub_type_a']);
+        }
         $r['sub_type_b'] = wf_crm_aval($r, 'contact_sub_type_b');
+        if (!is_null($r['sub_type_b'])) {
+          $r['sub_type_b'] = strtolower($r['sub_type_b']);
+        }
         $types[$r['id']] = $r;
       }
     }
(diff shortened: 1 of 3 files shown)


src/Utils.php
```diff
@@ -384,17 +384,17 @@ function wf_crm_get_contact_relationship_types($type_a, $type_b, $sub_type_a, $s
...
       // Reciprocal form - only show if different from above
       if ($reciprocal
         && ($t['type_a'] == $type_b || !$t['type_a'])
         && ($t['type_b'] == $type_a || !$t['type_b'])
-        && (in_array($t['sub_type_a'], $sub_type_b) || !$t['sub_type_a'])
-        && (in_array($t['sub_type_b'], $sub_type_a) || !$t['sub_type_b'])
+        && (in_array(strtolower($t['sub_type_a']), $sub_type_b) || !strtolower($t['sub_type_a']))
+        && (in_array(strtolower($t['sub_type_b']), $sub_type_a) || !strtolower($t['sub_type_b']))
       ) {
         $ret[$t['id'] . '_b'] = $t['label_b_a'];
```
KarinG: This is fixing a bug very similar to what we noticed in the `ContactDedupeTest` -> originally the Student - Organization relationship did not show up as an option b/c `Student` did not match `student`.

tests/src/FunctionalJavascript/ContactDedupeTest.php
```diff
@@ -3,7 +3,6 @@
 namespace Drupal\Tests\webform_civicrm\FunctionalJavascript;
 
 use Drupal\Core\Url;
-use Drupal\FunctionalJavascriptTests\DrupalSelenium2Driver;
```
KarinG, line 6: Tidy up.

KarinG:
And ✅ 🎉

src/Utils.php
jitendrapurohit: Do we always have `$sub_type_b` and `$sub_type_a` available in lower case? Given that in_array does a case sensitive comparison?

KarinG: A better solution would be to have an in_array non-case sensitive comparison of keys b/c we end up with `sub_type = Student` and `sub_type = student`, etc. We have seen that in a few places now. 

Implications are big: e.g. custom fields on Individual -> Student were not showing up (prior to using strtolower there); in this case you can not create a relationship between Individual -> Student and Organization b/c it does not think there are any relationships for Student (just for student).

I specifically created a sub_type -> Student (upper case) to test this. It's the most common scenario as well.

KarinG: Ok I've resolved this by ensuring we're always comparing contact sub_type in lower case 👆

KarinG: Will await test results before merging!

KarinG: Oh how did that fail? I need to take this back to my local.

KarinG: Ok I found a better place for the `strtolower()` 👍

src/Utils.php
```diff
@@ -357,7 +357,13 @@ function wf_crm_get_relationship_types() {
...
         $r['type_b'] = strtolower(wf_crm_aval($r, 'contact_type_b'));
         $r['sub_type_a'] = wf_crm_aval($r, 'contact_sub_type_a');
+        if (!is_null($r['sub_type_a'])) {
+          $r['sub_type_a'] = strtolower($r['sub_type_a']);
+        }
         $r['sub_type_b'] = wf_crm_aval($r, 'contact_sub_type_b');
+        if (!is_null($r['sub_type_b'])) {
+          $r['sub_type_b'] = strtolower($r['sub_type_b']);
+        }
         $types[$r['id']] = $r;
```
KarinG, line 367: The `!is_null` is important here! If it is `null` it needs to remain `null`.

KarinG:
These tests are passing on my local - fingers 🤞

jitendrapurohit:
great thanks @KarinG 👍

PR #431, opened 2021-01-09
Set title of list of lineitems to 'Payment Information' to use frontend title for contribution pages

mattwire:
Overview
This used to be the contributionpage title. In 8.x we are not using the contribution page and the title has changed to "Payment Information". In CiviCRM "frontend_title" was added to contribution pages and is what is should be shown publicly if any title is shown but we were still showing the "title" field. I think it makes more sense to switch 7.x to use the fixed "Payment Information" string instead of trying to load the contribution page title.

Before
`title` loaded from contribution page but not using public `frontend_title`.

After
`Payment Information` is the title.
image

Technical Details

Comments

### DIFF
--- a/includes/wf_crm_webform_preprocess.inc
+++ b/includes/wf_crm_webform_preprocess.inc
@@ -604,7 +604,7 @@ class wf_crm_webform_preprocess extends wf_crm_webform_base {
     ];
     return theme('table', [
       'sticky' => FALSE,
-      'caption' => $this->contribution_page['title'],
+      'caption' => $this->contribution_page['frontend_title'] ?? $this->contribution_page['title'],
       'header' => [],
       'rows' => $rows,
       'attributes' => ['id' => "wf-crm-billing-items"],


KarinG:
@mattwire - I don't think we should change the way that Contribution Pages work in D7WFC. That's all in D8WFC and it's tested there. It would be a bit scary to port that all to D7WFC and would require communications etc. Even this small change would surprise a lot of D7WFC users.

mattwire:
@KarinG Ok, problem is it's broken since the introduction of `frontend_title` on civicrm contribution pages. I originally thought about just changing the field to use that but then thought it made more sense to simplify and just use a fixed string. What do you think?

KarinG:
Ah understood. Around when did `frontend_title` get introduced?

mattwire:
It was added in 5.20: https://github.com/civicrm/civicrm-core/pull/15418

KarinG:
I've not seen this broken -> this is an ESR (5.27) site:

image

mattwire:
Ok @KarinG Try this?

KarinG:
I'm ok with that 👍

KarinG:
I've asked Pete for input as well.

petednz:
@jitendrapurohit can you pls review or comment

mattwire:
Merging this so we can get into the next webform 7.x release. Most of CiviCRM now prioritises the frontend (public) title field so webform should too

KarinG:
👍 Adding to the list to be reviewed/ported to 8.x

mattwire:
@KarinG I don't think it's required on 8.x - see https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/src/WebformCivicrmPreProcess.php#L646

KarinG:
Ah right - ok crossing it off my list 🙂

PR #433, opened 2021-01-11
D8 :Option for allow to create New Membership record instead of Membership type change

sunilpawar:
Option for allow to create New Membership record instead of Membership type change for existing membership record.

Overview
IF any contact have membership for Org A and now signing up for another Membership Type from Same Org, currently Membership Type get changed. If somebody want to Keep separate record of each membership type, its not feasible as of now.

With this : We would have option in Webform to decide to create new record in such situation.

d8_webform_membership_type_change

Before
Membership type get changed.

After
New Membership Type get created if Option is enabled for specific webform.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -1204,6 +1204,11 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $existing_type = $types[$mem['membership_type_id']];
           if ($type['domain_id'] == $existing_type['domain_id'] && $type['member_of_contact_id'] == $existing_type['member_of_contact_id']) {
             $params['id'] = $mem['id'];
+            // unset if setting is enabled and membership type are different
+            if (!empty($this->settings['skip_membership_type_change']) &&
+              $mem['membership_type_id'] != $params['membership_type_id']) {
+              unset($params['id']);
+            }
             // If we have an exact match, look no further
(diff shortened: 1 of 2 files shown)


KarinG:
Thank you @sunilpawar - we like adding this capability -> but would like to see it on the Membership tab, so that it can be configured per Membership Type similar to how custom fields that can contain multiple rows of data have a Create mode select field -> with options: Create Only and a Create/Edit. 

In terms of Membership -> the new Create Only option would then do what you want; and the Create/Edit option would = the standard CiviCRM behaviour (Create New or Update); 

For example:
image

KarinG:
Hi @sunilpawar - just checking in to see if you are able to make these 👆 edits? 

cc: @jackrabbithanna

sunilpawar:
@KarinG 
i will work suggested changes..

mattwire:
Linking #382 #383 which is doing similar for D7 - @sluc23 can you comment on this PR?

sluc23:
yes, we implemented this feature in our D7 webform_civicrm [fork](https://github.com/ixiam/webform_civicrm).
Cannot comment much about D8, because we are still in D7 version.. I can only say that we added a 3 option selector, instead a boolean setting, as mentioned in #383 :
1. Update existing membership of different type and change type.
2. Update existing membership of same type.
3. Add new membership and don't touch existing.

KarinG:
Closing this for now as outdated. 

@sunilpawar - please re-open when you have time to work on it and we'll be happy to help you see it through.

PR #434, opened 2021-01-13
Update civicrm versions and copy over Transaction API into webform civicrm code base

KarinG:
Overview
Bumping CiviCRM versions for testing -> CiviCRM Transaction API resurfaced -> handled by copying the function into the webform civicrm module itself

Before
Testing CiviCRM 5.32 and 5.33-rc
No deprecation notice for CiviCRM Transact API

After
Testing CiviCRM 5.33 and 5.34-rc
Loud deprecation notice for CiviCRM Transaction API resurfaced b/c of a change in 5.34-rc. 
Solved by copying over the function into webform civicrm module itself. 

Technical Details
Just copied over the relevant function into the webform civicrm code base. 

Comments
Also added a first live payment processor test! 🎉

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,7 @@ jobs:
       # Actually get the latest
       - name: Install iATS Payments CiviCRM Extension
         run: |
-          cd ~drupal/web/sites/default/files/civicrm/ext
+          cd ~/drupal/web/sites/default/files/civicrm/ext
           git clone https://github.com/iATSPayments/com.iatspayments.civicrm.git
       # not sure we need this but could be handy to enable the iATS Payments Extension
       - name: Install CV
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,17 +64,6 @@ jobs:
         run: |
           cd ~/drupal
           COMPOSER_MEMORY_LIMIT=-1 composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages}:${{ matrix.civicrm }} --no-suggest --prefer-dist
-      # curl -L -O https://github.com/iATSPayments/com.iatspayments.civicrm/archive/1.7.3.zip --output 1.7.3.zip
-      # Actually get the latest
-      - name: Install iATS Payments CiviCRM Extension
-        run: |
-          cd ~/drupal/web/sites/default/files/civicrm/ext
-          git clone https://github.com/iATSPayments/com.iatspayments.civicrm.git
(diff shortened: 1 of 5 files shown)


mattwire:
I think you'd be better copying over the "guts" of this function: https://lab.civicrm.org/extensions/contributiontransactlegacy/-/blob/master/api/v3/Contribution/Transact.php - the core one does not interact properly with Civi and if you copy it over the above extension will no longer work because it can't override. As a minimum that means Stripe stops working.

KarinG:
We don't want that! 

a) what specifically would you need copied over from your Extension to ensure Stripe can process transactions with webform civicrm? And how specific are your edits to Stripe?
b) I'm working on getting live payment processor testing as part of our automated testing - once I have that going for iATS Payments would you be able to add one for Stripe?

I'm thinking when we have replaced Transaction API and have confirmation the replacement works in our automated tests for both iATS and Stripe -> then we should merge this.

Changes in 5.34 have made this Transaction API notice louder... so I can no longer test for notices/errors on webform submission without the test balking at the Transaction API one.

jackrabbithanna:
This stripe payment processor sure has caused a lot of problems and necessary core changes, changes to webform civicrm, changes to civicrm entity. And at the end of the day, it still doesn't work worth a darn after all these years. We tried it again last year and ended up abandoning it for something that actually works. Maybe stripe should be sacrificed for the greater good and not the other way around.  I'm sorry this comment doesn't help but pox on stripe

KarinG:
Two ways to enable CiviCRM extensions:

Through the UI
`https://{{ domain }}/civicrm/admin/extensions?action=enable&id=com.iatspayments.civicrm&key=com.iatspayments.civicrm`
and press the `Enable` button

Or via cv (CiviCRM CLI tools)
`cv ext:enable com.iatspayments.civicrm`

KarinG:
Doh - of course that dir does not exist yet until CiviCRM is actually installed! 

Ok trying to download and install an extension via the API - that could/should work!

```
$result = civicrm_api3('Extension', 'download', [
    'key' => "com.iatspayments.civicrm",
]);
```
Yes! Seeing it now!

```
    18 => 
    array (
      'key' => 'com.iatspayments.civicrm',
      'type' => 'module',
      'name' => 'iATS Payments',
      'label' => 'iATS Payments',
      'file' => 'iats',
```

KarinG:
Happy dance 🎉 ✅ 

This is the first live transaction with a real payment processor. I had to dump the `$contribution` array to see the `trxn_id` for myself:

image

KarinG:
Matt says ok to merge - he'll PR Stripe specific bits separately!

PR #435, opened 2021-01-13
Fix Claro theme fieldset conflict.

KarinG:
Overview
Fix Claro theme fieldset conflict: can't view the Existing Contact Element -> Default value settings.  Issue only appears when Claro (new D9 admin theme) is used. All is well with Seven.

Before
image

After 🎉
image


homotechsual:
Test D9 build confirms that this PR seems to fix the issue :-)

PR #439, opened 2021-01-18
When is active is unchecked is_active is set to null string

pradpnayak:
Overview
https://www.drupal.org/project/webform_civicrm/issues/3190897
Before
dynamic custom fields from custom group on webform are not disabled when disabled from custom field edit form.

After
dynamic custom fields from custom group on webform are disabled when disabled from custom field edit form.

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -234,7 +234,7 @@ function webform_civicrm_civicrm_postSave_civicrm_custom_field($dao) {
   if (empty($dao->custom_group_id)) {
     $dao->find(TRUE);
   }
-  if ($dao->is_active) {
+  if ($dao->is_active == 1) {
     $admin_form = \Drupal::service('webform_civicrm.admin_form');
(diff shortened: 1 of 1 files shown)


KarinG:
@pradpnayak - thank you this is an important bug fix. 

Are you able to add a test for this: adding to the CustomFieldSubmission test? Currently that tests for three types of custom fields. Adding a 4th field of type text and setting that to is_active=0 and then testing that it indeed does not show up in the UI should be pretty straightforward. 

PS- if you did this work for one of our original D8WFC sponsors - please do let me know. I'm happy to add the test in that case.

KarinG:
Added 'needs work` label

@pradpnayak - are you able to return to this and add a test? Let me know - otherwise we'll close this for now.

KarinG:
I've asked @jitendrapurohit - to finish this PR.

KarinG:
Merging this as we've added a test 🎉 here -> https://github.com/colemanw/webform_civicrm/pull/593

PR #440, opened 2021-01-18
Fix saving of activity tag when a tag is selected in activity settings but it is not 'user select'

mattwire:
Overview
When you set a tag for an activity but don't add it to the form:
image

Before
Tag is not saved.

After
Tag is saved.

Technical Details
`enabled` is not set if field is not added to the form. But if you have configured a tag it should still be saved.

Comments
This is a bug I found on 7.x. I don't know if it is a problem on 8.x


KarinG:
Thanks @mattwire 👍

I've ported it to D8 and am running the already existing tests now. If this is super important to you `ActivitySubmissionTest` can be expanded to include a test for this. But I think your time is much better spent on a `ContributionStripeTest` 🙂

PR #445, opened 2021-01-26
Fix notice error when total amount is enabled on the webform

jitendrapurohit:
Overview
Fixes notice error on the webform when total amount is enabled.

Before
To replicate

- Create a webform with Contribution section enabled and include total amount. Make sure the total amount is added as a child of contribution page break (this is the default behaviour).

image

- Visit the webform main page. 
- The following notice is shown on the page(or logged in drupal).

>Notice: Undefined index: civicrm_1_contribution_1_contribution_total_amount in Drupal\webform_civicrm\WebformCivicrmPreProcess->displayLineItems() (line 589 of /Applications/MAMP/htdocs/d8/modules/contrib/webform_civicrm/src/WebformCivicrmPreProcess.php)

 
After
Fixed.

### DIFF
--- a/src/WebformCivicrmPreProcess.php
+++ b/src/WebformCivicrmPreProcess.php
@@ -581,7 +581,7 @@ class WebformCivicrmPreProcess extends wf_crm_webform_base implements WebformCiv
     // Support hidden contribution field
     $fid = 'civicrm_1_contribution_1_contribution_total_amount';
     if (!$this->line_items && isset($this->enabled[$fid])) {
-      $elements = $this->node->getElementsDecoded();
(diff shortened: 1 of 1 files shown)


KarinG:
I'm going to add a  `$this->htmlOutput();` to our `ContributionDummyTest.php` b/c to see the notice.

KarinG:
Not seeing it - weird! But I believe you :-) and the fix makes sense.

KarinG:
Should we always use `getElementsDecodedAndFlattened()` -> 

image

In addition to the one this PR replaces -> we also have a few more places where we use `getElementsDecoded()`

PR #447, opened 2021-01-26
[Test only] - Very first test for ExistingContactElement.

KarinG:
Overview
ExistingContactElement still has a few outstanding issues. 

Before
No test 

After
Very first test

Comments
To be expanded as we address some of the outstanding ExistingContactElement issues.


KarinG:
Visual confirmation -> screenshot output -4:
image

PR #452, opened 2021-01-30
Fix static widget for existing contact element

jitendrapurohit:
Overview
Fix Static widget for existing contact element. 

Before
1. Unnecessary horizontal line is displayed for static element.

image

2. Existing contact is not updated if cid is passed in the URL or if contact element is default to the logged in user. To replicate -
- Create a contact in civicrm.
- Use the cid in the URL. This will load the existing value of the contact.
- Update first name or last name to some other value and submit the webform.
- The values are not updated in the existing contact. Instead a NEW contact is created in civicrm.

After
- Deprecated code removed from d7.
- Handling of static widget is added.
- If the display name is enabled in settings -
image

The widget is displayed correctly on the webform without letting the user to update the value.

image

- FuncJs test added to assert both 1 and 2 problems mentioned above.

Comments
@KarinG

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -52,6 +52,7 @@ var wfCivi = (function ($, D, drupalSettings) {
     var cid = $field.val(),
       prep = null,
       hideOrDisable = $field.attr('data-hide-method'),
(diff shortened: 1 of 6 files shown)


KarinG:
Love the clean up! 🎉

PR #454, opened 2021-01-31
Fix select widget for contact element

jitendrapurohit:
Overview
Fix select widget for contact element

Before
Select Element displayed as autocomplete on the webform

image

After
Select Element is rendered for select widget.

image

Added unit test to assert the select element and verifies if the contact options are loaded as per the filter enabled on the element.

Comments
@KarinG :)


KarinG:
Awesome work 👍

PR #448, opened 2021-01-27
To significantly expand CustomFieldSubmission testing

KarinG:
Overview
----------------------------------------
Add test to switch to Static Options

Before
----------------------------------------
Testing with Live Options

Technical Details
----------------------------------------
DRAFT - I can't confirm (yet) that I'm selecting the field...

Goal
----------------------------------------
To significantly expand testing to include the following custom fields: 

- text ✅
- checkboxes (static - with one label altered) 🚒
- date/time ✅
- select (with values OptionA and OptionB - rather than 1 and 2) ✅

image

PR #455, opened 2021-01-31
Fix checkbox errors on webform

jitendrapurohit:
Overview
Fix checkbox errors on the webform

Before
Notice on the webform when checkbox custom fields are added.

image

- Single radio button rendered as a checkbox is missing the label on the form.  To replicate - add a custom field of type "radio" but with only 1 option.

From the screenshot the label of the element is `Agree to terms and conditions` 

image

but is not displayed on the webform

image

After
- Notice fixed.
- Label displayed for single radio option/checkbox.

image

Comments
@KarinG

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1764,13 +1764,13 @@ class AdminForm implements AdminFormInterface {
     if ($button === 'edit-delete' || ($button === 'edit-disable' && $this->settings['nid'])) {
       foreach ($delete_me as $id) {
         $field = $this->webform->getElementDecoded($id);
-        unset($enabled[$field['form_key']]);
+        unset($enabled[$field['#form_key']]);
         ++$deleted;
         if ($button === 'edit-delete') {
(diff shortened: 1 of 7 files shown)


KarinG:
Ok I'm merging this and then will rebase #448

src/AdminForm.php
```diff
@@ -1764,13 +1764,13 @@ public function postProcess() {
...
         ++$deleted;
         if ($button === 'edit-delete') {
           $this->webform->deleteElement($field['#form_key']);
         }
         else {
-          $field['form_key'] = 'disabled' . substr($field['form_key'], 7);
+          $field['#form_key'] = 'disabled' . substr($field['#form_key'], 7);
           $this->webform->setElementProperties($field['#form_key'], $field);
         }
       }
```
KarinG, line 1776: Oh I think this could well fix another Issue reported in the Drupal issue queue.

PR #456, opened 2021-01-31
Problems with multiple Tag checkboxes

ghost:
Overview
Fixes two problems with checkboxes on multiple Tag fields.

Before
1.  Trying to clear a Tag throws a fatal PHP error; trying to create a Tag with tag_id=0.

2.  If a Tag is originally set, and left unchanged on the form, it is nevertheless cleared.

After
Both work correctly

Technical Details
1.  In the routine `saveGroupsAndTags() in WebformCivicrmPostProcess.php`, the $add array is set to include all the checkboxes, with those to be cleared having value 0, which later cause the error.  This patch removes such zero values.

2.  In the routine `getExposedOptions() in WebformCivicrmBase.php` The 'civicrm_options' option fails to process the $exclude array parameter, unlike other cases.  This patch adds similar processing for this option.

Comments
Only tested against Tags, not other uses of checkboxes.

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -530,7 +530,11 @@ abstract class WebformCivicrmBase {
     }
 
     if ($field && $field['#type'] == 'civicrm_options') {
-      return $field['#options'] ?? [];
+	  $exposed = $field['#options'];
+	    foreach ($exclude as $i) {
+		  unset($exposed[$i]);
+	    }	  
+	  return $exposed;
     }
     if ($field && $field['#type'] == 'select') {
       // Fetch static options
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -910,6 +910,11 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         $field_name = 'civicrm_' . $c . '_contact_1_' . $fid;
         if (!empty($contact['other'][1][$type]) || isset($this->enabled[$field_name])) {
           $add = wf_crm_aval($contact, "other:1:$type", array());
+		  foreach ($add as $i => $tagval) {
+			if ($tagval == 0) {
+			  unset($add[$i]);
+			}
+		  }
           $remove = empty($this->existing_contacts[$c]) ? array() : $this->getExposedOptions($field_name, $add);
           $this->addOrRemoveMultivaluedData($field['table'], 'contact', $cid, $add, $remove);
         }


KarinG:
@maynardsmith -> we just merged a PR that also has this important bug fix re: checkboxes -> I suspect it may solve the issue you're describing here. Could you please try? git pull; drush cr;

image

ghost:
No, both the problems fixed in my PR are there in the latest version of the 8.x-5.x branch.  For info I attach the details of the error messages - The first one occurs in the beta4 version, but only appears in the log and isn't fatal.

Error1

Error2

(I installed by downloading the zip and installing manually, and clearing caches, but I checked that it did include the PR you quote.)

KarinG:
Thanks for re-verifying. This is our next priority.

KarinG:
Ok I've started working on this. Thanks for all the details - super helpful 👍 Will report back here today.

KarinG:
1. Reproduced:
image

KarinG:
@maynardsmith - I finished putting together the alternate PR -> #459

I did something very similar to what you did: except when comparing the `$remove` to what it looks like on a D7WFC project - I noticed that's exactly where `$remove` goes astray as well, essentially making the D8WFC `$add` and `$remove` look similar to what they looked like in D7WFC. 

I have intentionally pushed up all my comments/commits here to show you how I go about writing a test and a fix combined. You don't need a local development for testing. The tests themselves are quite readable. Please have a look and don't hesitate to let me know if you have any questions.

I've asked @jitendrapurohit to take a look as well.

PS - since this was originally your PR I'm happy if you would be revising your PR with these edits. But let's wait until @jitendrapurohit has had a chance to glance over it as well.

ghost:
Thank you.  I'm very happy to go with PR#459 since its ready and waiting.

I will look through your comments and tests in more detail, but at first glance while reading a test is not too difficult I would need a lot more background to be able to write one I'm afraid.  For example, I can't see where the '-10-44793467' on the end of the URL you quote comes from.

KarinG:
Re: '-10-44793467' on the end of the URL -> that's the fun part of Functional Javascript testing! 🎉

1. When tests are either failed or completed -> hit Details here (any of them):
image

2. Hit Artifacts -> 
image

3. This downloads a zip with all screenshots taken during the test. For example: look at the the Test that your interested in and step through its screenshots -> for example -> ContributionIatsTest-8 -> shows me that everything looks good after payment:
image

ghost:
Hi, thanks.  I seem to be missing something more fundamental though - how do you run a test?  I thought that URL was it.

KarinG:
As soon as you create a PR -> the tests will automatically start 🎉

KarinG:
Closing this in favour of #459

PR #457, opened 2021-02-01
[Test only] - Expanding CustomField Testing + ToDo: Test Static Options

KarinG:
Overview

- Added: Custom Field of Type Select with Option Values: OptionA and OptionB
- Changed previously added Checkboxes testing to checking just one Checkbox
- Added assertion for previously added Radio -> Checkbox button option

Added ToDo for Testing Static Options. That's currently not working in the Tests - and is likely a Testing issue re: AJAX requests (multiple).


KarinG:
Test fails - need to inspect the HTML output. 

`Behat\Mink\Exception\ElementNotFoundException : Form field with id|name|label|value|placeholder "properties[options][options][civicrm_option_1][label]" not found.`

PR #459, opened 2021-02-01
Fix for adding/removing Tags (checkboxes) - Includes Test

KarinG:
Overview
Drupal issue: https://www.drupal.org/project/webform_civicrm/issues/3194898

Fixing Fatal Error - on Submit Webform:
image

Added
Added GroupsTagsSubmissionTest -> save Tag checkbox, reload post submission (via API as well as on the webform). 

Fixed
Ensure that `saveGroupsAndTags()` only receives the data in the format it is expecting.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -911,17 +911,17 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         if (!empty($contact['other'][1][$type]) || isset($this->enabled[$field_name])) {
           $add = wf_crm_aval($contact, "other:1:$type", array());
           // ToDo - $add in D7WFC only contains the option(s) selected so unset everything else b/c addOrRemoveMultivaluedData is expecting that
-          foreach ($this->getExposedOptions($field_name) as $k => $v) {
-            if ($add[$k] === 0) {
-              unset($add[$k]);
-            }
+          // foreach ($this->getExposedOptions($field_name) as $k => $v) {
+          //  if ($add[$k] === 0) {
+          //    unset($add[$k]);
+          //  }
           }
           $remove = empty($this->existing_contacts[$c]) ? array() : $this->getExposedOptions($field_name, $add);
           // ToDo - $remove should only return the options not selected b/c addOrRemoveMultivaluedData is expecting that
-          foreach ($this->getExposedOptions($field_name) as $k => $v) {
-            if (array_key_exists($k, $add)) {
-              unset($remove[$k]);
-            }
+          // foreach ($this->getExposedOptions($field_name) as $k => $v) {
+          //  if (array_key_exists($k, $add)) {
+          //    unset($remove[$k]);
+          //  }
           }
           $this->addOrRemoveMultivaluedData($field['table'], 'contact', $cid, $add, $remove);
         }
@@ -2405,6 +2405,10 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
             if (in_array($itemName, $val)) {
               $customValue[] = $itemName;
             }
+          }
+
+          if ($table == 'other' && in_array($name, ['group', 'tag'])) {
+            $val = array_filter($val);
           }
 
           // Implode data that will be stored as a string
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -910,7 +910,19 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         $field_name = 'civicrm_' . $c . '_contact_1_' . $fid;
         if (!empty($contact['other'][1][$type]) || isset($this->enabled[$field_name])) {
           $add = wf_crm_aval($contact, "other:1:$type", array());
+          // ToDo - $add in D7WFC only contains the option(s) selected so unset everything else b/c addOrRemoveMultivaluedData is expecting that
+          foreach ($this->getExposedOptions($field_name) as $k => $v) {
+            if ($add[$k] === 0) {
+              unset($add[$k]);
+            }
+          }
           $remove = empty($this->existing_contacts[$c]) ? array() : $this->getExposedOptions($field_name, $add);
+          // ToDo - $remove should only return the options not selected b/c addOrRemoveMultivaluedData is expecting that
+          foreach ($this->getExposedOptions($field_name) as $k => $v) {
+            if (array_key_exists($k, $add)) {
+              unset($remove[$k]);
+            }
+          }
           $this->addOrRemoveMultivaluedData($field['table'], 'contact', $cid, $add, $remove);
         }
       }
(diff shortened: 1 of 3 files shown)


KarinG:
Ok settings form looks good:
image

But it looks like I've got the wrong edit-selector on the Build tab to switch from a List to Checkboxes. Let's get a screenshot!

KarinG:
It does help to get to the Build tab first before clicking Edit to configure the Element! Doh...

KarinG:
So now the Test should Fail on -> 

```
    $this->getSession()->getPage()->pressButton('Submit');
    $this->assertPageNoErrorMessages();

```
Instead we expect the Error notice - and yes -> 
in Drupal_Tests_webform_civicrm_FunctionalJavascript_GroupsTagsSubmissionTest-10-44793467.html we have:

image

KarinG:
Ok so we're correctly failing at: 

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\GroupsTagsSubmissionTest::testSubmitWebform
Behat\Mink\Exception\ResponseTextException: The text "New submission added to CiviCRM Webform Test." was not found anywhere in the text of the current page.
```

KarinG:
Ok pretty big rabbit hole!

This looks like the least invasive glue to shape these arrays into what `private function addOrRemoveMultivaluedData` expects them to look like. 

The description of `addOrRemoveMultivaluedData` says:
   * Handle adding/removing multivalued data for a contact/activity/etc.
   * Currently used only for groups and tags, but written with expansion in mind.

So that's why I decided to glue this for now - I've left a ToDo statement. 
   
This is working now.
image

Let's see if the tests now pass.

KarinG:
Looks like we've got a deja vu -> I remember seeing this when we were fixing checkboxes:

`Notice: Array to string conversion in Drupal\webform\WebformSubmissionStorage->saveData() (line 1343 of /Applications/MAMP/htdocs/d9civicrm.local/web/modules/contrib/webform/src/WebformSubmissionStorage.php) `

KarinG:
I'm commenting out that Notice for now. All values are stored properly in CiviCRM and I can fetch the Volunteer tag using the API. 

Should pass ✅ now.

src/WebformCivicrmPostProcess.php
```diff
@@ -910,7 +910,19 @@ private function saveGroupsAndTags($contact, $cid, $c) {
...
+              unset($add[$k]);
+            }
+          }
           $remove = empty($this->existing_contacts[$c]) ? array() : $this->getExposedOptions($field_name, $add);
+          // ToDo - $remove should only return the options not selected b/c addOrRemoveMultivaluedData is expecting that
+          foreach ($this->getExposedOptions($field_name) as $k => $v) {
+            if (array_key_exists($k, $add)) {
+              unset($remove[$k]);
+            }
+          }
```
jitendrapurohit: seems like this check should be done much before when forming civi data from the webform submission? So that `saveGroupsAndTags()`  already receives the data in the format of what it expects. 

Something like -

```
diff --git a/src/WebformCivicrmBase.php b/src/WebformCivicrmBase.php
index 74b7449..353c978 100644
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -530,7 +530,7 @@ abstract class WebformCivicrmBase {
     }
 
     if ($field && $field['#type'] == 'civicrm_options') {
-      return $field['#options'] ?? [];
+      return $field['#options'] ? array_diff_key($field['#options'], $exclude) : [];
     }
     if ($field && $field['#type'] == 'select') {
       // Fetch static options
diff --git a/src/WebformCivicrmPostProcess.php b/src/WebformCivicrmPostProcess.php
index 404667b..bd3f74a 100644
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2394,6 +2394,9 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
               $customValue[] = $itemName;
             }
           }
+          if ($table == 'other' && in_array($name, ['group', 'tag'])) {
+            $val = array_filter($val);
+          }
 
           // Implode data that will be stored as a string
           if ($table !== 'other' && $name !== 'event_id' && $name !== 'relationship_type_id' && $table !== 'contact' && $dataType != 'ContactReference' && strpos($name, 'tag') !== 0) {
```

Can you pls replace and see if it still passes the test added in this PR?

KarinG: 👍 Thank you for this @jitendrapurohit.  I'll first expand the test to multiple checkboxes/loading existing data etc first b/c that's working now and needs to be locked in before we further rework this.

KarinG:
Thanks for your review @jitendrapurohit - I reworked it and added more testing (to ensure a previously selected Tag loads properly on the webform) and we're ✅ - so merging this!

@maynardsmith - thank you for your work on Tags/getting to the bottom of the issue 👍

ghost:
@KarinG Thank you for carrying it through.

ghost:
I have now loaded the latest 8.x-5.0 branch from Github and unfortunately have found two problems with Tags using multiple checkboxes.  

1.  Trying to clear a Tag throws a fatal PHP error; trying to create a Tag with tag_id=0.
The fix proposed in PR#456, or equivalent, is not included in PR#459.

2.  If a Tag is originally set, and then cleared on the form, it remains set.  (If originally clear it remains cleared.)
As far as I can see this is because the $remove array in saveGroupsAndTags() in WebformCivicrmPostProcess.php is empty, and working back this is because $field['#options'] in getExposedOptions() in WebformCivicrmBase.php is also empty.  I'm not sure what it should be.

This shows first $field and then $remove when clearing one of two checkboxes.

image

KarinG:
I just added your 1. scenario -> https://github.com/colemanw/webform_civicrm/pull/463 
to show that we are able to clear tags now.

**Did you remember to:**
drush cr / clear all caches
create a new webform for testing (many element properties are set when the element is created)

ghost:
Thank you.  I had cleared caches, but failed to realise the importance of rebuilding the webform.  Please disregard my previous comment - these scenarios do work after doing so.

Incidentally, does exporting a webform, deleting it, and reimporting it count as 'rebuilding'?  I am constructing a pair of fairly complex forms, and it would be difficult if I had to rebuild them from scratch by hand if this sort of situation should arise again.

KarinG:
No worries! I appreciate you taking the time to test/report. And we got some extra test lines as a result :-)

What we will do is delete the Element itself. So for example in this case -> go to the CiviCRM Settings tab and then remove set Enable Tags and Groups -> 

toggle to No -> Save Settings.
set back to Yes -> and Save Settings.

That will limit the impact to one element.

Export/Import will not work b/c then the wrong / default settings will carry fwd.

ghost:
One more question, if you don't mind.  When one upgrades to a new version of software, is there any way of knowing which fields may need to be recreated? (I am thinking of formally released versions, not development states taken from Github.)  Or are formal releases different?

ghost:
Apologies...  I screwed up my testing yesterday (again).  After going through your recommended procedure I am still getting fatal errors when clearing a Tag.  Have raised a separate PR at https://github.com/colemanw/webform_civicrm/pull/468 to separate the discussion.

KarinG:
@maynardsmith - no worries - we recently (yesterday) discovered there is a small but very important issue re: webform 8.x-5.x and webform 6.x module re: #form_key 

We are going to tighten up requirements -> webform 6.x - I confirmed with Jake Rockowitz (author of webform module) yesterday that 6.x is the way to go for modules that depend on webform module.

That will help with the - works here - but not there - scenarios that we're running into when fixing these last outstanding bugs

Re: upgrades: once we are no longer in beta - we will be providing upgrade paths (where when you run drush updatedb there will be a webform_civicrm update). I will definitively add detailed release notes when issuing the next beta release. Our plan at this point is to issue one more beta (soon) and then a first full release a few weeks after that.

PR #460, opened 2021-02-05
Debug Static Option.

KarinG:
Overview
Working on the switching Live -> Static Option issue

### DIFF
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -214,22 +214,5 @@ class CivicrmOptions extends WebformElementBase {
     $element = NestedArray::getValue($form, array_slice($radio['#array_parents'], 0, -2));
     return $element;
   }
-
-  /**
-   * {@inheritdoc}
-   */
-  public function hasMultipleWrapper() {
-    return FALSE;
-  }
-
-  /**
-   * {@inheritdoc}
-   */
-  public function hasMultipleValues(array $element) {
-    if (!empty($element['#extra']['multiple']) || (!empty($element['#options']) && count($element['#options']) === 1)) {
-      return TRUE;
-    }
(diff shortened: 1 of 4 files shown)


KarinG:
Not sure why we get this in a test -> but it is working in the UI. 

image

KarinG:
on my local instance [older versions] - I always have a form_key -> this is in the form yml right after hitting save on the CiviCRM Settings screen - so immediately after the Element has been created:

image

src/Plugin/WebformElement/CivicrmOptions.php
```diff
@@ -215,4 +215,21 @@ public static function ajaxCallback(array $form, FormStateInterface $form_state)
...
+   * {@inheritdoc}
+   */
+  public function hasMultipleValues(array $element) {
+    if (!empty($element['#extra']['multiple']) || (!empty($element['#options']) && count($element['#options']) === 1)) {
+      return TRUE;
+    }
+    return FALSE;
+  }
+
 }
```
KarinG: I've lifted this out of this PR - to solve this Notice separately: 

```
  /**
   * {@inheritdoc}
   */
  public function hasMultipleWrapper() {
    return FALSE;
  }

  /**
   * {@inheritdoc}
   */
  public function hasMultipleValues(array $element) {
    if (!empty($element['#extra']['multiple']) || (!empty($element['#options']) && count($element['#options']) === 1)) {
      return TRUE;
    }
    return FALSE;
  }
```
`Array to string conversion in Drupal\webform\WebformSubmissionStorage->saveData() (line 1343 of /Applications/MAMP/htdocs/d9civicrm.local/web/modules/contrib/webform/src/WebformSubmissionStorage.php)`

jitendrapurohit:
closing in favor of https://github.com/colemanw/webform_civicrm/pull/467

PR #462, opened 2021-02-06
Fix the usage of invalid placeholder `!`.

jitendrapurohit:
Overview
Fix invalid placeholder to get rid of notice/warnings.

Before
! is not a valid placeholder in d8 https://api.drupal.org/api/drupal/core!lib!Drupal!Component!Render!FormattableMarkup.php/function/FormattableMarkup%3A%3AplaceholderFormat/8.3.x

image

After
Replaced it with @ or %. No notice displayed on the page.


KarinG:
Nice clean up @jitendrapurohit 🎉

PR #467, opened 2021-02-07
Fix loading of options for civicrm_option plugin.

jitendrapurohit:
Overview
Fix loading of options for civicrm_option plugin.

Before
Options not loaded.

image

After
Fixed -

image

Comments
@KarinG


KarinG:
Before: using D8.9.x + Webform 6.0.1

image

After: 

image

This definitively fixes things - even if we can't lock this into a Test just yet.

PR #468, opened 2021-02-07
Fixes fatal error when clearing a Tag belong to a TagSet using multiple checkboxes.

ghost:
Overview
Webform using a multiple checkbox field for Tags, throws PHP errors when trying to clear a Tag.

Before
Throws the following errors when clearing a tag.

image

image

After
Works correctly.

Technical Details
The function addOrRemoveMultivaluedData() is being called with an entry in $add having a zero value for the tag to be cleared. This is converted three lines later to a zero key and zero value, which goes on to try to create a Tag with zero value for tag_id.

This shows the two states of the $add array:

image

This is an export of the webform used.
[webform.webform.testwf3.yml.zip](https://github.com/colemanw/webform_civicrm/files/5939206/webform.webform.testwf3.yml.zip)

Comments
Previously mentioned in PR#456 and PR#459.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -910,8 +910,9 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         $field_name = 'civicrm_' . $c . '_contact_1_' . $fid;
         if (!empty($contact['other'][1][$type]) || isset($this->enabled[$field_name])) {
           $add = wf_crm_aval($contact, "other:1:$type", array());
-          foreach ($add as $k => $v) {
-            if ($v == 0) {
+          // ToDo - $add should only contains the option(s) selected so unset everything else b/c addOrRemoveMultivaluedData is expecting that and we need this to handle TagSets - this is essentially a fail-safe
+          foreach ($this->getExposedOptions($field_name) as $k => $v) {
+            if ($add[$k] === 0) {
               unset($add[$k]);
             }
           }
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -910,6 +910,11 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         $field_name = 'civicrm_' . $c . '_contact_1_' . $fid;
         if (!empty($contact['other'][1][$type]) || isset($this->enabled[$field_name])) {
           $add = wf_crm_aval($contact, "other:1:$type", array());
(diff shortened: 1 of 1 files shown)


KarinG:
What Drupal and Webform module version are you using?

KarinG:
Comparing yml -> 

image

Your form_key is very odd ->
'#form_key': civicrm_1_contact_1_other_tag**_14**

Mine is -> 
'#form_key': civicrm_1_contact_1_other_tag

Any idea how where/how the **_14** gets appended in your workflow?

PS - > if you hit the Build Tab -> then you can hit Source -> and that gives you a more readable version of the YAML

ghost:
Drupal 8.9.13.  Webform 6.0.1.  Rechecked clearing caches and removing/replacing the Tags field.  Still the same error.

These two Tags are contained in a Tagset, the 14 is the id of the Tagset.  I happened to do it that way because the tags were set up that way for live use.

KarinG:
Ah - it's a TagSet! Ok that's a different bug/issue. What I'll do is add in a fail-safe (very similar to yours). 

Can you please update your Drupal.org issue on Tags:
1. Clarify it's a TagSet
2. show your #form_key -> unpredictable form_keys like -> the _14 are likely what needs to be addressed.

src/WebformCivicrmPostProcess.php
```diff
@@ -910,6 +910,11 @@ private function saveGroupsAndTags($contact, $cid, $c) {
         $field_name = 'civicrm_' . $c . '_contact_1_' . $fid;
         if (!empty($contact['other'][1][$type]) || isset($this->enabled[$field_name])) {
           $add = wf_crm_aval($contact, "other:1:$type", array());
+          foreach ($add as $k => $v) {
+            if ($v == 0) {
+              unset($add[$k]);
+            }
+          }
           $remove = empty($this->existing_contacts[$c]) ? array() : $this->getExposedOptions($field_name, $add);
```
KarinG, line 919: If you can rewrite this to -> (see below) then I can merge this PR because I already tested this:

```
          // ToDo - $add should only contains the option(s) selected so unset everything else b/c addOrRemoveMultivaluedData is expecting that and we need this to handle TagSets - this is essentially a fail-safe
          foreach ($this->getExposedOptions($field_name) as $k => $v) {
            if ($add[$k] === 0) {
              unset($add[$k]);
            }
          }
```

ghost: Code changed as requested, and tested OK at my end.

KarinG:
And ✅ on all other tests.

PR #469, opened 2021-02-07
Lock in versions

KarinG:
Overview
Standardize versions -> requiring webform module ^6.0

Before
Sites could have webform module 8.x-5.x OR ^6.0 installed and though small we noticed some important differences esp. re: #form_key handling. Jake Rockowitz (maintainer of Drupal webform module) recommends we just focus on ^6.0 going forwards.

After
Requiring in the info.yml file:
  - drupal:webform (>= 6.0)

Github Actions specifically pulls in:
          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/webform:6.x-dev@dev' --no-suggest

Checking that
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,14 +66,13 @@ jobs:
           COMPOSER_MEMORY_LIMIT=-1 composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages}:${{ matrix.civicrm }} --no-suggest --prefer-dist
       # For some reason drupal/webform:5.x installs even if it is drupal:^9.0
       - name: Ensure Webform ^6.0
-        if: matrix.drupal == '^9.0'
         run: |
           cd ~/drupal
-          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/webform:^6.0@beta' --no-suggest
+          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/webform:6.x-dev@dev' --no-suggest
       - name: Install webform_civicrm
         run: |
           cd ~/drupal
-          COMPOSER_MEMORY_LIMIT=-1 composer require  drupal/webform_civicrm *@dev
+          COMPOSER_MEMORY_LIMIT=-1 composer require drupal/webform_civicrm *@dev
       - uses: nanasess/setup-chromedriver@master
       - name: Run chromedriver
         run: chromedriver &
--- a/webform_civicrm.info.yml
+++ b/webform_civicrm.info.yml
@@ -1,8 +1,8 @@
-name: 'CiviCRM Webform'
+name: 'Webform CiviCRM'
 type: module
-description: 'CiviCRM and Webform integration'
-core_version_requirement: ^8.7.7 || ^9
+description: 'Webform CiviCRM Integration'
+core_version_requirement: ^8.8 || ^9
 package: 'Custom'
 dependencies:
-  - webform
+  - drupal: webform (>= 6.0)
   - civicrm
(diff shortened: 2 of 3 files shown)


KarinG:
@jitendrapurohit @mglaman - this one is ready for quick eyes 👀 re: syntax

webform_civicrm.info.yml
```diff
@@ -1,8 +1,8 @@
...
+name: 'Webform CiviCRM'
 type: module
-description: 'CiviCRM and Webform integration'
-core_version_requirement: ^8.7.7 || ^9
+description: 'Webform CiviCRM Integration'
+core_version_requirement: ^8.8 || ^9
 package: 'Custom'
 dependencies:
-  - webform
+  - drupal:webform (>= 6.0)
```
mglaman, line 7: Do we want `(>= 6.0)`? Isn't 6.0 only D9 and not D8. Or is it 8.8? We'd also need to update the `composer.json` as well to require ^6.0

I think it's fine just having `drupal:webform` and the composer constraint allowing `^5 || ^6`

.github/workflows/main.yml
```diff
@@ -66,14 +66,13 @@ jobs:
           COMPOSER_MEMORY_LIMIT=-1 composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages}:${{ matrix.civicrm }} --no-suggest --prefer-dist
       # For some reason drupal/webform:5.x installs even if it is drupal:^9.0
       - name: Ensure Webform ^6.0
-        if: matrix.drupal == '^9.0'
         run: |
           cd ~/drupal
-          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/webform:^6.0@beta' --no-suggest
+          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/webform:6.x-dev@dev' --no-suggest
```
mglaman, line 71: Technically we don't need the `@dev` since we're targeting the dev branch

.github/workflows/main.yml
```diff
@@ -66,14 +66,13 @@ jobs:
           COMPOSER_MEMORY_LIMIT=-1 composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages}:${{ matrix.civicrm }} --no-suggest --prefer-dist
       # For some reason drupal/webform:5.x installs even if it is drupal:^9.0
       - name: Ensure Webform ^6.0
-        if: matrix.drupal == '^9.0'
```
mglaman, line 69: So we're dropping testing for Webform 5.x?

mglaman (commented):
Is Webform CiviCRM dropping Webform 5 support? That's how I read this and the test flows. 

It helps if I read:

> Sites could have webform module 8.x-5.x OR ^6.0 installed and though small we noticed some important differences esp. re: #form_key handling. Jake Rockowitz (maintainer of Drupal webform module) recommends we just focus on ^6.0 going forwards.

Then this looks good.

.github/workflows/main.yml
KarinG: As it turns out testing on GitHub Actions was already happening on 6.x.x only (even on the D8.9 -> composer was already pulling in 6.x.x).

webform_civicrm.info.yml
KarinG: Webform module 5.x-8.x and 6.x are both:
`core_version_requirement: ^8.8 || ^9`

.github/workflows/main.yml
KarinG: Ah I just copied this from https://www.drupal.org/project/webform/releases/6.x-dev

`Install with Composer: $ composer require 'drupal/webform:6.x-dev@dev'`

.github/workflows/main.yml
KarinG: And since both are `core_version_requirement: ^8.8 || ^9` - we can drop support for 5.x

webform_civicrm.info.yml
KarinG: I've added a commit to update `composer.json`

PR #471, opened 2021-02-09
d8 port of #180- Use the payment processor payment method instead of the default c…

jitendrapurohit:
…onfigured one

Overview
d8 port for #180  

Before
Default payment instrument is set for the contribution.

After
Payment processor instrument is set for the contirbution done via webforms.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2141,7 +2141,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         'domain_id' => $domain,
       ));
     }
-    if (empty($params['payment_instrument_id'])) {
+    if (empty($params['payment_instrument_id']) && !empty($params['payment_processor_id'])) {
       $params['payment_instrument_id'] = $this->getPaymentInstrument($params['payment_processor_id']);
     }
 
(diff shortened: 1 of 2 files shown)


src/WebformCivicrmPostProcess.php
```diff
@@ -2148,6 +2141,9 @@ private function contributionParams() {
         'domain_id' => $domain,
       ));
     }
+    if (empty($params['payment_instrument_id']) && !empty($params['payment_processor_id'])) {
```
jitendrapurohit, line 2144: included the changes from https://github.com/colemanw/webform_civicrm/pull/289 which was a fix for the regression caused by #180

KarinG: @jitendrapurohit - can you please add a few lines to the `ContributionIatsTest` -> section line 180-186 to check that with this PR the `payment_instrument_id` is indeed no longer `check` (default) but `credit card`?

jitendrapurohit: Done. Thanks for the pointer 👍  @KarinG

KarinG: Awesome - it's just a few lines but nice to secure this 👍

KarinG:
Squashing and merging ✅

PR #474, opened 2021-02-13
Fix Notice.

KarinG:
Overview
Fix Notice!


KarinG:
@jitendrapurohit - for `TagSets` we still need this. I'm just preventing the Notice we get when looking for `===0` in Tags not belonging to `TagSets`.

PR #473, opened 2021-02-12
Relationship type - User Select - fix

KarinG:
Overview
----------------------------------------
- User Select - not function properly. Reported first by @sunilpawar and also by @maynardsmith  -> https://www.drupal.org/project/webform_civicrm/issues/3197125

Before
----------------------------------------
image

After
----------------------------------------
Working 🎉

Technical Details
----------------------------------------
I ported over/rebased some the edits that Sunil PR-ed in re: $data -> in:
https://github.com/colemanw/webform_civicrm/pull/346/files

Comments
----------------------------------------
I'm not sure why this was missed at some point. Or why Custom Field Live/Static Options did not require this. But I've confirmed manually that this works and it does not fail any of our existing tests. 

PS - Still working on locking - User Select - in for relationship_type_id in a test.

PR #475, opened 2021-02-13
Relationship type - User Select - fix 

jitendrapurohit:
Overview
- User Select - not function properly.

Before
image

After
Working 🎉

Comments
Have picked some of the required test commits from https://github.com/colemanw/webform_civicrm/pull/473

### DIFF
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -152,13 +152,17 @@ class CivicrmOptions extends WebformElementBase {
     $as_list = !empty($element['#extra']['aslist']);
     $is_multiple = !empty($element['#extra']['multiple']);
     $use_live_options = !empty($element['#civicrm_live_options']);
+    $data = [];
+    if ($webform_submission) {
+      $data = $webform_submission->getWebform()->getHandler('webform_civicrm')->getConfiguration()['settings']['data'] ?? [];
+    }
 
     if (empty($element['#options'])) {
-      $element['#options'] = $this->getFieldOptions($element);
+      $element['#options'] = $this->getFieldOptions($element, $data);
     }
 
     if ($use_live_options) {
-      $new = $this->getFieldOptions($element);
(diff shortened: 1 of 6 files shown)


KarinG:
Documenting my testing:

1. Create a webform with Relationship Type -> - User Select - -> Issue reproduced
2. Pulled in this PR -> cleared caches -> 

image

🎉

src/Plugin/WebformElement/CivicrmOptions.php
```diff
@@ -152,13 +152,17 @@ public function prepare(array &$element, WebformSubmissionInterface $webform_sub
     $as_list = !empty($element['#extra']['aslist']);
     $is_multiple = !empty($element['#extra']['multiple']);
     $use_live_options = !empty($element['#civicrm_live_options']);
+    $data = [];
+    if ($webform_submission) {
+      $data = $webform_submission->getWebform()->getHandler('webform_civicrm')->getConfiguration()['settings']['data'] ?? [];
+    }
```
KarinG, line 159: Why if `$webform_submission` here?

src/WebformCivicrmPostProcess.php
```diff
@@ -146,6 +146,21 @@ public function validate($form, FormStateInterface $form_state, WebformSubmissio
...
+    $data = $webform_submission->getData();
+    foreach ($data as $field_key => $val) {
+      if (substr($field_key, -20) === 'relationship_type_id' && is_array($val)) {
+        $data[$field_key] = array_values($val);
+      }
+    }
+    $webform_submission->setData($data);
+  }
+
   /**
```
KarinG, line 164: 👍

KarinG:
Awesome  going to merge this before reviewing where we're at with other #form_key and live/static option issues.

src/Plugin/WebformElement/CivicrmOptions.php
jitendrapurohit: From param definition, it seems it can also arrive as NULL, hence avoiding any notice or fatal error in that case.

PR #477, opened 2021-02-14
Further Standardize Test Setup.

KarinG:
Overview
Further Standardize the Test Setup. We are seeing differences between our UI/Manual Testing and Automated Testing -> so review/add Permissions for the Test User + Set themes: Bartik and Seven.

### DIFF
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -51,16 +51,38 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
    */
   protected function setUp() {
     parent::setUp();
+
+    // Make sure we are using distinct default and administrative themes for
+    // the duration of these tests.
+    \Drupal::service('theme_installer')->install(['webform_test_bartik', 'seven']);
+    $this->config('system.theme')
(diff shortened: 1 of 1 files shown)


KarinG:
Perhaps we should take this further and first get edit Elements (Static and Live) and not losing #form_key on save working as uid=1? Feel free to change it to that - if you agree.

KarinG:
PS - I did find a bug in Claro theme on Edit of the Contact Element -> that has been resolved, but that's why (in this PR) I'd like to standardize the themes in the testing to the defaults at the moment: Bartik and Seven.

tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
```diff
@@ -51,16 +51,38 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
...
     parent::setUp();
+
+    // Make sure we are using distinct default and administrative themes for
+    // the duration of these tests.
+    \Drupal::service('theme_installer')->install(['bartik', 'seven']);
+    $this->config('system.theme')
+      ->set('default', 'bartik')
+      ->set('admin', 'seven')
+      ->save();
+
```
KarinG, line 62: I'm going to put this into a separate PR. We have addressed the permission issue already in #478 🎉

PR #480, opened 2021-02-15
[Test only] - Now that we are able to get to Static Options - test changing the Label.

KarinG:
Overview
Now that we are able to get to Static Options - test changing the Label.

After 🎉
image


KarinG:
image

PR #481, opened 2021-02-15
[Test only] - Change to use rootUser because we are making the Tags checkboxes.

KarinG:
Overview
We need to change this test to use rootUser because we're editing the `Tag(s)` Element.
We can no submit a webform and  `$this->assertPageNoErrorMessages();` ✅

Note I did need to rewrite the test a bit so we're no longer checking to see if we can load an existing Tag and remove it. That's more of an Existing Contact Element + Tags test.

### DIFF
--- a/tests/src/FunctionalJavascript/GroupsTagsSubmissionTest.php
+++ b/tests/src/FunctionalJavascript/GroupsTagsSubmissionTest.php
@@ -11,24 +11,9 @@ use Drupal\Core\Url;
  */
 final class GroupsTagsSubmissionTest extends WebformCivicrmTestBase {
 
-  private function addcontactinfo() {
-    // contact_id = 2 -> is the Drupal user
-    $params = [
-      'contact_id' => 2,
-      'first_name' => 'Maarten',
-      'last_name' => 'van der Weijden',
-    ];
-    $utils = \Drupal::service('webform_civicrm.utils');
-    $result = $utils->wf_civicrm_api('Contact', 'create', $params);
-    $this->assertEquals(0, $result['is_error']);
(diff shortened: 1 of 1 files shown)


KarinG:
image

tests/src/FunctionalJavascript/GroupsTagsSubmissionTest.php
```diff
@@ -57,57 +55,28 @@ public function testSubmitWebform() {
...
-    // ToDo: Fix Notice on GitHub tests Notice: Undefined index: #form_key in Drupal\webform_civicrm\Utils->wf_crm_enabled_fields() (line 477 of /home/runner/work/webform_civicrm/webform_civicrm/src/Utils.php).
-    // $this->assertPageNoErrorMessages();
     $this->assertSession()->pageTextContains('New submission added to CiviCRM Webform Test.');
 
     $utils = \Drupal::service('webform_civicrm.utils');
     $api_result = $utils->wf_civicrm_api('Contact', 'get', [
       'sequential' => 1,
       'return' => ["tag"],
-      'contact_id' => 2,
+      'contact_id' => 4,
```
jitendrapurohit, line 75: Can we use a $contactId by retrieving it from api? similar to what is done [here](https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/tests/src/FunctionalJavascript/CustomFieldSubmissionTest.php#L268-L273? ) .

jitendrapurohit: working on the same test file so will merge this and include the above fix in my PR.

PR #484, opened 2021-02-20
Fix live option handling for civicrm_options element

jitendrapurohit:
Overview
Fix dynamic load of option values on the webform.

https://www.drupal.org/project/webform_civicrm/issues/3188442

Before
If a new option is added to the tag, custom field, etc, it requires a rebuild on the setting page to show the updated values on the webform. 

After
Fixed. Newly added values to the custom field options or the tag, etc are loaded straight on the webform. No rebuild is required. 

Technical Details
Moved `civicrm_live_options` out of the `extra` key since it is handled as a main property in the plugins https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/src/Plugin/WebformElement/CivicrmOptions.php#L42.

Comments
@KarinG

### DIFF
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -139,18 +139,33 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
   }
 
   /**
-   * Modify settings so the element displays as a checkbox.
+   * Modify settings so the element displays as a checkbox
+   *
+   * @param string $selector
+   * @param bool $multiple
+   * @param bool $enableStatic
+   *   TRUE if static radio option should be enabled.
    */
-  protected function enableCheckboxOnElement($selector) {
+  protected function editCivicrmOptionElement($selector, $multiple = TRUE, $enableStatic = FALSE) {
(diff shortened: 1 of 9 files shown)


KarinG:
Awesome 🎉 -> totally agree it had to come out of 'extra' 🙂

So this would mean that people will need to remove / re-add some elements to their webforms - but that's ok because we've only tagged beta thus far. So timing is perfect for this kind of element refactoring.

tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
```diff
@@ -139,18 +139,33 @@ protected function assertPageNoErrorMessages() {
...
+   * Modify settings so the element displays as a checkbox
+   *
+   * @param string $selector
+   * @param bool $multiple
+   * @param bool $enableStatic
+   *   TRUE if static radio option should be enabled.
    */
-  protected function enableCheckboxOnElement($selector) {
+  protected function editCivicrmOptionElement($selector, $multiple = TRUE, $enableStatic = FALSE) {
     $checkbox_edit_button = $this->assertSession()->elementExists('css', '[data-drupal-selector="' . $selector . '"] a.webform-ajax-link');
```
KarinG, line 150: Oh nice refactoring of the test 👍

KarinG:
This looks great - I 'm going to give it some manual testing as well

KarinG:
Sorry - I did do some manual testing yesterday but forgot to circle back to document that here! So doing that now and merging it 👍

PR #487, opened 2021-02-22
Do not render civicrm_option element as listbox by default

jitendrapurohit:
Overview
Do not render civicrm_option element as listbox by default
 
Before
Civicrm Option element is displayed as listbox by default. Eg groups, tags, checkbox custom fields, etc.

image

After
Loads correct html by default.

image

Comments
@KarinG


KarinG:
Awesome! Merging. This will make many sitebuilders very happy. It’s a much more sensiible default.

PR #490, opened 2021-02-27
Load default option for civicrm-option element

jitendrapurohit:
Overview
Load default option for civicrm-option element (Tags, country, etc).

Before
Civicrm Option element does not load the default option on the page. Eg Tags, country, etc.

image

After
Fixed.

image

Comments
@KarinG

### DIFF
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -122,8 +122,8 @@ class CivicrmSelectOptions extends FormElement {
           return 1;
         }
 
-        $a_weight = $weight_values[$a];
-        $b_weight = $weight_values[$b];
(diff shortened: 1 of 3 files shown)


KarinG:
Awesome on the fix 🎉

re: the Test edits -> I'd like to keep the tests as readable as we can so that others can help with these in future. So for Tags -> instead of creating new Tags, I'd like us to work with the Tags that CiviCRM core ships with -> 'Volunteer', 'Major Donor', etc. I'm perfectly ok 'hard-coding' these as they will likely stay in CiviCRM forever and everyone knows them. Set one of these as the default and then check that it remained set as the default by going back into Edit Element?

jitendrapurohit:
Have updated the test @KarinG :)

KarinG:
👍 🙂

PR #491, opened 2021-02-27
Show Case Option to file activity On Case

sunilpawar:
Overview
Add activity to Existing Case

Before
Case option was not available..
file_activity_case_before

After
Option to add Activity to case option is available
file_activity_case

Technical Details
If Option Group Label is translated then option group and their option are not rendering.

Comments
https://www.drupal.org/forum/support/post-installation/2016-08-16/group-menus-select-list-with-optgroup-drupal-8
https://www.drupal.org/project/examples/issues/2283665

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -538,8 +538,8 @@ class AdminForm implements AdminFormInterface {
             }
             $wrap['case']["activity_{$n}_settings_case_type_id"]['#options'] = [
               t('- None -'),
-              t('This Webform') => $webform_cases,
-              t('Find by Case Type') => $case_types,
+              'This Webform' => $webform_cases,
+              'Find by Case Type' => $case_types,
             ];
           }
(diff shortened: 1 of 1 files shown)


src/AdminForm.php
```diff
@@ -538,8 +538,8 @@ private function buildActivityTab() {
             }
             $wrap['case']["activity_{$n}_settings_case_type_id"]['#options'] = [
               t('- None -'),
-              t('This Webform') => $webform_cases,
-              t('Find by Case Type') => $case_types,
+              'This Webform' => $webform_cases,
+              'Find by Case Type' => $case_types,
             ];
```
KarinG, line 543: Yeah that would be better! 🙂

KarinG:
Will wait for tests to pass - which they will!
Then we can merge it.

Thank you @sunilpawar 👍

PR #493, opened 2021-02-28
Fix add element error on webform

jitendrapurohit:
Overview
Fix add element error on webform

Before
Add element button sometimes fails with -

>|Drupal\Component\Plugin\Exception\PluginNotFoundException: Plugin ID 'webform_civicrm' was not found. in Drupal\Core\Plugin\DefaultLazyPluginCollection->initializePlugin() (line 79 of /var/www/drupal/web/core/lib/Drupal/Core/Plugin/DefaultLazyPluginCollection.php)

After
Fixed.

Comments
This isn't replicating on d9 so should be something wrong with how handlers were loaded in Drupal 8.x version. Confirming it before retrieving the handler must be able to fix the error.

@KarinG


KarinG:
We've only run into this issue with one project thus far - and confirming that this check successfully works around the problem.

PR #499, opened 2021-03-07
Do not load civicrm elements on the Add element form

jitendrapurohit:
Overview
Do not load civicrm elements on the Add element form

Before
CiviCRM Elements are loaded on Add Element form.

image

After
Not loaded.

Comments
@KarinG


KarinG:
Awesome - this will avoid so much confusion 🎉

PR #504, opened 2021-03-11
Fix submit button placement.

KarinG:
Overview
After Editing the Submit button element on the webform build tab -> I found the Submit button at the top of the Payment page!

Before
image

After
image

Technical Details
Should be a safe fix -> just using moving the Submit button down...

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -134,4 +134,6 @@ cj(function($) {
     $.extend(CRM.payment, payment);
   }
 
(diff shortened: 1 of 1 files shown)


KarinG:
Oh - I think this may be causing a new bug -> on pages where there is no submit button - this is Mac/Firefox (black background at the bottom of the form) -> On Mac/Chrome all is ok.

image

KarinG:
Ok fixed! By being more specific!

jitendrapurohit:
Tested with bartik, seven and claro theme. works fine 👍

PR #506, opened 2021-03-11
Membership select fixes

KarinG:
Before
If Membership is configured as - User Select - -> Membership is not saved (neither in Webform Results nor in CiviCRM).

After
Saved 🎉 

I did a quick check in `Fields.php` to see if there are any other Fields that have an incorrect select type - but nope - it was just this one.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -579,7 +579,7 @@ class Utils implements UtilsInterface {
       $label = Html::escape($label);
     }
     if ($html == 'wrap') {
-      $label = Markup::create('<span class="contact-label number-' . $n . '">' . $label . '</span>');
+      $label = Markup::create($label);
     }
     return $label;
   }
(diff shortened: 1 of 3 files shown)


KarinG:
@jitendrapurohit - this fixes the membership select issue + I finished the test for it. And it's ✅ 🎉

src/Utils.php
```diff
@@ -579,7 +579,7 @@ function wf_crm_contact_label($n, $data = [], $html = 'escape') {
       $label = Html::escape($label);
     }
     if ($html == 'wrap') {
-      $label = Markup::create('<span class="contact-label number-' . $n . '">' . $label . '</span>');
+      $label = Markup::create($label);
```
jitendrapurohit, line 582: Looks like this label should be updated when the label on the Contact tab is changed as per https://github.com/colemanw/webform_civicrm/pull/506/files#diff-29604b8d3d680574d348a450de44692476847958c077b350b4dd553c4864f07dL582

The class was not loading before this PR too so merging this and we can handle this issue in a separate PR.

PR #507, opened 2021-03-13
D8/9 Save multi-valued custom fields through webform

jitendrapurohit:
Overview
Save multi-valued custom fields through webform

Before
Multi-valued fields do not save if no. of exposed fields on the form > 1.

To replicate -

- Create multi-valued custom fields with 2 fields. Eg Month, Hour.
- Create a webform to enter 5 fields for this custom field.
- Submit the form with these values.
- Only first value is saved.

After
All values are saved.

Comments
@KarinG 

Screenshot describing the issue:
image


KarinG:
Best test yet! 🎉 ✅

PR #512, opened 2021-03-18
Add Element CivicrmOptions - spaceholder- b/c Webform seems to be expecting it

KarinG:
in certain cases.

Before
No file named `CivicrmOptions.php in src/Element`

After
File named `CivicrmOptions.php in src/Element`

We need this to avoid occasional errors like:
`Drupal\Component\Plugin\Exception\PluginNotFoundException: The "civicrm_options" plugin does not exist.`


KarinG:
Merging this - Matt Glaman suggested it!

PR #513, opened 2021-03-19
Do not override already existing default_value with default_option.

KarinG:
Overview
Using token module we can set defaults like this:

http://d8civicrm.local/form/paid-membership?membership=3

image

Before
`#default_value': '[current-page:query:membership]'`
had to be added in Source as the value would not save when Editing the Membership - User Select - Element

After
Saving it!

The first commit contains the fix and is composer requiring Token module. I'll be adding a Test.

### DIFF
--- a/src/Plugin/WebformElement/CivicrmSelect.php
+++ b/src/Plugin/WebformElement/CivicrmSelect.php
@@ -77,7 +77,9 @@ class CivicrmSelect extends WebformElementBase {
       }
       $element['#options'] = $new;
     }
-    $element['#default_value'] = $element['#default_option'];
+    if (isset($element['#default_option'])) {
+      $element['#default_value'] = $element['#default_option'];
+    }
   }
 
   protected function getFieldOptions($element) {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,6 +73,9 @@ jobs:
         run: |
           cd ~/drupal
           COMPOSER_MEMORY_LIMIT=-1 composer require drupal/webform_civicrm *@dev
+      - name: Install token
+        run: |
+          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/token:^1.9'
       - uses: nanasess/setup-chromedriver@master
       - name: Run chromedriver
         run: chromedriver &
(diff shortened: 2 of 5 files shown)


KarinG:
Ok calling it a night - I will continue with the actual test tomorrow!

src/Plugin/WebformElement/CivicrmSelect.php
```diff
@@ -77,7 +77,9 @@ public function prepare(array &$element, WebformSubmissionInterface $webform_sub
       }
       $element['#options'] = $new;
     }
-    $element['#default_value'] = $element['#default_option'];
+    if (isset($element['#default_option'])) {
+      $element['#default_value'] = $element['#default_option'];
+    }
   }
```
KarinG, line 83: Ran into this `Notice: Undefined index: #default_option in Drupal\webform_civicrm\Plugin\WebformElement\CivicrmSelect->prepare() (line 80 of /var/www/drupal/web/sites/default/modules/webform_civicrm/src/Plugin/WebformElement/CivicrmSelect.php)`

KarinG:
I'm going to completely refactor this test rather than jamming it in with Free Membership, as really it has little to do with Free Memberships :-) However -> the test must include a CiviCRM Options type of element.

KarinG:
This is from the test -> so far so good:
image

Getting:
`Behat\Mink\Exception\ElementNotFoundException : Element matching css "tabby-toggle_webform-tab--advanced" not found.`

Ah - just click!
`$this->getSession()->getPage()->clickLink('Advanced');`

KarinG:
Ok this is new:
`Drupal\Tests\webform_civicrm\FunctionalJavascript\MembershipSubmissionTest::testSubmitMembershipQueryParams
Exception: element not interactable`

KarinG:
Ah it's in the fieldset -> and it's collapsed:

image

KarinG:
Ok - one more notice -> here on view of the webform:

image

KarinG:
How do we not have a `#form_key`?

Submission of data is fine - I'm going to leave the notices on form view and form submit as a ToDo for now.

KarinG:
I’d like to merge this PR and then follow up with an additional PR that will prevent filling dblog for people using such URL query arguments. Notices are easy to reproduce (see two ToDo statements).

KarinG:
@jitendrapurohit - definitively a 'squash and merge' on this one 🙂

KarinG:
Oh ... deja vu! Must use `rootUser` when going into Edit Element (of type CiviCRM Options) -> else we loose `#form_key` Tests should all pass now: my local is green ✅

jitendrapurohit:
Have tested this and works great 👍. Thanks @KarinG

PR #517, opened 2021-03-21
Fix Notices in Recurring section.

KarinG:
Overview
Refactored transact api is producing a flatter result 🙂

Before
Found some notices in function `testSubmitMembershipAutoRenew()` on webform submission. 

After
No notices! Checked by adding `$this->assertPageNoErrorMessages();` immediately after Submission.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1947,13 +1947,20 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
 
     // If transaction was successful - Update the Recurring Series - currency must be resubmitted or else it will re-default to USD; invoice_id is that of the initiating Contribution
     if (empty($result['error_message'])) {
-      $resultUpdateRecur = $utils->wf_civicrm_api('ContributionRecur', 'create', [
+      $recurParams = [
         'id' => $resultRecur['id'],
         'currency' => $contributionParams['currency'],
         'next_sched_contribution_date' => date("Y-m-d H:i:s", strtotime('+' . $contributionRecurParams['frequency_interval'] . ' ' . $contributionRecurParams['frequency_unit'])),
-        'invoice_id' => $result['invoice_id'],
-        'payment_instrument_id' => $result['payment_instrument_id'],
-      ]);
+       ];
+      if ($APIAction == 'transact') {
+        // Now that we include Transact within webform_civicrm module it comes back flattened:
+        $recurParams['invoice_id'] = $result['invoice_id'];
+        $recurParams['payment_instrument_id'] = $result['payment_instrument_id'];
+      } else {
+        $recurParams['invoice_id'] = $result['values'][$result['id']]['invoice_id'];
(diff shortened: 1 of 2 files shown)


src/WebformCivicrmPostProcess.php
```diff
@@ -1951,8 +1951,8 @@ private function contributionRecur($contributionParams, $paymentType = 'live') {
         'id' => $resultRecur['id'],
         'currency' => $contributionParams['currency'],
         'next_sched_contribution_date' => date("Y-m-d H:i:s", strtotime('+' . $contributionRecurParams['frequency_interval'] . ' ' . $contributionRecurParams['frequency_unit'])),
-        'invoice_id' => $result['values'][$result['id']]['invoice_id'],
-        'payment_instrument_id' => $result['values'][$result['id']]['payment_instrument_id'],
+        'invoice_id' => $result['invoice_id'],
+        'payment_instrument_id' => $result['payment_instrument_id'],
```
jitendrapurohit: think we'll still have the notice if the create api is executed instead of `transact` @KarinG. Eg I see `$APIAction` variable is set to `create`  for condition `if ($paymentType === 'deferred')` so if pay later is used in this test, we'll still get the notice error? probably something like -

```diff
diff --git a/src/WebformCivicrmPostProcess.php b/src/WebformCivicrmPostProcess.php
index f697e1e..0e4cc59 100644
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1947,13 +1947,19 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
 
     // If transaction was successful - Update the Recurring Series - currency must be resubmitted or else it will re-default to USD; invoice_id is that of the initiating Contribution
     if (empty($result['error_message'])) {
-      $resultUpdateRecur = $utils->wf_civicrm_api('ContributionRecur', 'create', [
+      $recurParams = [
         'id' => $resultRecur['id'],
         'currency' => $contributionParams['currency'],
         'next_sched_contribution_date' => date("Y-m-d H:i:s", strtotime('+' . $contributionRecurParams['frequency_interval'] . ' ' . $contributionRecurParams['frequency_unit'])),
-        'invoice_id' => $result['values'][$result['id']]['invoice_id'],
-        'payment_instrument_id' => $result['values'][$result['id']]['payment_instrument_id'],
-      ]);
+      ];
+      if ($APIAction == 'transact') {
+        $recurParams['invoice_id'] = $result['invoice_id'];
+        $recurParams['payment_instrument_id'] = $result['payment_instrument_id'];
+      }
+      else {
+        $recurParams['payment_instrument_id'] = $result['values'][$result['id']]['payment_instrument_id'];
+      }
+      $utils->wf_civicrm_api('ContributionRecur', 'create', $recurParams);
     }
     return $result;
   }

```

KarinG: You’re right. I’d forgotten about ‘deferred’ mode. I’ll make the edit!

KarinG:
@jitendrapurohit - updated to account for the not transact API case!

jitendrapurohit:
Thanks @KarinG. looks great 👍

PR #519, opened 2021-03-24
D8/9 - Fix the submission of multiple contact reference custom fields

jitendrapurohit:
Overview
Fix the submission of activity custom field with type = contact reference

Before
Contact reference custom fields on activity is not saved if no of fields added to the webform > 1.

To replicate -

- Add 2 contact ref custom fields on acivity.
- Create a webform with 3 contacts.
- Enable them on the webform with custom field 1 = Contact 2 and custom field 2 = Contact 3.
- Submit the webform.
- Only value for custom field 1 is saved. 

After
All contact ref custom fields are saved.

Technical Details
This was a problem with the variable usage in the foreach loop. On the first iteration, the tableName is changed leading to a key miss from the second loop execution.

Comments
@KarinG I recall you mentioned having the same problem few days ago? This is also a problem in d7 so would raise a d7 PR too. I'll be adding a test soon.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2338,16 +2338,16 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
           $this->data[$ent][$c][$table][$n][$name] = [];
           foreach ($val as $v) {
             if (is_numeric($v) && !empty($this->ent['contact'][$v]['id'])) {
-              $table = $customOnly ? $ent : $table;
-              $this->data[$ent][$c][$table][$n][$name][] = $this->ent['contact'][$v]['id'];
+              $tableName = $customOnly ? $ent : $table;
+              $this->data[$ent][$c][$tableName][$n][$name][] = $this->ent['contact'][$v]['id'];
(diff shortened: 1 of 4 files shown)


jitendrapurohit:
There were other related issues which is fixed in the latest change.

**Issue 1**: Contact Reference field included in a multi-value custom field is not submitted if it is set to a value form the webform contact. To replicate

- Create a webform with 5 contacts.
- Include a multivalue custom group on the webform with 2 values.
- Set a contact ref custom field as shown below -

image
 
On submitting the webform, only the first value is saved.

**Issue 2**: If a custom contact reference field is set to a contact from the webform, it is not saved if Existing Contact Widget is not selected on the webform.

To replicate: Consider the same example from above. Do not select Existing Contact widget for any of the contacts.

Webform creates new contacts on submission but they are NOT stored in the above custom fields.

Included test to cover both the use-cases.

CDNTaxReceipts:
Awesome work @jitendrapurohit 🎉 ✅ - I'm confirming this is working. And `protected function editContactElement` I'm sure we'll use that in future (you may have to remind me 🙂)

PR #525, opened 2021-04-04
D8/9 Allow type to be updated on civicrm_options element

jitendrapurohit:
Overview
Allow retyping on civicrm_options element

Before
Cannot change `#type` on civicrm_option elements. You need to modify the yaml in order to use the feature like  conditionals, etc.

image

After
`#type` can be changed via UI. This provides additional bits like  a) make conditionals via ui possible + layout options (columns) , etc.

image

The list of elements depends on the config set on the `civicrm_options` form, so if multiple checkbox is enabled for eg Groups, will load only checkboxes. Similarly for fields like Gender, where `multiple` is disabled, will load radios, select element on the form -

image
 

Technical Details
CiviCRM Options element now extends OptionsBase so it belongs to the family of other option elements, eg checkbox, select, radio, tableselect, etc.

Comments
@KarinG

### DIFF
--- a/tests/src/FunctionalJavascript/GroupsTagsSubmissionTest.php
+++ b/tests/src/FunctionalJavascript/GroupsTagsSubmissionTest.php
@@ -11,16 +11,24 @@ use Drupal\Core\Url;
  */
 final class GroupsTagsSubmissionTest extends WebformCivicrmTestBase {
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+    foreach (['GroupA', 'GroupB', 'GroupC'] as $group) {
+      $this->groups[$group] = $this->utils->wf_civicrm_api('Group', 'create', [
+        'title' => $group,
+      ])['id'];
+    }
+  }
+
   public function testSubmitWebform() {
     $this->drupalLogin($this->rootUser);
     $this->drupalGet(Url::fromRoute('entity.webform.civicrm', [
       'webform' => $this->webform->id(),
     ]));
-    // The label has a <div> in it which can cause weird failures here.
-    $this->assertSession()->waitForText('Enable CiviCRM Processing');
-    $this->assertSession()->waitForField('nid');
-    $this->getSession()->getPage()->checkField('nid');
-    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->enableCivicrmOnWebform();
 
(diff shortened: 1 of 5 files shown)


KarinG:
Ok documenting my live testing -> 

1. CiviCRM Custom Field on Activityy -> Select list -> allow Multiple -> when first included on the Webform -> it looks like:
image

2. Build Tab -> Topics -> Edit -> 🎉 Here is the Change button and we can make it into Checkboxes
image

3. That makes available options for layout e.g. 🎉
image

4. And we can now add conditionals via the UI 🎉
image
image

5. Activity + its custom fields are properly saved in CiviCRM 🎉

One small issue -> 
**dblog** -> two notices (when viewing Webform -> Results) -> 

```
Warning: array_keys() expects parameter 1 to be array, bool given in Drupal\webform\Plugin\WebformElementBase->formatHtmlItems() (line 1414 of /var/www/drupal/web/modules/contrib/webform/src/Plugin/WebformElementBase.php)
Warning: Invalid argument supplied for foreach() in Drupal\webform\Plugin\WebformElementBase->formatHtmlItems() (line 1414 of /var/www/drupal/web/modules/contrib/webform/src/Plugin/WebformElementBase.php)
```

Not sure these 👆 notices - are re: re-typing Elements as this live webform has a lot of elements/fields on it. We may need to start adding into tests -> Webform View Results - and `$this->assertPageNoErrorMessages();`

tests/src/FunctionalJavascript/GroupsTagsSubmissionTest.php
```diff
@@ -38,15 +46,18 @@ public function testSubmitWebform() {
...
+    $this->saveCiviCRMSettings();
 
     $this->drupalGet($this->webform->toUrl('edit-form'));
-    $this->assertSession()->waitForField('Checkboxes');
+    $this->assertSession()->waitForField('Group(s)');
     $this->htmlOutput();
+
+    //Change type of group field to checkbox.
+    $this->editCivicrmOptionElement('edit-webform-ui-elements-civicrm-1-contact-1-other-group-operations', FALSE, FALSE, NULL, 'checkboxes');
+
```
KarinG, line 60: Awesome 🎉 this makes it easy to test re-typing of Elements.

KarinG:
Merging this! I expect we may see some possible bugs as a result of this change - but it's a good step forward towards tighter integration with native webform elements (and that has benefits).

PR #532, opened 2021-04-10
D8/9 Fix fatal error on preferred communication element

jitendrapurohit:
Overview
Fix fatal error on preferred communication element

Before
Preferred communication returns a fatal error when submitted without any value. To replicate -

- Enable `Preferred Communication Method(s)` on the webform.
- Disable multiple checkbox for the element.

image

- Submit the webform without selecting any value for the radio

image

After
Fixed.

Technical Details
When webform retrieves the value from civi, Preferred Communication Method(s) is returned as an array if not already set on the contact. This is set as a default value on the element even if the element is of type radio. An empty array isn't a valid value for radio element, hence we should avoid it to set during preprocessing.

Comments
@KarinG :)

### DIFF
--- a/src/WebformCivicrmPreProcess.php
+++ b/src/WebformCivicrmPreProcess.php
@@ -518,7 +518,7 @@ class WebformCivicrmPreProcess extends WebformCivicrmBase implements WebformCivi
             if ($element['#type'] != 'checkboxes' && $element['#type'] != 'date'
               && empty($element['#multiple']) && is_array($val)) {
               // If there's more than one value for a non-multi field, pick the most appropriate
-              if (!empty($element['#options'])) {
+              if (!empty($element['#options']) && !empty(array_filter($val))) {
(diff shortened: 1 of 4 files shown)


src/WebformCivicrmPreProcess.php
```diff
@@ -518,7 +518,7 @@ private function fillForm(&$elements, $submitted = []) {
             if ($element['#type'] != 'checkboxes' && $element['#type'] != 'date'
               && empty($element['#multiple']) && is_array($val)) {
               // If there's more than one value for a non-multi field, pick the most appropriate
-              if (!empty($element['#options'])) {
+              if (!empty($element['#options']) && !empty(array_filter($val))) {
                 foreach ($element['#options'] as $k => $v) {
```
KarinG, line 522: Awesome - makes sense now that checkboxes were working just fine 🙂

PR #535, opened 2021-04-18
D8/9 Fix loading issues on custom field edit/delete in core

jitendrapurohit:
Overview
Fix loading issues on custom field edit/delete in core

Before
Takes longer when a custom field is edited/deleted in core.

After
Fixed.

Technical Details
1. Replaced Webform::loadMultiple wih entityQuery so it only loads the webforms where the custom field is enabled.
2. Return early when the selected custom group is not dynamic.

Comments
@KarinG

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -2340,20 +2340,31 @@ class AdminForm implements AdminFormInterface {
   public static function handleDynamicCustomField($op, $fid, $gid) {
     $utils = \Drupal::service('webform_civicrm.utils');
     $sets = $utils->wf_crm_get_fields('sets');
-    // @todo Start using webform_civicrm_forms to track enabled webforms.
+    $field_name = $element = "cg{$gid}_custom_$fid";
+    if ($op === 'create' || $op === 'enable') {
+      $element = "cg{$gid}_custom_";
+    }
+    // Retrieve all webforms with this custom group.
+    $webforms = \Drupal::entityQuery('webform')
+      ->condition('handlers.webform_civicrm.status', 1)
+      ->condition('elements', $element, 'CONTAINS')
+      ->execute();
     /** @var \Drupal\webform\WebformInterface[] $webforms */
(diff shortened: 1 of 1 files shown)


KarinG:
Testing this now on a site with a couple dozen webforms where adding/deleting CiviCRM fields in the backend had become noticeably slow(er). 

1. Adding a custom field within an existing CG -> really fast - confirmed it's visible in the webform -> settings -> civicrm admin interface
2. Deleting -> super fast as well - massive improvement 🎉 Confirmed the field is correctly removed. 
3. Checked for log/notices -> none ✅

I'm going to review the code - then will add Hans to try this PR.

src/AdminForm.php
```diff
@@ -2340,20 +2340,31 @@ public function setSettings(array $settings) {
...
+    $field_name = $element = "cg{$gid}_custom_$fid";
+    if ($op === 'create' || $op === 'enable') {
+      $element = "cg{$gid}_custom_";
+    }
+    // Retrieve all webforms with this custom group.
+    $webforms = \Drupal::entityQuery('webform')
+      ->condition('handlers.webform_civicrm.status', 1)
+      ->condition('elements', $element, 'CONTAINS')
+      ->execute();
     /** @var \Drupal\webform\WebformInterface[] $webforms */
```
KarinG, line 2352: This really speeds things up 🎉

KarinG:
I've tested this but would like someone else to test this PR as well - preferably Hans who reported it. I'll ping him.

KarinG:
PS - I've also pinged @bmango in Mattermost and here as well 🙂

bmango:
I have tested this patch and it defientely speeds things up significantly when editing/deleting custom fields. Thank you! :-)

KarinG:
Ok merging this! ✅

KarinG:
Hans (on drupal.org) also confirmed this works for their site with 1200 webforms!

PR #539, opened 2021-05-02
D8/9 - Fix pagebreaks on webforms

jitendrapurohit:
Overview
Properly inject page breaks when contribution is enabled.

Before
When the contribution is enabled, the page break is added for payment section, but the contact elements remain as it is. Webform Build page looks like - 

image

No page is available for contact elements, so when webform is viewed, it all renders on a single page -

image

There's an easy workaround to create a page manually for contact element, but it should be automized.

After
Fixed. Root page added and all other elements are nested on save.

image

Webform is loaded as -

image

Comments
Works on webforms created after this PR is added. @KarinG

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1946,6 +1946,9 @@ class AdminForm implements AdminFormInterface {
           }
         }
       }
+      if (isset($enabled['contribution_pagebreak'])) {
+        $this->setParentOnElements($enabled);
+      }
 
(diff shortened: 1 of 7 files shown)


KarinG:
Tested 👍

image

Further on automating things -> I'm thinking we should make Billing Address default to Yes... If Enable Contribution is Yes?

KarinG:
But that 👆 can be a follow-up PR. I'm merging this 🎉

PR #541, opened 2021-05-14
fix multivalue custom fields on contacts other than contact 1

MegaphoneJon:
Overview
Multivalue custom fields don't work on contacts other than contact 1.

Before
Multivalue custom field data is given the wrong index in the contact array, so isn't picked up by the functions that create/update contacts

After
Multivalue custom fields are picked up.

Technical Details
I'm not sure why this line exists, hopefully @jitendrapurohit can respond.  But `$c` is the contact ID, and `$n` is incremented once for each set of this particular custom field group.  But the `$n` is accounted for in `getNameForMultiValueFields()` so I'm not sure what the intent here was.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2375,7 +2375,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
             unset($this->data[$ent][$c][$table][$n][$name]);
             $tableName = substr($name, 0, 6) == 'custom' ? $ent : $table;
             $fld = $customName ?? $name;
-            $this->data[$ent][$c][$tableName][$n][$fld] = $this->ent['contact'][$val]['id'];
+            $this->data[$ent][$c][$tableName][1][$fld] = $this->ent['contact'][$val]['id'];
           }
           elseif (substr($name, 0, 6) == 'custom') {
             $this->data[$ent][$c]['update_contact_ref'][] = $name;
@@ -2511,7 +2511,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
                 $n = $c;
               }
             }
-            $this->data[$ent][$c][$ent][$n][$name] = $customValue ?? $val;
+            $this->data[$ent][$c][$ent][1][$name] = $customValue ?? $val;
           }
         }
       }


KarinG:
Test fail b/c of CiviCRM install. I need to update the test matrix. Will do that and then re-run actions on this PR.

KarinG:
Githhub Actions is failing on - something new:

image

KarinG:
I upped the CiviCRM versions + worked around the temporary `scssphp` issue -> https://github.com/colemanw/webform_civicrm/pull/542 -> and I merged that. Hit re-run actions on your PR should run with the new `main.yml` config file. I'm going to try once more!

KarinG:
But no it runs the old `main.yml` file that this PR was originally submitted with. @MegaphoneJon can you please push something else to this branch? Perhaps you can find a typo to correct in a comment? 🙂

jitendrapurohit:
@MegaphoneJon Thanks for raising. The assignment of `$n` is done to make sure all the multi-value custom fields are included for the insert/update. 

With this removal, only the first custom field value will be saved and all others will be ignored. To replicate -

- Create a multivalue custom field.
- Include > 1 multivalue custom field on the webform 

image

- Submit the webform, it only saves the value for the first field.

---------

Apart from the above, i do agree that the current code is unable to save multi-value field for > 1st contact. I think we should still keep the assignment but change it to `1` instead of `$c`. Does that make sense? If yes, can you pls update and i'll be happy to add a test for this use-case.

Thanks.

MegaphoneJon:
@jitendrapurohit Yes, that makes sense - I actually considered that but wasn't sure I understood the ramifications.  I actually need to test that scenario today so I'll try it out and update this.

MegaphoneJon:
@jitendrapurohit Thank you for looking this over.  I confirmed the bug I introduced, and confirmed that implementing your suggestion fixed it.  I think this is ready to merge.

KarinG:
Ah ok so this is just making it very clear that this works for contact_1 only (much like contributions are added to contact_1)?

KarinG:
If that's the case, then I think the PR just needs a title Edit and then I can merge it.

src/WebformCivicrmPostProcess.php
```diff
@@ -2511,7 +2511,7 @@ private function fillDataFromSubmission() {
                 $n = $c;
               }
             }
-            $this->data[$ent][$c][$ent][$n][$name] = $customValue ?? $val;
+            $this->data[$ent][$c][$ent][1][$name] = $customValue ?? $val;
```
jitendrapurohit, line 2514: ah no - i meant to change the above assignment of 

    $n = $c;

 to 

    $n = 1;

instead. The current change will affect all other entities as well. Sorry if that wasn't clear @MegaphoneJon.

KarinG:
This will be part of - PR https://github.com/colemanw/webform_civicrm/pull/558 - so we can close this.

PR #545, opened 2021-05-16
D8/9 Fix save on setting page if an element is deleted

jitendrapurohit:
Overview
Fixes https://www.drupal.org/project/webform_civicrm/issues/3211923

Before
If an element is deleted, None of the enabled elements are added to the webform.

After
Fixed.

Technical Details
The diff might look large, but it's a simple removal of an else block https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/src/AdminForm.php#L1809

The current code does not rebuild the data if the delete button is clicked. This is clearly an error since we should process this part of the code after the setting page is saved. 

Comments
fyi @petednz @KarinG

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1847,124 +1847,122 @@ class AdminForm implements AdminFormInterface {
     }
 
     // CiviCRM enabled
-    else {
-      $this->rebuildData();
-      if (!$this->settings['toggle_message']) {
-        $this->settings['message'] = '';
+    $this->rebuildData();
+    if (!$this->settings['toggle_message']) {
+      $this->settings['message'] = '';
+    }
+    // Index disabled components
+    $disabled = [];
+    // @todo there is no disabled?
+    /*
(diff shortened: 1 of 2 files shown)


petednz:
I just tested this using this recipe

1.  Add 1 contact and 1 'user select' Activity - adds 4 fields
2. Save
3. Add 2nd contact and change Activity type to a specified option
4. Get warning: These existing fields are no longer needed for CiviCRM processing based on your new form settings. Contact 1 Activity 1: Activity Type
5. Get success msg: Deleted field: Activity Type, Added 2 fields to the form, Saved CiviCRM settings

Seems like a PASS to me

KarinG:
Thank you @petednz for confirming your testing here. 

@jitendrapurohit - the tests failed b/c Github Actions was experiencing an issue this past weekend. I think just need push a commit (a comment edit e.g.) and then it should re-run with the latest Github Action test matrix.

jitendrapurohit:
Rebased. I think it should pass the build now.

KarinG:
It sure does! 🎉 ✅

PR #546, opened 2021-05-17
dedupe with postal address when sharing is enabled

MegaphoneJon:
Overview
The postal address is not passed to the Civi dedupe function if "Use shared address" is enabled, and the shared address is present on the form.

Before
False positives on dedupe (depending on your dedupe rule of course)

After
Address is passed correctly.

Comments
This is a pretty obvious fix if you look at the line directly preceding the one I've changed.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -675,7 +675,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         $m = $contact['address'][1]['master_id'];
         // If master address is exposed to the form, use it
         if (!empty($this->data['contact'][$m]['address'][1])) {
-          $contact['address'][1] = $this->data['contact']['address'][1];
+          $contact['address'][1] = $this->data['contact'][$m]['address'][1];
         }
         // Else look up the master contact's address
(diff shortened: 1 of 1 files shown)


KarinG:
Yes very obvious! Thanks @MegaphoneJon 👍

src/WebformCivicrmPostProcess.php
```diff
@@ -675,7 +675,7 @@ private function findDuplicateContact($contact) {
         $m = $contact['address'][1]['master_id'];
         // If master address is exposed to the form, use it
         if (!empty($this->data['contact'][$m]['address'][1])) {
-          $contact['address'][1] = $this->data['contact']['address'][1];
+          $contact['address'][1] = $this->data['contact'][$m]['address'][1];
         }
         // Else look up the master contact's address
         elseif (!empty($this->existing_contacts[$m])) {
```
KarinG, line 681: @MegaphoneJon - -> looks like we could use this in D7 as well. Do you want to backport it?

https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_webform_postprocess.inc#L590-L596

image

MegaphoneJon:
@KarinG We will backport this some time this month.

PR #548, opened 2021-05-18
require any email for contact 1 for billing, not just email 1

MegaphoneJon:
Overview
Currently, if you have this configuration on contact 1:
Selection_1099

You get this error: `You must enable an email field for :contact in order to process transactions.`  However, the forms works if you have *any* email present for contact 1.

This configuration is necessary because email 1 is always set as primary.  In order to keep an existing record's primary email (assuming a dupe is found), you must uncheck "Email 1" and use "Email 2" instead.

Before
Configuration above works, but throws an error on every save.

After
No error unless ALL email checkboxes are unchecked.

### DIFF
--- a/js/webform_civicrm_admin.js
+++ b/js/webform_civicrm_admin.js
@@ -522,7 +522,7 @@ var wfCiviAdmin = (function ($, D) {
       function billingMessages() {
         var $pageSelect = $('[name=civicrm_1_contribution_1_contribution_enable_contribution]');
         // Warning about contribution page with no email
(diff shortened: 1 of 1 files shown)


KarinG:
I didn't know that this actually works 🙂 Tests are all passing ✅ including a check when one forgets to enable any Email fields: 

```
function testSubmitMembershipAutoRenew() {
    ...
    $this->assertSession()->pageTextContains('You must enable an email field for Contact 1 in order to process transactions.');
    $this->getSession()->getPage()->pressButton('Enable It');
    ...
```

I don't think we have a screenshot for it in the test artifacts download - but those assertions are pretty clear.

KarinG:
See -> https://www.drupal.org/project/webform_civicrm/issues/3214860

PR #550, opened 2021-05-21
support filling textareas from 'Existing Contact'

MegaphoneJon:
Overview
When you have an "Existing Contact" widget that loads CiviCRM data via AJAX, the `fillValues()` function that populates the form ignores textareas.

Before
Textareas are ignored.

After
Textareas are populated like other fields.

Technical Details
There are two components to this:
* Use `data-civicrm-field-key` instead of `name` to identify elements, the `name` has all sorts of quirks, including with textareas.
* That solved the issue for textareas, but not "HTML Textarea" fields, because we need to feed the data into the WYSIWYG.  So I detect if this is a WYSIWYG field and populate that as well.

Comments
I was surprised this worked on D8 - I figured I'd have to patch that then backport - but D8 has a much saner naming convention for elements that avoids this problem.

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -206,7 +206,7 @@ var wfCivi = (function ($, D) {
         return;
       }
       // First try to find a single element - works for textfields and selects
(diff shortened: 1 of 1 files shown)


MegaphoneJon:
This has a "8.x/9.x" tag but is D7-only.

PR #554, opened 2021-05-26
Use text_format type.

puresyntax71:
Overview
It seems to be an artifact of the D7 version. It mentions a module named `webform_html_textarea` which I think isn't needed anymore for data types with `RichTextEditor`. It might've been an autofix of a tool.

Before
Data type of `RichTextEditor` is always a plain textarea despite its data type.

After
Makes data type of `RichTextEditor` use the webform element for `text_format`.

Technical Details

Comments


jitendrapurohit:
Confirmed the changes - Richtext is displayed as textarea by default -

After the patch -

image

Thank You for the patch @puresyntax71

PR #555, opened 2021-05-27
fix checkbox element line item calculation

francescbassas:
Overview
same as [here](https://github.com/colemanw/webform_civicrm/pull/398) but with checkboxes instead of radio elements

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -118,7 +118,7 @@ cj(function($) {
     updateLineItem(lineKey, amount, label);
   }
 
-  $('.civicrm-enabled.contribution-line-item:not(".form-radios")')
+  $('.civicrm-enabled.contribution-line-item:not(".form-radios, .form-checkboxes")')
     .each(calculateLineItemAmount)
     .on('change keyup', calculateLineItemAmount)
     .each(function() {


KarinG:
@francescbassas - thank you for this. We landed with this in D8/D9 -> 

https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/js/webform_civicrm_payment.js#L116-L119

```
$('.civicrm-enabled.contribution-line-item')    
  .each(calculateLineItemAmount)    
  .on('change keyup', calculateLineItemAmount)    
  .each(function() {
```

Can you please verify that this would work in D7? If yes - then let's update to that so that we can keep the codebase as tight/similar as possible.

francescbassas:
@KarinG thanks.

No, D8/D9 solution doesn't work for me when I have a checkbox (radio also) elements. I get this error on the browser console: 

`Uncaught TypeError: fieldKey is undefined` referencing line 546

```
542  function calculateLineItemAmount() {
543    var fieldKey = $(this).data('civicrmFieldKey'),
544      amount = getFieldAmount(fieldKey),
545      label = $(this).closest('div.webform-component').find('label').text() || Drupal.t('Contribution'),
546      lineKey = fieldKey.split('_').slice(0, 4).join('_');
547    updateLineItem(lineKey, amount, label);
548  }
```

Are you sure this not happen on D8/D9 instances? Look at https://github.com/colemanw/webform_civicrm/pull/398#issuecomment-745365922

KarinG:
Ah right... Ok checking -> radio buttons for `civicrm_1_contribution_1_contribution_total_amount`
is working ok in D8/D9->

image

It's in our google doc to review for D8/D9 - I'll add a test. Will merge this after I've done that or else I'll forget again!

KarinG:
#560 is passing ✅ - no port to D8/D9 required.

KarinG:
Thank you @francescbassas 👍

PR #560, opened 2021-06-05
[Test Only] D8/9 - Add test for radio and checkbox amount widget and address chromedriver 91 issues

jitendrapurohit:
Overview
Add test to verify payments with radio and checkbox widget for contribution amount field.

Before
No test.

After
Test added.

Comments
This needed an extra fix in D7 - https://github.com/colemanw/webform_civicrm/pull/398 and https://github.com/colemanw/webform_civicrm/pull/555. But works fine in D8 without any extra handling since the fieldset holds the `civicrm-field-key` by default.

@KarinG

### DIFF
--- a/tests/src/FunctionalJavascript/ContactSubmissionTest.php
+++ b/tests/src/FunctionalJavascript/ContactSubmissionTest.php
@@ -145,24 +145,20 @@ final class ContactSubmissionTest extends WebformCivicrmTestBase {
     $this->drupalGet($this->webform->toUrl('canonical', ['query' => ['cid1' => $contact['id']]]));
     $this->assertPageNoErrorMessages();
 
-    //Check if no autocomplete is present on the page.
+    // Check if no autocomplete is present on the page.
     $this->assertSession()->elementNotExists('css', '.token-input-list');
 
-    //Check if name fields are pre populated with existing values.
+    // Check if name fields are pre populated with existing values.
     $this->assertSession()->fieldValueEquals('First Name', $contact['first_name']);
     $this->assertSession()->fieldValueEquals('Last Name', $contact['last_name']);
 
-    //Update the name to some other value.
-    $params = [
-      'first_name' => 'Frederick' . substr(sha1(rand()), 0, 7),
-      'last_name' => 'Pabst' . substr(sha1(rand()), 0, 7),
-    ];
-    $this->addFieldValue('First Name', $params['first_name']);
-    $this->addFieldValue('Last Name', $params['last_name']);
+    // Update the name to some other value.
+    $this->addFieldValue('First Name', 'Alanis');
+    $this->addFieldValue('Last Name', 'Morissette');
     $this->getSession()->getPage()->pressButton('Submit');
     $this->assertSession()->pageTextContains('New submission added to CiviCRM Webform Test.');
 
-    //Verify if the modified value is updated for the contact.
+    // Verify if the modified value is updated for the contact.
     $contact_result = $this->utils->wf_civicrm_api('contact', 'get', [
       'sequential' => 1,
       'id' => $contact['id'],
@@ -171,12 +167,12 @@ final class ContactSubmissionTest extends WebformCivicrmTestBase {
 
     $this->assertArrayHasKey('count', $contact_result, $result_debug);
     $this->assertEquals(1, $contact_result['count'], $result_debug);
-    $this->assertEquals($params['first_name'], $contact_result['values'][0]['first_name'], $result_debug);
-    $this->assertEquals($params['last_name'], $contact_result['values'][0]['last_name'], $result_debug);
+    $this->assertEquals('Alanis', $contact_result['values'][0]['first_name'], $result_debug);
+    $this->assertEquals('Morissette', $contact_result['values'][0]['last_name'], $result_debug);
 
-    //Enable Autocomplete on the contact Element.
+    // Enable Autocomplete on the contact Element.
     $this->drupalGet($this->webform->toUrl('edit-form'));
-    //Assert civicrm elements are not loaded on Add Element form.
+    // Assert CiviCRM elements are not loaded on Add Element form.
     $this->assertSession()->elementExists('css', '#webform-ui-add-element')->click();
     $this->assertSession()->assertWaitOnAjaxRequest();
     $this->htmlOutput();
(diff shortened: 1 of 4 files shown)


tests/src/FunctionalJavascript/ContactSubmissionTest.php
```diff
@@ -107,8 +107,12 @@ public function testSelectContactElement() {
     }
     $this->getSession()->getPage()->selectFieldOption('Existing Contact', $this->contacts[1]['id']);
     $this->assertSession()->assertWaitOnAjaxRequest();
-    $this->getSession()->getPage()->fillField('First Name', 'Frederick');
-    $this->getSession()->getPage()->fillField('Last Name', 'Pabst');
+    $params = [
+      'first_name' => 'Frederick' . substr(sha1(rand()), 0, 7),
+      'last_name' => 'Pabst' . substr(sha1(rand()), 0, 7),
+    ];
+    $this->getSession()->getPage()->fillField('First Name', $params['first_name']);
```
KarinG: I'd favour that we remove all rand pre and post append from names. I'd rather see descriptive names.

tests/src/FunctionalJavascript/ContactSubmissionTest.php
```diff
@@ -120,8 +124,8 @@ public function testSelectContactElement() {
     $result_debug = var_export($contact_result, TRUE);
 
     $this->assertEquals(1, $contact_result['count'], $result_debug);
-    $this->assertEquals('Frederick', $contact_result['values'][0]['first_name'], $result_debug);
-    $this->assertEquals('Pabst', $contact_result['values'][0]['last_name'], $result_debug);
+    $this->assertEquals($params['first_name'], $contact_result['values'][0]['first_name'], $result_debug);
+    $this->assertEquals($params['last_name'], $contact_result['values'][0]['last_name'], $result_debug);
   }
```
KarinG, line 127: And same here -> I do like actual names in the Assert. I find it makes the test more readable.

tests/src/FunctionalJavascript/ContributionPayLaterTest.php
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @param string $amountType
+   */
+  protected function submitWebform($amountType) {
+    $this->drupalGet($this->webform->toUrl('canonical'));
+    $this->assertPageNoErrorMessages();
+    $this->getSession()->getPage()->fillField('First Name', 'Frederick');
+    $this->getSession()->getPage()->fillField('Last Name', 'Pabst');
+    $this->getSession()->getPage()->fillField('Email', 'fred@example.com');
+
```
KarinG, line 52: For Example -> in this PayLater test we can use:
```
    $this->getSession()->getPage()->fillField('First Name', 'Homer');
    $this->getSession()->getPage()->fillField('Last Name', 'Simpson');
    $this->getSession()->getPage()->fillField('Email', 'HomerSimpson@example.com');
```

KarinG:
This test by itself works just fine!

image

It fails on cross over -> 
```
-'Frederick66dbd49'
+'c6a7175Frederick66dbd49'
```

See comments - I'd favour using descriptive names rather than hashes - then it's easier to track down issues like these.

KarinG:
I've reproduced the test failure on my local -> 

image

KarinG:
So it must be the recent edit to `ContactSubmissionTest` that caused this - can't see what it would be though... debugging!

KarinG:
These are the contacts in the database -> right after `createGroupwithContacts()`

```
contact_id =1 -> 'Default Organization'
contact_id =2 -> email -> begugnpc@example.com
contact_id =3 -> 'bbdf738 0d438f0'
contact_id =4 -> '688b649 8309070'
contact_id =5 -> 'de38360 299af47'
contact_id =6 -> '1fefaf2 99a35dc'
contact_id =7 -> 'aadaeb7 e1c584b'
contact_id =8 -> '845f3f9 421ca34'
```

So that 👆 all looks good.

KarinG:
So contact_id =4 -> becomes something like (actual hash varies) '55cdb53Frederick2bb952d bcdc38fPabst1d233e6' between these two `throw new \Exception` statements:

So what fillField does here -> is appending to the original hash, hash -> so we end up with:
hash . 'Frederick' . substr(sha1(rand()), 0, 7),
hash . 'Pabst' . substr(sha1(rand()), 0, 7),

Why are we using fillField? This is a select contact from a list test - is it not?

```
    // $result_debug = var_export($contact_result, TRUE);
    // throw new \Exception(var_export($contact_result, TRUE));

    // ALL is ok up to this point

    $params = [
      'first_name' => 'Frederick' . substr(sha1(rand()), 0, 7),
      'last_name' => 'Pabst' . substr(sha1(rand()), 0, 7),
    ];
    $this->getSession()->getPage()->fillField('First Name', $params['first_name']);
    $this->getSession()->getPage()->fillField('Last Name', $params['last_name']);
    $this->getSession()->getPage()->pressButton('Submit');

    $contact_result = $this->utils->wf_civicrm_api('contact', 'get', [
      'sequential' => 1,
      // 'id' => $this->contacts[1]['id'],
    ]);
    // $result_debug = var_export($contact_result, TRUE);
    throw new \Exception(var_export($contact_result, TRUE));

    // not ok

**Comment by jitendrapurohit** [comment]
Seems there were 2 problems/regression with the latest chromedriver 91.

1. `fieldValueEquals()` fails to assert the 'refreshed' value on the field. So eg if the DOM is loaded with first name = "John" and with the autocomplete selecion, the first_name is updated to "Fred" - fieldValueEquals('first_name', 'Fred') will fail as it compares it with the original value loaded on the element. Alternate solution is to use the newly added `assertFieldValue()`.

2. As per @KarinG's comment - `fillField()` appends the text to the loaded value on the element. So if first_name is loaded as "John" => `fillField('first_name', 'Fred')` will enter `JohnFred` in the element field. Alternate function to use = `replaceFieldValue()`.

File: tests/src/FunctionalJavascript/WebformCivicrmTestBase.php [hunk 4]
```diff
@@ -143,6 +148,40 @@ protected function assertPageNoErrorMessages() {
...
+      throw new ElementNotFoundException($this->getDriver(), 'form field', 'id|name|label|value|placeholder', $locator);
+    }
+    $field->doubleClick();
+    $field->setValue($value);
+  }
+
+  /**
+   * Assert populated values on the field.
+   * fieldValueEquals() fails for populated values on chromedriver > 91
+   *
```
KarinG, line 175: Just noting that it still works fine for load and check (assert). Things only go sideways when we're testing an update/rewriting a field already containing content.

tests/src/FunctionalJavascript/ContactSubmissionTest.php
KarinG: I've pushed a commit to your fork to introduce you to Jann Arden. More Canadian artists to come! 🙂

PR #561, opened 2021-06-07
D8/9 - Fix issues with 'Existing Contact' element

jitendrapurohit:
Overview
Fix issues with 'Existing Contact' element

Details
### **Problem 1.** 

If `Display Contact Name` is set to Yes for the Existing Contact Element and `Static` Form widget - The label of the field is missing - 

**Before** - 

image

**After** -

image

It works fine for other widgets like Select, autocomplete etc.

### **Problem 2**

For `Form Widget` = Static - If no contact is loaded on the webform, the static widget loads as empty even if it is set to Show in config (`Display Contact Name ` = Yes). It does NOT show the no result text but is just blank (and missing the title) so doesn't look like it's there.

**Before** -

image

**After** -

image

### **Problem 3**

Related Contact is not loaded if main contact is selected via autocomplete (works fine via URL load).

To replicate 

- Set default on contact 3 element to load as the child contact of main contact 1.

image

- Remove default load on primary contact.

Load the webform. Select Primary contact via autocomplete -

image

Related Contact is not filled with ajax response.

**After**

Immediately after selecting the primary contact - the child contact is loaded -

image

### **Problem 4** -

**Before** - No options are loaded to hide/lock fields on contact element.

image

**After -**

image

Comments
Here's a [test form](https://gist.githubusercontent.com/jitendrapurohit/670e6829bc26e07cfd7cff3aee3858c1/raw/1b6104def299a8560e21f36ffcc38f2126f193ea/webform_contact_element.yml) to replicate all the above issues on a single webform.

### DIFF
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -2,8 +2,11 @@
   D.behaviors.webform_civicrm_contact = {
     attach: function (context) {
       $('[data-civicrm-contact]', context).once('webform_civicrm_contact').each(function (i, el) {
-        var toHide = []
-        var field = $(el)
+        var field = $(el);
+        var toHide = [];
+        if (field.data('hide-fields')) {
+          toHide = field.data('hide-fields').split(', ');
+        }
         var autocompleteUrl = D.url('webform-civicrm/js/' + field.data('form-id') + '/' + field.data('civicrm-field-key'));
         var isSelect = field.data('is-select');
         if (!isSelect) {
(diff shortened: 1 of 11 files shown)


KarinG:
I’ll try to get to this before the weekend!

KarinG:
Just did a round of testing and 🎉 

Confirming e.g. -> 
image

Awesome 👍

I'll do a code review tomorrow.

KarinG:
This is great work - I've already pushed it to a live site  🎉 - merging!

PR #564, opened 2021-06-13
D8/9 - Fix missing #options in civicrm_options element

jitendrapurohit:
Overview
Fix missing #options in civicrm_options element

Before
To replicate

- Create a form with a CRM Options element
- Add a Visible conditional to that field
- Remove that conditional
- The edit buttons on the form can then no longer be accessed.

Another way to replicate - 

- Add a new custom field of type select with no options.
- Add this custom field on the webform from the settings page.
- Try to edit any element => Error.

It replicates for any option element. Eg if a user adds an event dropdown but there are no existing events in civicrm, the webform would lead to an error. Similarly for campaign_id dropdown, etc.

After
Fixed.

Technical Details
The change ensures #options are always set on the civicrm_option element.

Comments
@KarinG

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -2223,11 +2223,12 @@ class AdminForm implements AdminFormInterface {
     }
     // Retrieve option list
     if ($field['type'] === 'select') {
+      //To-do: Make this as a default type in src/Fields.php.
+      $field['type'] = 'civicrm_options';
       if ($options = $utils->wf_crm_field_options($field, 'component_insert', $settings['data'])) {
         $field['options'] = $options;
         $field['extra']['items'] = $utils->wf_crm_array2str($options);
         $field['extra']['aslist'] = $field['extra']['aslist'] ?? FALSE;
-        $field['type'] = 'civicrm_options';
       }
     }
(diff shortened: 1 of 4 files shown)


src/AdminForm.php
```diff
@@ -2223,11 +2223,12 @@ public static function insertComponent(&$field, &$enabled, $settings, $create_fi
     }
     // Retrieve option list
     if ($field['type'] === 'select') {
+      //To-do: Make this as a default type in src/Fields.php.
+      $field['type'] = 'civicrm_options';
       if ($options = $utils->wf_crm_field_options($field, 'component_insert', $settings['data'])) {
```
jitendrapurohit, line 2228: @KarinG Before this change, civicrm_option is only set on an element, if there are valid options found in civicrm. For empty options, a normal select widget is created with no option values leading to an error.

I can't think of a field where we would need to add a "select" element instead of our "civicrm_options" field. The latter can be displayed in checkboxes, select, radio etc so I think we should default it in src/Fields.php. Do you agree?

KarinG: Yes agreed 👍

KarinG:
Reviewed and merging  🎉

PR #565, opened 2021-06-14
D8/9 - Move civi tab from inside 'Settings' to the top level

jitendrapurohit:
Overview
Move civi tab from inside 'Settings' to the top level

Before
Civi tab available under settings.

image

After
CiviCRM setting tab is moved to the base route similar to what we had in D7.

image

Comments
@KarinG Seems valid? Not sure why it was sitting under the settings tab. It always required an extra click to arrive at the civi config form.


KarinG:
👍 Sounds good!

PR #566, opened 2021-06-17
Only perform validation on state/province field when it is set

JKingsnorth:
See https://www.drupal.org/project/webform_civicrm/issues/3219316

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -330,17 +330,19 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
           $dt = $element['#civicrm_data_type'];
           // Validate state/prov abbreviation
           if ($dt == 'state_province_abbr') {
-            $ckey = str_replace('state_province', 'country', $key);
-            if (!empty($this->crmValues[$ckey]) && is_numeric($this->crmValues[$ckey])) {
-              $country_id = $this->crmValues[$ckey];
-            }
-            else {
-              $country_id = wf_crm_get_civi_setting('defaultContactCountry', 1228);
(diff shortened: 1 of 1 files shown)


KarinG:
Port of https://github.com/colemanw/webform_civicrm/pull/567

KarinG:
Removed port as this is not a backport of a D8 PR that is merged.

jitendrapurohit:
Hi @JKingsnorth 

Just tried to replicate this on d7 using the steps from https://www.drupal.org/project/webform_civicrm/issues/3219316

Before the PR -

image

After applying the PR, it still returns the same error. As I think the change here does an empty check but on hiding the address fields, the state value still has the victoria selected in my case. 

Am i missing anything? [Here's](https://gist.githubusercontent.com/jitendrapurohit/16833b438029341995393f5b035d4cf1/raw/651bcce354ff64be5c4c3c4f3857cc1b2c7b3804/state_select_error.txt) the exported form if that helps.

jitendrapurohit:
Ah, probably some caching issue. Yes, I can confirm it works fine and submits the webform without any issue 👍

JKingsnorth:
Great, thanks @jitendrapurohit -  do you think this can be merged now @KarinG ?

KarinG:
@JKingsnorth - I'll check in with Jitendra this weekend to finalize this!

jitendrapurohit:
It seems to be doing what was already being done at eg https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/includes/wf_crm_webform_postprocess.inc#L329, i.e, we were already validating the state value when filled. It is just another variable checked in this PR. Looks fine to me. Merging.

PR #567, opened 2021-06-17
D8/9 Only validate state/province if a value for it was submitted

JKingsnorth:
See https://www.drupal.org/project/webform_civicrm/issues/3219316

### DIFF
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -300,7 +300,7 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
     $contactElementEdit = $this->assertSession()->elementExists('css', "[data-drupal-selector=\"{$params['selector']}\"] a.webform-ajax-link");
     $contactElementEdit->click();
     // $this->htmlOutput();
-    $this->assertSession()->assertWaitOnAjaxRequest();
+    // $this->assertSession()->assertWaitOnAjaxRequest();
     $this->assertSession()->elementExists('css', '[data-drupal-selector="edit-form"]')->click();
     if (!empty($params['title'])) {
       $this->getSession()->getPage()->fillField('title', $params['title']);
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -299,7 +299,7 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
 
     $contactElementEdit = $this->assertSession()->elementExists('css', "[data-drupal-selector=\"{$params['selector']}\"] a.webform-ajax-link");
     $contactElementEdit->click();
-    $this->htmlOutput();
+    // $this->htmlOutput();
     $this->assertSession()->assertWaitOnAjaxRequest();
     $this->assertSession()->elementExists('css', '[data-drupal-selector="edit-form"]')->click();
     if (!empty($params['title'])) {
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -299,8 +299,8 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
 
     $contactElementEdit = $this->assertSession()->elementExists('css', "[data-drupal-selector=\"{$params['selector']}\"] a.webform-ajax-link");
     $contactElementEdit->click();
-    // $this->htmlOutput();
-    // $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->htmlOutput();
+    $this->assertSession()->assertWaitOnAjaxRequest($timeout = 20000);
     $this->assertSession()->elementExists('css', '[data-drupal-selector="edit-form"]')->click();
     if (!empty($params['title'])) {
       $this->getSession()->getPage()->fillField('title', $params['title']);
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -300,7 +300,7 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
     $contactElementEdit = $this->assertSession()->elementExists('css', "[data-drupal-selector=\"{$params['selector']}\"] a.webform-ajax-link");
     $contactElementEdit->click();
     $this->htmlOutput();
-    $this->assertSession()->assertWaitOnAjaxRequest($timeout = 20000);
+    $this->assertSession()->assertWaitOnAjaxRequest();
     $this->assertSession()->elementExists('css', '[data-drupal-selector="edit-form"]')->click();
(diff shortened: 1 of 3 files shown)


JKingsnorth:
Re-test this please

KarinG:
@JKingsnorth - tests are running now on 4 combinations of Drupal and CiviCRM. Tests will kick off automatically when an approved user pushes any commit. So from hereon forwards they will just run.

KarinG:
Passes on D8 (5.33 and 5.35) but fails on 9.1 / CiviCRM RC testing -> I'll have to dive in and see why that is. I suspect it's more likely the RC then your PR - but that's just as important to figure out. 

Drupal\Tests\webform_civicrm\FunctionalJavascript\ExistingContactElementTest::testRenderingOfExistingContactElement

JKingsnorth:
Thanks for taking a look @KarinG . The pr just adds an extra is_empty, so I'd be surprised too if it broke tests.

KarinG:
@JKingsnorth -> upgrading to D9 is not as bad as you may think (saw your Mattermost inquiry)! And rebuilding webform and views is actually what you would in order to get at all the new functionality.

KarinG:
Ah - and with D9.2 released -> we're automatically testing against that:

image

KarinG:
Narrowed it down to -> 
`$this->assertSession()->assertWaitOnAjaxRequest();`

We use that a lot to wait for a page refresh after in many places - so this tells us we really have an issue with existing contact element on the combination of D9.2 and CiviCRM rc.

KarinG:
I've added 9.1.10 specifically to the test matrix and pushed that to your patch-1 branch - narrowing this down.

KarinG:
Ok so it looks like it is the RC that's the issue. Pulling this to my local to dig in today.

KarinG:
We absolutely need 
`$this->assertSession()->assertWaitOnAjaxRequest();`

Or subsequent elements don't have time to become visible.

KarinG:
Ok - running this on my local -> w/ 5.36 I think? Definitively not 5.39-dev.
image

Now upgrading -> 
`karins-MBP:d9civicrm.local sysadmin$ composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages}:5.39.x-dev --no-suggest --prefer-dist --with-dependencies`

db updates -> 
`karins-MBP:d9civicrm.local sysadmin$ cv upgrade:db
`
Yep it was 5.36:

```
karins-MBP:d9civicrm.local sysadmin$ cv upgrade:db
Found CiviCRM database version 5.36.0.
Found CiviCRM code version 5.39.beta1.
Checking pre-upgrade messages...
```

Ok let's run that same test again!

KarinG:
And it passes ... 
image

Ah one more thing to upgrade:
"civicrm/civicrm-drupal-8": "5.36.0"

`  - Upgrading civicrm/civicrm-drupal-8 (5.36.0 => 5.39.x-dev b2ba66b)`

Running test once more... and still passes ✅ on my local 🙃 🤔

KarinG:
Ok so the issue is definitely 9.2.x and not the CiviCRM RC

tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
```diff
@@ -300,7 +300,7 @@ protected function editContactElement($params) {
     $contactElementEdit = $this->assertSession()->elementExists('css', "[data-drupal-selector=\"{$params['selector']}\"] a.webform-ajax-link");
     $contactElementEdit->click();
     $this->htmlOutput();
-    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->assertSession()->assertWaitOnAjaxRequest($timeout = 20000);
     $this->assertSession()->elementExists('css', '[data-drupal-selector="edit-form"]')->click();
```
KarinG: We can remove this - I was just trying to see if we needed to increase the timeout on this `assertWaitOnAjaxRequest` in 9.2.0

KarinG: Ok removing this now

KarinG:
Hi @JKingsnorth - ok we have resolved the issue with 9.2.0 - onwards we go to the actual substance of your PR. This is re: Country (not County right)?

KarinG:
Other than that this makes sense 👍

KarinG:
Trying to follow your use case / reproducing the issue -> 

If I add conditional logic to the webform to hide the address fields if 'Show address' is set to 'No'
-> how then I can I use the webform to update the address?

JKingsnorth:
Hi @KarinG , thanks again for your work on this.

-> how then I can I use the webform to update the address?

You can't! That's deliberate. The problem is that when the address fields in the webform are 'hidden', CiviCRM still tries to validate the state/province field. But because there is no 'Country' field value, it defaults to the 'default' country set for CiviCRM.

So in our case, the CiviCRM default country is 'United Kingdom'.

If I set up the webform as described in the steps to recreate, and first give myself an address in Australia, with the state of Victoria.

Then I go back to the webform, and 'hide' the address fields, and save the form.

Then I receive an error that 'Victoria' is not part of 'United Kingdom'.

The fix here is that if the processing doesn't find a value passed in for state/province (either because the form elements were hidden, or because one was not set) - don't try to validate it against the country field.

Does that make more sense?

KarinG:
```
- Add another webform component, such as a boolean 'Show address? : Yes/No'
- Add conditional logic to the webform to hide the address fields if 'Show address' is set to 'No'
- Use the webform to update your address to one NOT in your default country, and assign a province (eg: Australia > Victoria)
```

So these are two form submissions? First: Show address yes -> submit form
Then go back to form -> Show address no -> and then observe the validation error?

Can you please update all County references to Country (PR and drupal.org ticket)?

JKingsnorth:
Yes, you're right in the steps to recreate there.

But I think we don't need to change references to 'country' because it's not the 'country' we're validating - it's the state/province field. If it helps I can change 'county' to 'state/province' (but I'm from the UK and they are synonymous, sorry if that caused confusion.).

We only want to perform validation on the state/province field if there is a value for it.

KarinG:
Ah - ok -> yes please make it state/province instead of county. County to me reminds me of this AGH extension -> https://civicrm.org/extensions/us-county-loader -> which loads US Counties into the mix. 

I'm jamming on a client deadline but hope to circle back tomorrow or Saturday to reproduce/merge this!

KarinG:
Looks like we have some bigger issues with state/province validation -> this is me -> and that's where I grew up:

image

and:
image

Need to debug this. Looks like we can select provinces that belong to the DefaultCountry

JKingsnorth:
@KarinG - I've renamed the issues.

Interesting, I tried that on my Drupal 7 instance just now and it seems to be working OK. So it must be specific to D8/9 or configuration?

image

---

Any more blockers to getting these merged in?

KarinG:
Thanks for confirming that. We just need to work through some more bits as in D8/9 the validation fails because state_province_id becomes unset (when a conditional hides it) - possibly Th/Fr this week!

KarinG:
I looked at this with @jitendrapurohit today and with PR https://github.com/colemanw/webform_civicrm/pull/574 -> there is no more issue re: Country and State/Province field - and conditionals are working in D8/D9.

So this issue can be closed. For a D7 solution -> look at backporting #574

PR #569, opened 2021-06-20
D8/9 - Fix form defaults on multi-step forms

jitendrapurohit:
Overview
Extend https://github.com/colemanw/webform_civicrm/pull/562 to return early when `Previous` button is clicked + Add test.

Before
`alterForm()` code is executed completely when previous button is clicked

After
Returned early as it was done in d7 + test.

Comments
fyi @KarinG @maynardsmith

### DIFF
--- a/src/WebformCivicrmPreProcess.php
+++ b/src/WebformCivicrmPreProcess.php
@@ -85,14 +85,13 @@ class WebformCivicrmPreProcess extends WebformCivicrmBase implements WebformCivi
       }
     }
     // Early return if the form (or page) was already submitted
-    /*
     $triggering_element = $this->form_state->getTriggeringElement();
-    if ($triggering_element && $triggering_element['#id'] == 'edit-previous'
-      || (empty($this->form_state->isRebuilding()) && empty($this->form['#submission']->is_draft))
+    if ($triggering_element && $triggering_element['#id'] == 'edit-wizard-prev'
+      || (empty($this->form_state->isRebuilding()) && !empty($this->form_state->getValues()) && empty($this->form['#submission']->is_draft))
       // When resuming from a draft
       || (!empty($this->form['#submission']->is_draft) && empty($this->form_state->getUserInput()))
     ) {
-      $this->fillForm($this->form['submitted']);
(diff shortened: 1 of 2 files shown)


KarinG:
@maynardsmith - can you please read through the test in this PR to ensure it matches (functionally) the scenario that you ran into w/ next and back buttons?

lalgwebdev:
@KarinG  Yes, it addresses the basic scenario of losing updates to fields which were defaulted from a 'cid=xxx' in the URL.  It wouldn't find the error in my first version of the PR which failed (it still restored the default after Next/Previous) when a field was deleted rather than being updated with a different value.  

It also doesn't check the other scenario of losing information between the page it is entered on, and the Payment page, if there is another page in between.  This however is a direct consequence of the first simple scenario and may not be worth testing for separately.

ghost:
Oops - that was me with a different hat on.

ghost:
@KarinG  PS.  If it's not a protected field, changing `'CooperUpdated'` on lines 109 and 116 to `''` (or equivalent syntax) would address my first caveat about testing for deleting a field.

Also to clarify, my comment about losing Payment information applies in our situation to Membership renewals.  I have not looked at other payment scenarios, but would not expect a problem unless there is a similar situation where the initial value is defaulted from the prior state of the Contact, derived from the ?cid= in the URL, or the current user.

KarinG:
Hi @jitendrapurohit - when you have a chance can you please rebase this PR?

KarinG:
> @KarinG PS. If it's not a protected field, changing `'CooperUpdated'` on lines 109 and 116 to `''` (or equivalent syntax) would address my first caveat about testing for deleting a field.
> 
> Also to clarify, my comment about losing Payment information applies in our situation to Membership renewals. I have not looked at other payment scenarios, but would not expect a problem unless there is a similar situation where the initial value is defaulted from the prior state of the Contact, derived from the ?cid= in the URL, or the current user.

Hi @maynardsmith - when this is merged - feel free to submit a PR that clears out the last_name field (rather than _updated) - to run/test your scenario. No need to run these tests on your local first. Just make the edit and PR it and let Github Actions do the work for you 🙂

PR #574, opened 2021-06-26
D8/9 - Fix incorrect validation error when non US country is selected

jitendrapurohit:
Overview
Fix validation error on submitting the webform with Country != US.

Before
Validation error when country = UK or any value except the default US -

image

After
Fixed.

Comments
@KarinG


KarinG:
Tested/Confirmed 🎉

PR #575, opened 2021-06-27
D8/9 - Allow existing relationships to expire on webform submission.

jitendrapurohit:
Overview
Allow existing relationships to expire on webform submission.

Before
In a webform where Contact 2 is e.g. Emergency Contact -> everything works great -> except for the part that if I want to remove a relationship (User Select on the webform) -> e.g. Melanie is no longer my Partner -> if I uncheck the box Partner off -> that Relationship continues to exist and will continue to prefill when the form is loaded -> so I can add checkboxes but can’t remove any. 

After
On submit, unselected relationship types are expired.

Comments
@KarinG

### DIFF
--- a/src/AdminHelp.php
+++ b/src/AdminHelp.php
@@ -312,6 +312,16 @@ class AdminHelp implements AdminHelpInterface {
     $this->fee();
   }
 
+  protected function relationship_relationship_type_id() {
+    return '<p>' .
+      t('Click the + button to select more than one option.') .
+      '</p><p>' .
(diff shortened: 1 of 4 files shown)


KarinG:
Ok testing this now!

KarinG:
Before:
image

After 👍 🎉
image

In CiviCRM -> 
image

oh and test is ✅ 

We may need to think about about the Additional Options -> it's becoming a bit of a mix of things. That's not holding up this PR - just flagging it as a place that I typically forget to look in :-)

Checking dblog -> no notices. Looking at code now.

PR #579, opened 2021-07-02
3221896: Don't enforce locked fields when contact is anonymous

JKingsnorth:
See https://www.drupal.org/project/webform_civicrm/issues/3221896

Will require QA and port for D8/9.


KarinG:
@JKingsnorth we are D8/D9 first -> so I'll need this reproduced on D8 first and then we'll backport to D7 if applicable.

PR #587, opened 2021-07-21
Added SelectAll SelectNone feature in CiviCRM static options element

aaayyuusshh:
Drupal Issue URL: https://www.drupal.org/project/webform_civicrm/issues/3224544

Work in progress: Checking off this checkbox will not enable all countries(check off checkboxes below it) like it is supposed to yet.

Overview
Added a checkbox beside "ENABLED"  in the static options for the country field. However, the checkbox is not responsive, i.e: clicking it will check itself but it doesn't do anything to the checkboxes below it. 
Expected: Checking off the checkbox should enable all countries (i.e: checkoff all checkboxes associated with countries)

### DIFF
--- a/js/webform_civicrm_options.js
+++ b/js/webform_civicrm_options.js
@@ -56,9 +56,8 @@ if (typeof jQuery.fn.prop !== 'function') {
         }
       }).change();
 
-     //Currently we are looking at using #tableSelect thus, we are commenting out 'select-all-civi-options' code for now
-     //until we are certain that we can remove it.
-      
+      // Currently we are looking at using #tableSelect thus, we are commenting out 'select-all-civi-options' code for now
+      // until we are certain that we can remove it.
       /**
       $('input.select-all-civi-options').once('wf-civi').change(function() {
         if ($(this).is(':checked') ) {
@@ -68,7 +67,7 @@ if (typeof jQuery.fn.prop !== 'function') {
           $('input.civicrm-enabled, input.select-all-civi-options, input.select-all-civi-defaults').prop('checked', false);
         }
         $('input.civicrm-enabled').change();
-      }); 
+      });
       */
 
       var defaultName = 'civicrm_options_fieldset[civicrm_defaults]';
--- a/js/webform_civicrm_options.js
+++ b/js/webform_civicrm_options.js
@@ -56,6 +56,10 @@ if (typeof jQuery.fn.prop !== 'function') {
         }
       }).change();
 
+     //Currently we are looking at using #tableSelect thus, we are commenting out 'select-all-civi-options' code for now
+     //until we are certain that we can remove it.
+      
+      /**
       $('input.select-all-civi-options').once('wf-civi').change(function() {
         if ($(this).is(':checked') ) {
           $('input.civicrm-enabled, input.select-all-civi-options').prop('checked', true);
@@ -64,7 +68,8 @@ if (typeof jQuery.fn.prop !== 'function') {
           $('input.civicrm-enabled, input.select-all-civi-options, input.select-all-civi-defaults').prop('checked', false);
         }
         $('input.civicrm-enabled').change();
-      });
+      }); 
+      */
 
       var defaultName = 'civicrm_options_fieldset[civicrm_defaults]';
 
--- a/css/webform_civicrm_options.css
+++ b/css/webform_civicrm_options.css
@@ -1,0 +1,4 @@
+/* Edit CiviCRM Options Element */
+#webform-civicrm-options-wrapper .table-select .form-item {
+  display:none;
+}
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -97,6 +97,7 @@ class CivicrmSelectOptions extends FormElement {
           'group' => 'weight',
         ],
       ],
+      '#value_callback' => [get_called_class(), 'valueCallback'],
     ];
 
     if ($element['#civicrm_live_options']) {
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -101,6 +101,7 @@ class CivicrmSelectOptions extends FormElement {
 
     if ($element['#civicrm_live_options']) {
       $element['options']['#tabledrag'] = [];
+      $element['options']['#tableselect'] = FALSE;
     }
 
     $current_options = $element['#default_value'];
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -39,7 +39,7 @@ class CivicrmSelectOptions extends FormElement {
       // return static::convertOptionsToValues($options, $element['#options_description']);
     }
     elseif (is_array($input) && isset($input['options'])) {
-       return $input['options'];
+      return $input['options'];
     }
     else {
       return $element['#default_value'];
@@ -60,16 +60,14 @@ class CivicrmSelectOptions extends FormElement {
 
     $element['options'] = [
       '#type' => 'table',
-      '#tableselect' => FALSE,
+      '#tableselect' => TRUE,
       '#header' => [
         'item' => [
           'data' => ['#markup' => 'Item',]
         ],
         'enabled' => [
           'data' => [
-              '#type' => 'checkbox',
-              '#checked' => 'checked',
-            '#title' => "<b> ENABLED </b>",
+            '#markup' => 'Enabled',
             '#access' => !$element['#civicrm_live_options'],
           ]
         ],
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -67,7 +67,9 @@ class CivicrmSelectOptions extends FormElement {
         ],
         'enabled' => [
           'data' => [
-            '#markup' => 'Enabled',
+              '#type' => 'checkbox',
+              '#checked' => 'checked',
+            '#title' => "<b> ENABLED </b>",
             '#access' => !$element['#civicrm_live_options'],
           ]
         ],
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -174,7 +174,7 @@ class CivicrmSelectOptions extends FormElement {
       ];
       $weight++;
     }
-
+    $element['#attached']['library'][] = 'webform_civicrm/civicrmoptions';
     return $element;
   }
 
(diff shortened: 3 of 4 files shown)


KarinG:
👍 It’s a checkbox! Great start.

KarinG:
This is a Drupal 7 screenshot -> 
image

Note the class `select-all-civi-options`

All JS from D7 including the bit that pertains to the select-all functionality -> 
https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/js/webform_civicrm_options.js#L59-L67

was moved into the D8 (which is also the D9) branch -> it lives here -> https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/js/webform_civicrm_options.js#L59-L67 

So the puzzle now is why is it not working?

KarinG:
PS you filed an issue on drupal.org for this. It's good habit to put the link to the drupal issue in the very top of this PR -> hit edit on your description at the top here and add the URL.

aaayyuusshh:
Hey Dr.Karin, I have set up php storm, it is ready to  go on my end. I will
edit my issue on drupal ASAP when I get to work and I will also continue
trying to fix the issue. Thank you!

Best regards,
Aayush

aaayyuusshh:
@KarinG @jitendrapurohit 
Upon adding '#attributes' => ['class' => ['select-all-civi-options']] into 'enabled', the checkbox is still not reactive:
Screen Shot 2021-08-10 at 6 42 02 PM

When I inspect the checkbox, the class is correctly being updated to select-all-civi-options.
This is the inspect tab:
Screen Shot 2021-08-10 at 6 41 21 PM

However, unlike the 7th version, it says "select-all-civi-options form-checkbox" rather than "select-all-civi-options wf-civi-processed". I'm wondering what the problem is?
Thank you.

KarinG:
@aaayyuusshh @Aashirbadd and I discussed this this morning -> ToDo:

**1. update the PR to include:** ->
```
diff --git a/src/Element/CivicrmSelectOptions.php b/src/Element/CivicrmSelectOptions.php
index 557f0dd..8170420 100644
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -60,16 +60,16 @@ class CivicrmSelectOptions extends FormElement {
 
     $element['options'] = [
       '#type' => 'table',
-      '#tableselect' => FALSE,
+      '#tableselect' => TRUE,
```

**2. Comment out the javascript in:** https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/js/webform_civicrm_options.js#L59-L67

**3. Add a comment above this section -> that we're looking at using `#tableselect`** - so we're keeping the code for now until we're positive we can remove it

**4. Hide (i.e. `display:none`) the first column of checkboxes** as users need to hit the checkboxes in the enabled column if they wish to select some specific countries

And:
**5. Add a test to ensure the select all is working**

KarinG:
With `#tableselect' => TRUE,` -> the select all/deselect all is working nicely:

image

aaayyuusshh:
- updated CivicrmSelectOptions.php ✅ : Added #tableSelect => TRUE , since we are trying to use #tableSelect
- updated webform_civicrm_options.js ✅  : Commented out L59-67  (code relating 'select-all-civi-options') since we want to keep the code until we are 100% sure we can remove it.

js/webform_civicrm_options.js
```diff
@@ -56,6 +56,10 @@ if (typeof jQuery.fn.prop !== 'function') {
         }
       }).change();
 
+     //Currently we are looking at using #tableSelect thus, we are commenting out 'select-all-civi-options' code for now
+     //until we are certain that we can remove it.
+
```
KarinG: @aaayyuusshh - indent this one more space and also put a space between so //Current -> // Current and //until -> // until

KarinG:
I see the test is failing -> not sure yet why though.

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\CustomFieldSubmissionTest::testSubmitWebform
Behat\Mink\Exception\DriverException: Impossible to check the element with XPath "(//html/.//*
[self::input[not(translate(./@type, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz') = 'radio' or translate(./@type, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz')
```

KarinG:
Ah - I think I know!

KarinG:
In CustomFieldSubmissionTest.php -> we have a test that involves setting the options to 'static' (so that too will now have the select all checkbox).

https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/tests/src/FunctionalJavascript/CustomFieldSubmissionTest.php#L319-320

The error occurs in:
`/home/runner/work/webform_civicrm/webform_civicrm/tests/src/FunctionalJavascript/CustomFieldSubmissionTest.php:361`

Which is this line - so the test can not check a field 'Yes' - there likely is no such field
`$this->getSession()->getPage()->checkField('Yes');`

The thing to do is to recreate the test on an actual Drupal site and to follow its steps.

KarinG:
Check this out -> I downloaded the Artifacts -> 

image

Then opened the archive and looked at the latest screenshot generated for the CustomFieldSubmissionTest -> 
image

Notice that the radio button has 'Yes' and 'Yes' as options. So I agree with Github Actions - it's impossible to check the 'Yes' option :-)

Question is - how does using `#tableselect` -> cause this?

KarinG:
Digging in a little further -> I indeed see the `#tableselect` also applies to the live options (we only want it for static) -> 

image

KarinG:
Check out -> https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/src/Element/CivicrmSelectOptions.php#L102-L104

Here we say if we're using live options -> people should not be allowed to drag elements (up and down to resort them). Try adding a line in this if statement  that says:

`$element['options']['#tableselect'] = FALSE;`

KarinG:
Yeah - I see this when we're configuring the Custom Radio Field -> somehow this is where we end up with two Yes options.

image

Will look further in a bit - need to make coffee first!

js/webform_civicrm_options.js
```diff
@@ -64,7 +68,8 @@ if (typeof jQuery.fn.prop !== 'function') {
           $('input.civicrm-enabled, input.select-all-civi-options, input.select-all-civi-defaults').prop('checked', false);
         }
         $('input.civicrm-enabled').change();
-      });
+      }); 
+      */
```
KarinG, line 71: There is an extra space behind the ; here -> 
image

js/webform_civicrm_options.js
```diff
@@ -56,6 +56,10 @@ if (typeof jQuery.fn.prop !== 'function') {
         }
       }).change();
 
+     //Currently we are looking at using #tableSelect thus, we are commenting out 'select-all-civi-options' code for now
```
KarinG: There are extra spaces here in the line below:

image

KarinG:
I've reproduced the issue:

1. simple cg radio button
2. add to webform
3. switch to static
4. then save

-> note the extra No :-(

image

KarinG:
PS -> edit element still looks fine -> 
image

KarinG:
So we're doubling up with both [static]checkboxes as well as [static]radios when using `#tableselect`; static[select] works well.

jitendrapurohit:
@KarinG @aaayyuusshh looks like the duplicate options are because the code uses the default value callback method to save the options. Adding our custom value_callback to the option list fixes the problem - eg -

```
diff --git a/src/Element/CivicrmSelectOptions.php b/src/Element/CivicrmSelectOptions.php
index 557f0dd..742430d 100644
--- a/src/Element/CivicrmSelectOptions.php
+++ b/src/Element/CivicrmSelectOptions.php
@@ -97,6 +97,7 @@ class CivicrmSelectOptions extends FormElement {
           'group' => 'weight',
         ],
       ],
+      '#value_callback' => [get_called_class(), 'valueCallback'],
     ];
 
     if ($element['#civicrm_live_options']) {
```

js/webform_civicrm_options.js
KarinG: 👍

KarinG:
@aaayyuusshh -> don't forget to apply this one line mentioned here -> that will make the test ✅
https://github.com/colemanw/webform_civicrm/pull/587#issuecomment-899035238

aaayyuusshh:
@KarinG @jitendrapurohit  Just came across a bug (when #tableSelect=> TRUE in CivicrmSelectOptions.php)

- When all checkboxes are checked off in "Static Options" (using the top right checkbox), you cannot successfully switch back to "Live Options"
Screen Shot 2021-08-16 at 12 52 05 PM
Screen Shot 2021-08-16 at 12 52 19 PM

- When all checkboxes are unchecked (using the top right checkbox), now you can finally successfully switch back to "Live Options" 
Screen Shot 2021-08-16 at 1 04 12 PM
Screen Shot 2021-08-16 at 1 01 10 PM


- Also found out that if you just uncheck some checkboxes manually in Static Options, you still won't be able to successfully switch to Live Options

KarinG:
> @KarinG @jitendrapurohit Just came across a bug (when #tableSelect=> TRUE in CivicrmSelectOptions.php)
> 
>     * When all checkboxes are checked off in "Static Options" (using the top right checkbox), you cannot successfully switch back to "Live Options"
>       Screen Shot 2021-08-16 at 12 52 05 PM
>       Screen Shot 2021-08-16 at 12 52 19 PM
> 
>     * When all checkboxes are unchecked (using the top right checkbox), now you can finally successfully switch back to "Live Options"
>       Screen Shot 2021-08-16 at 1 04 12 PM
>       Screen Shot 2021-08-16 at 1 01 10 PM
> 
>     * Also found out that if you just uncheck some checkboxes manually in Static Options, you still won't be able to successfully switch to Live Options

I don't see this on my local - and the error message re: `File upload 8MB` suggests that something else is causing it. I think we don't need to worry about this at this time. But 👍 for detailing and noting this.

KarinG:
I need to update my local CiviCRM version [which could break things] so am going to add my notes here on how to get the CSS to work so that it will look like this:

image

1. define a new library called 'civicrmoptions' in webform_civicrm.libaries.yml - and add css
2. create that css file in the webform_civicrm/css directory and I think this works (is specific enough)->

```
/* Edit CiviCRM Options Element */
#webform-civicrm-options-wrapper .table-select .form-item {
  display:none;
}
```

3. finally attach the CSS to this edit element form -> in CivicrmSelectOptions.php -> add this line just before we return $element

    $element['#attached']['library'][] = 'webform_civicrm/civicrmoptions';
    return $element;
  }

Ok - now I can go and mess up my local env!

aaayyuusshh:
@KarinG I just made all those changes you mentioned above in my local and it looks like the screenshot you attached above.
Screen Shot 2021-08-16 at 2 28 08 PM


Do you want me to update the PR with these changes as well?

KarinG:
Awesome 🎉 and yes - please add those changes to this PR as well!

PR #591, opened 2021-07-31
D8/9 - Fix usage of webform_civicrm tokens in emails.

jitendrapurohit:
Overview
Fix usage of webform civicrm tokens in emails.

Before
WC has the following tokens in d7. None of them works in d8. To replicate - click on `Browse available token`

image

Lists the following tokens in d7 -

image

To replicate in drupal 8 - 

Add an email handler and try to use these tokens -

```
Existing Contact - [webform_submission:values:civicrm_1_contact_1_contact_existing] 
Activity 1 ID - [webform_submission:activity-id:1] 
Activity 2 ID - [webform_submission:activity-id:2] 
Webform CiviCRM Contacts IDs - [webform_submission:contact-id:1] 
Webform CiviCRM Contacts Links - [webform_submission:contact-link:1]
```

Another related bug -

Existing Contact field displays an ID in the results page -

image

After
The tokens are rendered correctly.

Existing contact displays a link to the civic contact.

image

Comments
@KarinG

### DIFF
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -9,6 +9,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\webform\Plugin\WebformElementBase;
 use Drupal\webform\WebformInterface;
 use Drupal\webform\WebformSubmissionInterface;
+use Drupal\Core\Url;
 
 /**
  * Provides a 'textfield' element.
@@ -559,6 +560,32 @@ class CivicrmContact extends WebformElementBase {
         $element['#empty_option'] = Xss::filter($element[$element['#options'] ? '#search_prompt' : '#none_prompt']);
       }
     }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function formatHtmlItem(array $element, WebformSubmissionInterface $webform_submission, array $options = []) {
+    $value = parent::formatHtmlItem($element, $webform_submission, $options);
+    $format = $this->getItemFormat($element);
+    $cid = $value['#plain_text'] ?? NULL;
+
+    if ($format === 'raw' || empty($cid) || !is_numeric($cid)) {
+      return $value;
+    }
+    $utils = \Drupal::service('webform_civicrm.utils');
+    $contact = $utils->wf_crm_apivalues('contact', 'get', ['id' => $cid], 'display_name');
+    if (!empty($contact[$cid])) {
+      if (empty($options['email']) && \Drupal::currentUser()->hasPermission('access CiviCRM')) {
+        unset($value['#plain_text']);
+        $cidURL = Url::fromUri('internal:/civicrm/contact/view', ['query' => ['reset' => 1, 'cid' => $cid]])->toString();
+        $value['#markup'] = t('<a href=":link">@name</a>', [':link' => $cidURL, '@name' => $contact[$cid]]);
+      }
+      else {
+        $value['#plain_text'] = $contact[$cid];
+      }
+    }
+    return $value;
   }
 
 }
--- a/tests/src/FunctionalJavascript/ExistingContactElementTest.php
+++ b/tests/src/FunctionalJavascript/ExistingContactElementTest.php
@@ -3,6 +3,7 @@
 namespace Drupal\Tests\webform_civicrm\FunctionalJavascript;
 
 use Drupal\Core\Url;
+use Drupal\Core\Test\AssertMailTrait;
 
 /**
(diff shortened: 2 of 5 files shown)


KarinG:
I’ll review this PR today. We’ve not had any problems sending out emails with tokens (and even ics attachement) for appointment reminders.

KarinG:
Thank you for working on this Jitendra/Fuzion 👍

Aashirbadd:
@jitendrapurohit Hey, I am trying to replicate this issue and I'm unsure how to "add an email handler" (I'm not familiar with CiviCRM mailing settings), and then get to the point in your screen where it says "E-Mail template" and gives you options to add tokens. By this do you mean "Mailhandler" which requires a module to add to Drupal? Or do you mean add a webform scheduled email handler. Thanks! Any help is appreciated!

KarinG:
@Aashirbadd -> when you're on a Webform -> hit the Settings tab (next to Build) -> and then add Email/handlers (sub-tab) -> then hit the +Add email blue button and you'll be ready to go!

Aashirbadd:
@KarinG Thanks! I'll try it out!

Aashirbadd:
@jitendrapurohit Hey, I can verify that the existing contact problem gets fixed with this PR.

Before:
Screen Shot 2021-08-04 at 11 52 10 AM

After:
Screen Shot 2021-08-04 at 11 52 57 AM

However, I cannot verify the other part, because I don't have tokens available to me on my local setup, only on my CiviCRM.io setup (where I don't have the CiviCRM specific tokens). Is there any extension I'm missing that allows me to add tokens on my local setup too? And also, how do I get to the "Email Template" settings?

jitendrapurohit:
@Aashirbadd You dont need any template. You can simply add an email handler and paste the text from the PR description in the Message section -

image

These tokens should render with actual values when you receive it in your inbox. Hit me up on Mattermost (jitendra-fuzion) if you need more info.

I'll try to clear the test build failure till then, its running fine on my side.

Thanks.

jitendrapurohit:
weird, adding a screenshot passes the build. Removing the last commit.

Aashirbadd:
@jitendrapurohit Hey! I'll try it out again with your suggestion and I'll let you know how it goes!

Aashirbadd:
@jitendrapurohit Hey, if I try to add those tokens in civiCRM.io, I recieve this:
image

However, if I setup these tokens on my local setup, it shows "This handler was successfully updated". The problem is, I still can't test if this works, because my local setup does not send emails like civiCRM.io does (even if I'm using the same handler). Is there a setup step I'm missing which is why this is not sending?

Here are my handler as well:
image

image

KarinG:
Hi @Aashirbadd - to facilitate testing, I've pulled this PR into civicrm.io

Aashirbadd:
@KarinG Thanks Karin! I'll test out if the PR fixes the issue!

Aashirbadd:
@jitendrapurohit @KarinG Hey, so I can verify that this PR fixes the issue with CiviCRM tokens!

So first, now I'm able to add the tokens with no error.
image

The Email is sent Successfully:
image
 
The Existing contact ID and Link is correct.

Same with the activity ID for both Activity 1 and Activity 2
(Activity 1 id = 128 verified)
Screen Shot 2021-08-09 at 2 42 17 PM
(Activity 2 id = 129 verified)
Screen Shot 2021-08-09 at 2 42 11 PM

KarinG:
Adding notes to this PR -> for documentation:

The tokens fixed in the PR are set that is specifically provided by webform civicrm. The tokens are - 

```
[webform_submission:values:civicrm_1_contact_1_contact_existing]
[webform_submission:activity-id:1] - eg aid1 is loaded from URL
[webform_submission:activity-id:2]
[webform_submission:contact-id:1]
[webform_submission:contact-link:1]

```

and case related tokens.

The first token adds display name of the contact 1 submitted on the webform. This is different from what data. provides since even if we dont have fname, lname fields available on the form, it can still print the display name by retrieving from civi. It was there in d7 as well, but wasn't working well in d8 which is fixed in the PR.

`[webform_submission:activity-id:1]` can be used to print id of the activity that was created with the webform submission. This isn't available in any of the core data tokens. Similarly `[webform_submission:activity-id:2]` is the second activity created on the form. We use this to build a custom url and include it in email.

`[webform_submission:contact-id:1]` can print the contact id created from the submission, even if the webform does not have an existing contact field in it. Helpful in eg if the webform is submitted anonymously and we need to send a checksum link with cid1 in response.

KarinG:
Tested & Reviewed -> merging! 🎉

KarinG:
Thank you for the extensive testing @Aashirbadd 👍

PR #593, opened 2021-08-16
D8/9 - Add test for dynamic custom fields

jitendrapurohit:
Overview
Test cover for #439 

Before
No test.

After
Test added.

Comments
This will fail the build until 439 is merged. It also fixes few notice errors on display.


tests/src/FunctionalJavascript/CustomFieldSubmissionTest.php
```diff
@@ -20,7 +20,7 @@ private function createCustomFields() {
 
     $params = [
       'custom_group_id' => $customgroup_id,
-      'label' => 'Text',
+      'label' => 'Label for Custom Text field',
       'name' => 'text',
```
KarinG, line 24: Yeah - I like that 🎉

KarinG:
Merging this - should pass now that I've merged #439

PR #595, opened 2021-08-16
Use composer v2

demeritcowboy:
Overview
and remove no-longer-supported option `--no-suggest` which does nothing in v2 and will be a fatal in v3

As discussed with @KarinG


KarinG:
I paused this trying to sort out the $9.75 != $10.00 issue in the RC -> re-running the tests now!

KarinG:
We hit that timeout again in GroupsTagsSubmissionTest.php - I've bumped it a bit.

KarinG:
Et voila! 🎉

PR #597, opened 2021-08-17
Remove Question Marks in Admin Form + add Description to Admin Screen

Aashirbadd:
Overview
Relating to the issue in: https://www.drupal.org/project/webform_civicrm/issues/3228229

Before
Screen Shot 2021-08-17 at 4 39 31 PM

After
Screen Shot 2021-08-17 at 4 40 05 PM

Comments
Me and @aaayyuusshh will work on adding the icons beside the select elements next!

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1318,7 +1318,7 @@ class AdminForm implements AdminFormInterface {
     $this->form['additional_options'] = [
       '#type' => 'details',
       '#group' => 'webform_civicrm',
-      '#title' => t('Additional Options'),
+      '#title' => t('Additional Settings'),
       '#attributes' => ['class' => ['civi-icon-prefs']],
     ];
     $this->form['additional_options']['checksum_text'] = [
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -427,7 +427,7 @@ class AdminForm implements AdminFormInterface {
       }
     }
   }
-fa-id-badge
+
   /**
    * Configure messages
    */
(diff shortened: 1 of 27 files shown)


src/AdminForm.php
```diff
@@ -1369,6 +1371,7 @@ private function buildOptionsTab() {
       }
       $this->form['additional_options']['confirm_subscription']['#description'] .= implode(', ', $ml) . '</em>';
     }
+
     else {
```
KarinG: Be careful adding lines/space -> best to delete this extra line here.

Aashirbadd: @KarinG Sounds good, I'll remove that extra line!

src/AdminForm.php
```diff
@@ -427,18 +427,18 @@ private function buildContactTab($n, $c) {
       }
     }
   }
-
+fa-id-badge
   /**
```
KarinG: This line/note is what's currently failing your test 🙂

Aashirbadd: I'll have that changed!

aaayyuusshh:
@Aashirbadd Hey, let's get to updating the outdated description next (:

Aashirbadd:
@KarinG I added spaces where there are break tags (new lines) for better readability, however I can change it back if needed!

Aashirbadd:
We seemed to have hit an issue with the icons, so we just focused on removing the question marks and updating the description. The icons should be replaced in a separate PR.

KarinG:
We just can't get any class into the HTML! Looking further into that.

PR #601, opened 2021-08-26
Check for empty fields.

puresyntax71:
Overview

When you have a custom group for "Membership" and no enabled fields, there is a warning when you enable "Membership" in the "CiviCRM" settings page:

> Notice: Undefined index: fields in Drupal\webform_civicrm\AdminForm->buildMembershipTab() (line 1038 of modules/contrib/webform_civicrm/src/AdminForm.php).

Before

Has error.

After

Checks if there are enabled fields.


KarinG:
Thank you 👍

PR #604, opened 2021-09-01
Don't create a blank billing address or email

MegaphoneJon:
Overview
When a billing address isn't required, Webform-CiviCRM will still create a blank billing address record.

Before
Selection_1211

After
Selection_1212

Comments
In the olden days, billing address was required, which is, I assume how this happened.

Note the open D7 version of this PR here: https://github.com/colemanw/webform_civicrm/pull/317.  Note also that there's a logic error in that PR that needs fixing before it's merged.


jitendrapurohit:
Tested. Thanks for the fix @MegaphoneJon 👍

PR #605, opened 2021-09-04
D8/9 - Remove contact paging if it is disabled from additional options

jitendrapurohit:
Overview
Remove contact paging if disabled from additional options

Before
- Disable Contact Paging was only implemented to skip adding contact page break on the webform. 
- contact_pagebreak is present in the `#parent` key even if the paging is disabled.

After
- Extended the fix so when Disable Contact Paging? is turned on - It removes the contact paging from the webform.
- Removes all traces of `contact_pagebreak` fielset on the webform.

Comments
@KarinG @MegaphoneJon Can you pls confirm if it works for you?

https://www.drupal.org/project/webform_civicrm/issues/3230899

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1996,7 +1996,7 @@ class AdminForm implements AdminFormInterface {
     $handler->setConfiguration($handler_configuration);
 
     $this->addEnabledElements($enabled);
-    $this->sortPaging();
+    $this->setPaging();
     // Update existing contact fields
     foreach ($existing as $fid => $id) {
       if (substr($fid, -8) === 'existing') {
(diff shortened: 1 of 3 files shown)


MegaphoneJon:
OK, this is a significant step in the right direction, but still has a few issues.

* If "Disable Contact Paging" is *not* ticked, then the `#webform_parents` property of an element still contains itself.  See the snippet below of element `parent_name_1` and how it has a #webform_parent of `parent_name_1`.  This continues to cause tokens to fail.

```yaml
    parent_name_1:
      '#type': flexbox
      '#format': container
      '#webform': agbu_san_fernando_valley_scouts_
      '#webform_id': agbu_san_fernando_valley_scouts_--parent_name_1
      '#webform_key': parent_name_1
      '#webform_parent_key': civicrm_1_contact_1_fieldset_fieldset
      '#webform_parent_flexbox': false
      '#webform_depth': 2
      '#webform_children': {  }
      '#webform_multiple': false
      '#webform_composite': false
      '#title': null
      '#admin_title': null
      '#webform_plugin_id': webform_flexbox
      '#webform_parents':
        - contact_pagebreak
        - civicrm_1_contact_1_fieldset_fieldset
        - parent_name_1
```

* When "Disable Contact Paging" *is* ticked, each save of the Webform-CiviCRM settings causes my "total" field on the payment page to add itself as a parent each time you save.  E.g. in the snippet below, I pressed "Save" three times:

```yaml
total:
  '#type': computed_twig
  '#title': Total
  '#admin_title': Total
  '#template': |
    {% set 	total = 0 %}
    {% if data.how_many_children_are_you_enrolling_ == 1 %}
    {% set total = 225 %}
    {% endif %}
    {% if data.how_many_children_are_you_enrolling_ >= 2 %}
    {% set total = 225 + ((data.how_many_children_are_you_enrolling_ - 1) * 200)%}
    {% endif %}
    ${{  total  }}.00

  '#whitespace': trim
  '#ajax': true
  '#webform': agbu_san_fernando_valley_scouts_
  '#webform_id': agbu_san_fernando_valley_scouts_--total
  '#webform_key': total
  '#webform_parent_key': ''
  '#webform_parent_flexbox': false
  '#webform_depth': 0
  '#webform_children': {  }
  '#webform_multiple': false
  '#webform_composite': false
  '#webform_plugin_id': webform_computed_twig
  '#webform_parents':
    - total
    - total
    - total
```

* Going from unticked to ticked on "Disable Contact Paging" removes the `contact_pagebreak` element - but causes all of its children to be reversed in order, and to lose their "Required" checkbox.

jitendrapurohit:
Hi @MegaphoneJon 

I was able to replicate

> lose their "Required" checkbox

and have updated the PR to fix.  Can you pls attach an export of your webform here? I can't replicate the other issues mentioned above. If Disable Contact paging is ticked, it removes contact_pagebreak on the webforms and i don't see any reversing of children elements.

Also, do you see this problem only on existing webforms? or also on new webforms created after this PR?

jitendrapurohit:
@MegaphoneJon Hmm, I see the keys like `webform_` should be not be set from the settings page. Eg - If I create a normal webform by

- Enabling contact fields from the civicrm tab.
- Add page manually from the build form. 
image

- There are no webform_parents, or any `webform_` keys added to the yaml.

So i think the same process should be through our settings page as well. I've made another update so none of these fields are added to the yaml from webform civicrm. I'll soon write a unit test to cover this scenario.

I still think the existing webforms will need to be fixed manually from the yaml config for tokens to work.

KarinG:
Hi @MegaphoneJon - it’s difficult to provide an upgrade path this and existing webforms will need some manual intervention. I do believe this is working as expected for new webforms now. Let me know if this is helpful and/or if we can do more.

jitendrapurohit:
I've made some more changes so that contact page_break doesn't cause any issues with the existing elements. @MegaphoneJon Are you able to give it a quick testing on your side? 

Thanks.

KarinG:
Hi @MegaphoneJon - I'm queuing up a new release - I think this is better IN then not, would you agree? If we need to fine tune further, you can always let us know. We're happy to circle back at a later date.

MegaphoneJon:
Oh sorry I hadn't seen the latest message from Jitendra.  Definitely better in than out!

KarinG:
Thanks Jon! Merging this 🎉

PR #423, opened 2020-12-24
7.x port of #420 (doPayment exception / doTransferCheckout)

mattwire:
Overview
----------------------------------------
See #420

Before
----------------------------------------

After
----------------------------------------

Technical Details
----------------------------------------

Comments
----------------------------------------
Port

PR #610, opened 2021-09-06
Port #420 - Remove doTransferCheckout

jitendrapurohit:
Overview
Port of #420.

Before

>Error: Call to protected method CRM_Core_Payment::doTransferCheckout() from context 'wf_crm_webform_postprocess' in wf_crm_webform_postprocess->submitIPNPayment() (line 1941 of .../sites/all/modules/webform_civicrm/includes/wf_crm_webform_postprocess.inc).

After
Fixed.

Comments
The title of https://github.com/colemanw/webform_civicrm/pull/423 says it was the port of #420, but it wasn't the actual port and didn't removed doTransferCheckout() call from the file? The function is still present on 7.x version and resulting in a fatal error for payments.


KarinG:
Happy to merge this!

PR #611, opened 2021-09-14
Need to map is_recur for payment processors not using propertyBag.

KarinG:
Overview
And using old/existing params in if statements (e.g. is_recur via online Contribution Pages, isRecur via Order API/propertyBag).

Before
Recurring Contributions w/ iATS not generating/storing payment_tokens - essentially breaking recurring payments with D8/D9 Webform CiviCRM module. No issue in D7 (it's using our version of Transact API).

After
Working - as far as I can tell at this point. 

Technical Details
propertyBag has it's own nomenclature (amount vs total_amount, isRecur vs is_recur). I was debating reverting to Transact API but for now it looks like this simple mapping works. 

Needs a test.


KarinG:
Merging this (lost my local branch!) + adding test shortly.

PR #616, opened 2021-09-27
[Test only] - Unit tests - Fix matrix versions and deal with scssphp

demeritcowboy:
Overview
`~5.N` will install the latest 5.x, even beyond "N", so `~5.39` actually installs 5.41 as of right now.

But then when installing 5.39 for real need to deal with scssphp problem in https://lab.civicrm.org/dev/drupal/-/issues/164

Before
Testing 5.41 three times.

After
Testing all different versions.

Technical Details
Composer

Comments
Composer

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,14 +12,12 @@ jobs:
       matrix:
         include:
           - drupal: '^8.9'
-            civicrm: '~5.33'
+            civicrm: '~5.35.0'
           - drupal: '^8.9'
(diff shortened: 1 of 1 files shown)


KarinG:
Oh that's a real failure - 5.33 though so will not be fixed

KarinG:
Can you please remove 5.33 from the matrix Dave?

KarinG:
Thanks Dave - this is a merge on pass ✅

PR #617, opened 2021-09-29
[Test only] - Adding recurring test -> check for Payment Token.

KarinG:
Overview
Adding recurring test -> check for Payment Token.


KarinG:
This is what we're testing -> 

image

If I comment out PR 611 - then the test fails 👍

PR #619, opened 2021-10-01
D8/9 - Group filter on autocomplete widget produces error on the webform.

jitendrapurohit:
Overview
Group filter on autocomplete widget produces error on the webform.

Before
To replicate -

- Add contact to the webform.
- Enable group filter on the contact element from the build page.
- Load the webform. Error -

image

After
Fixed.

Technical Details
'group' property key on any element makes the webform to call `Drupal\Core\Render\Element\RenderElement::processGroup()` which expects this param as a string. We use it for a different purpose, i.e, to filter the list of contacts and is in the array format. 

When this mismatch occurs, drupal displays a warning msg on the form. PR here sets the #process key in advance so `processGroup()`  is not called further in the code.

Another way of fixing it would probably be to update the `group` property key to eg `filter_group` but it requires a bigger step to modify it everywhere in the code and also handle upgrade to update its usage on existing webforms. I think this is the more safer approach for now since it's just an e-notice fix? Thoughts @KarinG


KarinG:
I completely agree with the safer fix. And thank you for making the test less abstract / more readable 🎉 ✅

PR #620, opened 2021-10-02
D8/9 - Ensure price field is always set for line items

jitendrapurohit:
Overview
Ensure price field is always set for line items

Before
Price field id is kept empty by webform CiviCRM since there is no price field attached to the payment done. Civi seems to trigger an error and requires it to be set for every line item.

After
Set price field on the line item to default quick config price field value.


KarinG:
Merging this ahead of a 6.0.0 release - will keep a close eye on the Actions run on it.

PR #622, opened 2021-10-03
[Test only] - D8/9 - Membership pay later test

jitendrapurohit:
Overview
Add test to verify dates for pay later memberships.

Before
No test.

After
Test Added.

Comments
@KarinG 🎉

### DIFF
--- a/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
+++ b/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
@@ -13,7 +13,7 @@ use Drupal\webform\Entity\Webform;
 final class ContributionPayLaterTest extends WebformCivicrmTestBase {
 
   /**
-   * Test Membership submission using Pay Later option.
+   * Test Membership submission using the Pay Later option.
    */
   public function testSubmitPaylaterMembership() {
     $this->createMembershipType(20);
--- a/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
+++ b/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
@@ -12,6 +12,62 @@ use Drupal\webform\Entity\Webform;
  */
 final class ContributionPayLaterTest extends WebformCivicrmTestBase {
 
+  /**
+   * Test Membership submission using Pay Later option.
+   */
+  public function testSubmitPaylaterMembership() {
+    $this->createMembershipType(20);
+
+    $this->drupalLogin($this->rootUser);
+    $this->drupalGet(Url::fromRoute('entity.webform.civicrm', [
+      'webform' => $this->webform->id(),
+    ]));
+    $this->enableCivicrmOnWebform();
+
+    $this->configureContributionTab(TRUE, 'Pay Later');
+
+    $this->getSession()->getPage()->clickLink('Memberships');
+    $this->getSession()->getPage()->selectFieldOption('membership_1_number_of_membership', 1);
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->getSession()->getPage()->selectFieldOption('civicrm_1_membership_1_membership_membership_type_id', 'Basic');
+    $this->htmlOutput();
+
+    $this->saveCiviCRMSettings();
+
+    $this->drupalLogout();
+    $this->drupalGet($this->webform->toUrl('canonical'));
+    $this->assertPageNoErrorMessages();
+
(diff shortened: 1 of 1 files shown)


agileware-fj:
I think this test is incorrect – the behaviour up until 5.40 (specifically [7670664e](https://github.com/civicrm/civicrm-core/commit/7670664e)) was for us that even pay later memberships would automatically have the start date and join date set to the current day, and the end date offset accordingly.

Per my comments on [mattermost](https://chat.civicrm.org/civicrm/pl/kprhjydg1bgh9fmzpzmtztpszy):

I would have expected 5.35 and 5.39 to fail that test, because join_date & start_date should be the current date, and end_date should be the n terms offset from that.
The empty dates only showed up for me after the 5.40 update

This test on the other hand appears to be asserting that the dates **should** be empty, contrary to this.

The behaviour for contribution pages and admin forms is also to set the dates for any new memberships, including with the pay later workflow.

One more thing - we're actually observing a dates problem on Drupal 7 (no D8/9 sites with Civi here); is it difficult to backport this test to the D7 version of the module?

KarinG:
@agileware-fj - thank you for adding your notes. 

We don't have a test infrastructure for D7. But what we do is backport fixes secured with tests in the D8/D9 version and passing all other (now 3-4 dozen tests) -> to D7. 

We will look further [and also test manually] - we believe you. Isolating it in a test is an easy way to help Eileen track down the issue and it also gives us reassurance that this will not be regressing in future.

jitendrapurohit:
ok - that seems to be an unintentional change of behaviour in CiviCRM. I've confirmed that the dates used to set correctly in 5.35. Not sure why this test did not fail. It was raised to check the different behaviors in these versions.

I've opened a gitlab ticket here https://lab.civicrm.org/dev/core/-/issues/2902. 

It can easily be fixed for us by setting a start_date and join_date in the membership API call. But would prefer discussion on the gitlab first since civi used to handle it automatically before 5.40.

jitendrapurohit:
ok - i just reversed the checks. It fails on my local as expected. With civi v5.41.2.

```
$ phpunit --filter testSubmitPaylaterMembership web/modules/contrib/webform_civicrm/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
PHPUnit 8.5.15 by Sebastian Bergmann and contributors.

Testing Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionPayLaterTest
E                                                                   1 / 1 (100%)

Time: 1.19 minutes, Memory: 16.00 MB

There was 1 error:

1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionPayLaterTest::testSubmitPaylaterMembership
Undefined index: join_date

/Users/jitendra/www/d9/web/modules/contrib/webform_civicrm/tests/src/FunctionalJavascript/ContributionPayLaterTest.php:66
/Users/jitendra/www/d9/vendor/phpunit/phpunit/src/Framework/TestResult.php:703

ERRORS!
Tests: 1, Assertions: 14, Errors: 1.
```

KarinG:
Oh I think we may be getting bits of 5.42 when we're testing 5.35 -> 

image

KarinG:
@jitendrapurohit -> Dave (@demeritcowboy) will do a PR for Tim to update his release process is just to specify a version in the action script (so that we don't accidentally mix and match CiviCRM versions).

demeritcowboy:
Maybe a small miscommunication there but yes that is an option. The easiest would be to simply specify a version for drupal-8 in the action script here, i.e. the yml workflow script.

demeritcowboy:
Hi @jitendrapurohit the drupal-8 stuff is more or less worked-out but FYI there is also https://github.com/colemanw/webform_civicrm/pull/625#issuecomment-939395973.

KarinG:
Ok - I believe the issues with the composer v1 vs v2 (i.e. composer decision making) are resolved with PR #627 - so I'm going to push a small commit to this branch to trigger a re-run.

KarinG:
That's exactly what we expect right now for this newly added Test -> no Join Date for 5.42.x and dev-master but ok for 5.35.x and 5.39.x

image

KarinG:
PR to address this regression by @eileenmcnaughton -> https://github.com/civicrm/civicrm-core/pull/21784

KarinG:
@jitendrapurohit - please go ahead and merge this if you feel it provides sufficient coverage for pay-later membership.

PR #623, opened 2021-10-09
specify drupal-8 version

demeritcowboy:
Overview
Install same version of civicrm-drupal-8 as the version of civicrm-core being installed.

Previously it was being installed indirectly as a requirement of webform_civicrm. And also there's nothing in civicrm-drupal-8's composer.json that ties it to civicrm-core's version.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,13 +11,13 @@ jobs:
     strategy:
       matrix:
         include:
-          - drupal: '^8.9'
-            civicrm: '~5.35.0'
-          - drupal: '^8.9'
-            civicrm: '~5.39.0'
-          - drupal: '~9.2.0'
+          - drupal: '8.9.*'
+            civicrm: '5.35.*'
+          - drupal: '8.9.*'
+            civicrm: '5.39.*'
+          - drupal: '9.2.*'
             civicrm: '5.42.x-dev'
-          - drupal: '~9.2.0'
+          - drupal: '9.2.*'
             civicrm: 'dev-master'
     name: Drupal ${{ matrix.drupal }} | CiviCRM ${{ matrix.civicrm }}
     services:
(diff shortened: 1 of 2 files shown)


KarinG:
Great 🎉 - and can we also do something like 5.35.* (so that it doesn't look like we're testing 5.35.0?

demeritcowboy:
> 
> 
> Great 🎉 - and can we also do something like 5.35.* (so that it doesn't look like we're testing 5.35.0?

To do that would need to put the exact version in the matrix. For 5.35 we know there isn't going to be another 5.35 release so yes. I'll update.

KarinG:
That leaves 5.39.0 though. Why does ^8.9 work to get the latest Drupal 8.9.x but ^5.35 gave us 5.41? Or am I misremembering? I find this SO confusing!

demeritcowboy:
Actually `*` might work. Let's try that.

KarinG:
Fingers crossed 🤞

KarinG:
I think it worked?!

demeritcowboy:
Yes it appears to be working. Although why does dev-master only run 3 tests and skip 2 of them?

The problem earlier with `^5.39` installing 5.41 was that yeah it's not intuitive how composer restraints work without the 3rd component.

demeritcowboy:
Is `drupal/webform_civicrm (5.0.0-beta3)` right?

KarinG:
Where do you see `drupal/webform_civicrm (5.0.0-beta3)`?

demeritcowboy:
At https://github.com/colemanw/webform_civicrm/runs/3847640368?check_suite_focus=true#step:12:10

I'm tempted to think it's just a one-time drupal-packagist  / network glitch since the other runs installed beta8.

KarinG:
So weird... 

That entire 9.2.* w/ dev-master CiviCRM looks weird [only 2 tests; no artifacts; and webform -beta3);

KarinG:
Going to hit re-run to see what happens

KarinG:
Same issue -> and it finishes early b/c only a handful of tests. Have never seen this before [successfull in 1min]

image

KarinG:
Provisional merge!

.github/workflows/main.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       - name: Install webform_civicrm
         run: |
           cd ~/drupal
-          COMPOSER_MEMORY_LIMIT=-1 composer require drupal/webform_civicrm *@dev
+          COMPOSER_MEMORY_LIMIT=-1 composer require drupal/webform_civicrm:dev-5.x@dev
       - name: Install token
```
KarinG, line 83: @demeritcowboy - I just checked some other Matt github actions and it does look like -> this is correct:
`           COMPOSER_MEMORY_LIMIT=-1 composer require drupal/webform_civicrm *@dev`

civicrm entity e.g. has:
`          COMPOSER_MEMORY_LIMIT=-1 composer require drupal/civicrm_entity *@dev`

PR #624, opened 2021-10-09
[Ad-hoc PR] - Temporary workaround for restriction in composer.json

demeritcowboy:
Overview
composer.json is currently restricting civicrm-drupal-8 to `~5.0`, which means it can't install version dev-master. It's been updated but the packagist registry (or wherever composer is looking - maybe a cache) still has the restriction.

Can revert this once that has worked itself out.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,11 @@ jobs:
       - name: Install CiviCRM ${{ matrix.civicrm }}
         run: |
           cd ~/drupal
-          COMPOSER_MEMORY_LIMIT=-1 composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages,drupal-8}:${{ matrix.civicrm }} --prefer-dist
+          COMPOSER_MEMORY_LIMIT=-1 composer require civicrm/civicrm-asset-plugin:'~1.1' civicrm/civicrm-{core,packages}:${{ matrix.civicrm }} --prefer-dist
+          # Todo remove the next 3 lines when drupal packagist gets updated with the new composer.json requirement and put drupal-8 back into the line above.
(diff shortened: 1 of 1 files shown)


KarinG:
Oh is this working? :-) Checking the Checks -> yes I think looks good.

Anything else we need to check before I merge this?

Downloading drupal/recommended-project (8.9.19)
  - Locking civicrm/civicrm-core (5.35.2)
  - Locking drupal/webform_civicrm (dev-5.x 6586c0b)

Downloading drupal/recommended-project (8.9.19)
  - Locking civicrm/civicrm-core (5.39.1)
  - Locking drupal/webform_civicrm (dev-5.x 6586c0b)

Downloading drupal/recommended-project (9.2.7)
  - Locking civicrm/civicrm-core (5.42.x-dev 32fff2f)
  - Locking drupal/webform_civicrm (dev-5.x 6586c0b)

Downloading drupal/recommended-project (9.2.7)
  - Locking civicrm/civicrm-core (dev-master 5aaf127)
  - Locking drupal/webform_civicrm (dev-5.x 6586c0b)

demeritcowboy:
There's something related this is making me think about but it doesn't change the PR. Noting that 6586c0b is the same as beta8, like before.

KarinG:
Sorry not following... what do we need to do next? 🙂

KarinG:
This is where I found that line -> Locking drupal/webform_civicrm (dev-5.x 6586c0b)

image

demeritcowboy:
I think the goal here is met - use a matching version of civicrm-drupal-8 where possible. But I'm confused about the role that beta8 plays in the tests, but that's no different from before, it just displays as a git hash instead of version string.

KarinG:
Ok awesome - merging! 🎉 🥳

PR #625, opened 2021-10-09
what

demeritcowboy:
Overview
Just running some attempts to see what is what.


demeritcowboy:
I think the latest commit solves it but I haven't squished yet just to keep the evidence. It seems to be a change from composer v1 to v2, where since the tests were changed to use v2 it doesn't use the PR branch it uses the latest beta.

This is the line that I think confirms the latest commit is working:
`- Installing drupal/webform_civicrm (dev-38370779aef0405d1f6dc215330fd41530a46446): Symlinking from /home/runner`

It matches the hash from the checkout plugin and is obviously using a local symlink. And the previous commit output confirms the base.php file being run is the right file: https://github.com/colemanw/webform_civicrm/runs/3849315149?check_suite_focus=true#step:17:22 and https://github.com/colemanw/webform_civicrm/runs/3849315149?check_suite_focus=true#step:12:53

PR #626, opened 2021-10-10
Adjust for change in repo priority in composer v2

demeritcowboy:
Overview
Clean version of #625. The ordering of the repos is more important now.

Before
Script was using the latest beta version not the PR branch.

After
It checks out the PR branch.

Technical Details
The way the script works, it uses the checkout plugin to check out the PR branch locally. Then it configures composer with a virtual repo that points to that local folder. So when it requires webform_civicrm it installs from that local repo instead of the real one at drupal-packagist. When the script was changed to use v2, it started using the real repo, and so was always checking out the latest beta not the PR branch.

Also should now be able to remove the workaround for drupal-8 (needed because before the composer.json was requiring `~5.0` which prevented installing dev-master when the matrix specified master).

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,8 +59,9 @@ jobs:
           composer config minimum-stability dev
           composer config prefer-stable true
           composer config preferred-install dist
-          composer config repositories.0 path $GITHUB_WORKSPACE
-          composer config repositories.1 composer https://packages.drupal.org/8
+          # Note in composer v2 the order is more important. We want the path one to have higher priority.
+          composer config repositories.0 composer https://packages.drupal.org/8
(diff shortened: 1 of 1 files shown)


KarinG:
Sorry - I think I have a local path left here -> https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/tests/src/FunctionalJavascript/ContributionIatsTest.php#L344

demeritcowboy:
He he yep.
And the other fails I think are real too.

KarinG:
@jitendrapurohit - these failures are our too priority now👆

KarinG:
@jitendrapurohit - I’ll merge this so that every test going fwd will use branch (as opposed to beta8.

PR #627, opened 2021-10-10
Fix test failures after #626

jitendrapurohit:


KarinG:
Consistently fails on 5.35 -> but just passed on 5.39 [which is consistent with @jitendrapurohit telling us it passes on 5.41 on his local]

image

KarinG:
Ok locating ->      `$this->createScreenshot($this->htmlOutputDirectory . 'loaded_webform.png');`

Not in the browser artifacts -> why not? 

image

Ah perhaps b/c we only get one screenshot per combination of Drupal and CiviCRM [but we do get 4 html-s] - checking there now.

KarinG:
These are different (5.35 and 5.39) -> 
image

KarinG:
We really need to find  a way to print Drupal and CiviCRM version on all Webforms - easier to untangle the brower output HTML. Adding that to my list.

KarinG:
Ok I just tested that I can push a commit to your branch. Going to work on getting Drupal and CiviCRM version into our Webforms so the screenshots are easier to decipher.

KarinG:
So that ContributionDummyTest -> indeed only fails on 5.35:

image

KarinG:
Ok we now have clearly marked html output -> (left -> 5.39.1; right -> 5.35.2). 

Note for Basic Membership -> we don't have a contact in 5.35.2 -> 

image

KarinG:
Not sure why we're not getting any 'loaded_webform' screenshots? Perhaps there is a limit? We need a local D8/5.35 local set up. Ok on that!

KarinG:
Tested on 5.35 - manually (as uid 1) and that all works. So let's try bump the permissions.

KarinG:
Ok -> now reproduced on my d8civicrm.local with D8.9.x and CiviCRM 5.35.2 -> within test env -> 

image

KarinG:
Ok bumping permissions worked locally! 

image

KarinG:
Fingers 🤞🤞

KarinG:
YES - 🎉 🥳

KarinG:
Just pushed a small clean up. 

I do not understand yet why the .png-s are not showing up in the artifacts. 

```
    $filename = 'loaded_webform' . substr(sha1(rand()), 0, 7) .'.png';
    $this->createScreenshot($this->htmlOutputDirectory . $filename);
```

PR #632, opened 2021-10-25
Evaluate checksum on webform submit

jitendrapurohit:
Overview
Evaluate checksum on submitting the webform.

Before

If a checksum element is present on the webform, it does not evaluate the cs value for the contact on submission. Tokens, etc are not rendered correctly. To replicate

- Create a contact with Fname, Lname and Checksum field
- Submit the webform.
- Result page displays 7 (cs limit) in the checksum column.

image

After

Fixed.

image

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -2299,6 +2299,7 @@ class AdminForm implements AdminFormInterface {
     if (isset($field['value_callback'])) {
       $method = 'get_default_' . $table . '_' . $name;
       $field['value'] = self::$method($field['form_key'], $options);
+      unset($field['value_callback']);
     }
     // For hidden+select fields such as contribution_page
     if ($field['type'] == 'hidden' && !empty($field['expose_list']) && !empty($settings[$field['form_key']])) {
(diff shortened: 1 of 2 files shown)


mpaulso:
I have tested PR632 and it works as described.
image

src/AdminForm.php
```diff
@@ -2299,6 +2299,7 @@ public static function insertComponent(&$field, &$enabled, $settings, $create_fi
     if (isset($field['value_callback'])) {
       $method = 'get_default_' . $table . '_' . $name;
       $field['value'] = self::$method($field['form_key'], $options);
+      unset($field['value_callback']);
     }
```
KarinG, line 2303: @jitendrapurohit - why do we need this here?

jitendrapurohit: @KarinG If we dont remove this here, it ends up in the yml `'#value_callback': true` where this property isn't supported. So you get something like this when you try to edit any element from the build page.


image

The main requirement of this was to set a default value which i see is done just above this unset line 

    $field['value'] = self::$method($field['form_key'], $options);

Once that is done, i think we can unset it so it isn't stored in yml. Hope that helps :).

KarinG:
I'm going to merge this [@mpaulso has confirmed it works as advertised] even though the functionality could be locked in with a test.

PR #633, opened 2021-10-25
Multiple Case role selection on case configuration

jitendrapurohit:
Overview
Multi select case roles in webform

Before
On d7 - we were able to select multiple contacts for case roles -

image

On d9 - we can't.

After
Fixed -

image


KarinG:
I think this is ok without being secured in a test!

PR #636, opened 2021-11-01
D8/9 - Fix the submission of a locked field when it has a default value

jitendrapurohit:
Overview
Fix the submission of a locked field when it has a default value set on it.

Before
Locked field + submit disabled field setting does not submit a value to civi. To replicate -

- Enable a contact with `Relationship Description` field on a webform.
- Edit the existing contact element and lock the Relationship field set from `Fields to Lock` setting.
- Enable `Submit disabled field value` setting.
- Edit the `Relationship Description` field element and set a default value on it.
- Load the form. The element is hidden.
- Submit the form.
 
After
The value is saved.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2606,12 +2606,18 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         if (in_array($type, $component['#hide_fields'])) {
           $value = wf_crm_aval($this->loadContact($c), "$table:$n:$name");
           // Check to see if configured to Submit disabled field value(s)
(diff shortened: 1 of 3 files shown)


mpaulso:
Was able to confirm that a relationship default description is locked in the data base when it has an existing value.

For the language in the tool tip, should this be ‘locked’ rather than disabled for the field values?

image

mpaulso:
Karin applied this PR, we set the default for the relationship description, but default relationship description was not saved in database [likely b/c we already had an existing value]

image

mpaulso:
If we remove the existing value of the relationship description, this successfully saves the default value configured in the webform.
image

KarinG:
Hi @jitendrapurohit - merging this - I'll put up a small follow up PR for the language edits.

PR #637, opened 2021-11-02
D8/9 render tags/symbols in status message

jitendrapurohit:
Overview
HTML tags/symbols are not rendered in status message.

Before
HTML tags/symbols are not rendered in status message. To replicate

- Create a membership type with label "A & B".
- Add a membership for the contact.
- Create a webform and enable Membership type with user select option.
- Load the webform, the status message displayed above render as `A &amp; B`, something like -

Screenshot 2021-11-02 at 6 18 59 PM

After
Fixed.


KarinG:
I don't think this needs test -> it carries no risk to critical functionality. Merging! 🎉

PR #638, opened 2021-11-02
D8/9 - extend membership count to 10

jitendrapurohit:
Overview
Extend the range of membership to 10.

Before
Only 9 memberships can be created on a contact.

After
Extended to 10.

Comments
We use this on live from few weeks. ok to have this in core @KarinG?

 I'll add a test soon.


KarinG:
Absolutely ok - and no test required.

PR #639, opened 2021-11-03
D8/9 - Page break labels are overwritten on civicrm settings page submit

jitendrapurohit:
Overview
Page break element titles are replaced by default when Civi tab is saved.

Before
Page break element titles are replaced by default when Civi tab is saved. To replicate.

- Create a webform with Enable contribution = Yes.
- Change the label of contribution page break to something else eg `Payment Details`.
- Navigate to civi tab again. Add any field.
- Contribution page break label is replaced with the default `Payment`.

After
Fixed.


KarinG:
Awesome - thank you 🥳

Existing automated tests ✅
Manual testing ✅
Live site confirmation ✅

Screenshot:

image

PR #640, opened 2021-11-05
D8/9 - Add option to enable public groups on the webform

jitendrapurohit:
Overview
Add option to enable public groups on the webform

Before
https://chat.civicrm.org/civicrm/pl/u9pouzzjcidmbmf3dnmmcuwfsw

>iirc for Webform - currently we have the Groups field we can set to 'user select' but that then puts all groups on show, and one has to manually untick all non-public ones if the goal is to have this field function as a newsletter sign up group - correct? if so, do others see the benefit in having an option in the setting dropdown for eg
user selects (all groups)
user selects (public groups only)
and then hopefully that list would change on the form dynamically as Public Groups were added/removed. thoughts?
(edited)

After
Group field has a new option -

image

If chosen - only public groups are loaded on the webform.

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1508,6 +1508,9 @@ class AdminForm implements AdminFormInterface {
       if ($field['type'] != 'hidden') {
         $options += ['create_civicrm_webform_element' => t('- User Select -')];
       }
+      if ($name == 'group') {
+        $options += ['public_groups' => t('- Public Groups -')];
+      }
       $options += $utils->wf_crm_field_options($field, 'config_form', $this->data);
       $item += [
         '#type' => 'select',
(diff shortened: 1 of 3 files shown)


KarinG:
Hi @jitendrapurohit - so -> "and one has to manually untick all non-public ones" is not necessary any more (we have since extended the native webform select all / select none option which will quickly let someone decide what groups to present in the static option. So I'm wondering if this is adding complexity which we don't really need as it's all easily configurable now?

KarinG:
PS - test failure for 5.35.* and 5.43.x-dev appear unrelated.

KarinG:
If one wanted all public groups on the form all the time (i.e. dynamic/live) - then I agree this would make sense. 

Can you please rename to:

```
- User Select - 
- User Select (public groups) -
```

jitendrapurohit:
closing this for now since we have a toggle_all checkbox on the build page.

PR #621, opened 2021-10-03
D7 port fix - Ensure price field is always set for line items

jitendrapurohit:
Port for https://github.com/colemanw/webform_civicrm/pull/620
Drupal issue -> https://www.drupal.org/project/webform_civicrm/issues/3236621

PR #641, opened 2021-11-06
Fix undefined variable from PR621

jitendrapurohit:
Overview
Fix undefined variable from https://github.com/colemanw/webform_civicrm/pull/621. I think this was a miss in porting the PR from d8. `$utils` isn't available in d7 code so shouldn't be present. @KarinG Can you pls merge?


KarinG:
Thank you for catching that! Merging.

PR #644, opened 2021-11-15
Add support for contact image.

puresyntax71:
Overview
This is a fix for this issue - https://www.drupal.org/project/webform_civicrm/issues/3248702.

Before
Contact Image disabled.

After
Make contact image work.

Comments
Implementation is a bit different as compared to the old one as it seems that public allowed is default in the Drupal 7 version. The Drupal 7 implementation loads the contact image from the **Drupal file system** while this one loads it from the **CiviCRM file system** since private is the default storage for images and CiviCRM isn't able to access it.

~Currently, loading of the image upon webform submission doesn't work though i.e. when you create a new submission and the form contains file elements, custom file elements are loading the default image correctly although the contact image isn't loading the correct one - https://take.ms/KZb27.~

~I've added a `@todo` temporarily here - https://github.com/colemanw/webform_civicrm/pull/644/files#diff-c92b516a71a531e151759b309eb473bf58592139e399edb412d9be9217960d0aR801.~

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -718,6 +718,9 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     if ($params['contact_type'] == 'Individual' && empty($params['first_name']) && empty($params['last_name'])) {
       $params['display_name'] = $params['sort_name'] = empty($params['nick_name']) ? $contact['email'][1]['email'] : $params['nick_name'];
     }
+    if (isset($params['image_URL'])) {
+      \CRM_Contact_BAO_Contact::processImageParams($params);
+    }
     $result = $utils->wf_civicrm_api('contact', 'create', $params);
     return wf_crm_aval($result, 'id', 0);
   }
@@ -759,6 +762,9 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       }
     }
     $params['id'] = $this->ent['contact'][$c]['id'];
+    if (isset($params['image_URL'])) {
+      \CRM_Contact_BAO_Contact::processImageParams($params);
+    }
     $utils->wf_civicrm_api('contact', 'create', $params);
   }
 
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -718,9 +718,6 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     if ($params['contact_type'] == 'Individual' && empty($params['first_name']) && empty($params['last_name'])) {
       $params['display_name'] = $params['sort_name'] = empty($params['nick_name']) ? $contact['email'][1]['email'] : $params['nick_name'];
     }
-    if (isset($params['image_URL'])) {
-      \CRM_Contact_BAO_Contact::processImageParams($params);
-    }
     $result = $utils->wf_civicrm_api('contact', 'create', $params);
     return wf_crm_aval($result, 'id', 0);
   }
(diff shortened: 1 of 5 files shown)


src/WebformCivicrmPostProcess.php
```diff
@@ -718,6 +718,9 @@ private function createContact($contact) {
     if ($params['contact_type'] == 'Individual' && empty($params['first_name']) && empty($params['last_name'])) {
       $params['display_name'] = $params['sort_name'] = empty($params['nick_name']) ? $contact['email'][1]['email'] : $params['nick_name'];
     }
+    if (isset($params['image_URL'])) {
+      \CRM_Contact_BAO_Contact::processImageParams($params);
+    }
     $result = $utils->wf_civicrm_api('contact', 'create', $params);
     return wf_crm_aval($result, 'id', 0);
   }
```
KarinG: We've been working hard to remove all BAO function calls (replacing them with API calls). There are only few left to go! So we should not be adding any new ones. Can you replace this with an API call?

KarinG:
@jitendrapurohit commented: "I just did a manual test and the image is stored correctly in civi." and also noted concerns about the BAO function call.

puresyntax71:
Yeah, I can't seem to find alternatives to using the CiviCRM API. I used Drupal API instead for getting the image URL.

Yeah, tried to add a test although for some reason, the "Upload image" field isn't being rendered in the HTML when I was trying to investigate it. Would be nice to get some tests working. So far these are my changes:

```diff
     $configurable_contact_field_groups = [
+      'contact' => [
+        'image_URL',
+        'nick_name',
+      ],
       'address' => [
         'street_address',
         'city',
@ tests/src/FunctionalJavascript/ContactSubmissionTest.php:401 @ final class ContactSubmissionTest extends WebformCivicrmTestBase {
     // address => 'street_address' or 'city'
     foreach ($configurable_contact_field_groups as $field_group => $field_value_key) {
       if (isset($contact_values[$field_group])) {
-        $this->assertTrue(is_array($contact_values[$field_group]));
-        $this->assertTrue(isset($contact_values[$field_group][0]));
-        $this->getSession()->getPage()->selectFieldOption('contact_1_number_of_' . $field_group, count($contact_values[$field_group]));
+        if ($field_group === 'contact') {
+          $this->assertTrue(is_array($contact_values[$field_group]));
+          $this->assertTrue(isset($contact_values[$field_group]));
+        }
+        else {
+          $this->assertTrue(is_array($contact_values[$field_group]));
+          $this->assertTrue(isset($contact_values[$field_group][0]));
+          $this->getSession()->getPage()->selectFieldOption('contact_1_number_of_' . $field_group, count($contact_values[$field_group]));
+        }
```

jitendrapurohit:
> Yeah, tried to add a test although for some reason, the "Upload image" field isn't being rendered in the HTML when I was trying to investigate it

@puresyntax71 seems you need to enable more modules from https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php#L16 (or override it in local file)? eg 

```
'file',
'file_module_test',
'field_ui'
```

I've retested with the latest PR and yes, the image upload works fine 👍 

I think we're good to merge as soon as the test is up @KarinG

KarinG:
I'm going to pull this into a client's instance and give this a live-test as well. Then I'll merge it. Not having an automated test will mean this functionality could break in future (so would be great if you could circle back to it) - but I've seen a few people looking for this functionality - so I'm happy to merge this PR without a test.

KarinG:
Existing automated tests ✅
Manual testing ✅
Live site confirmation ✅

Screenshots:

image

image

image

image

KarinG:
@puresyntax71 - thank you for this PR 🎉 🥳

PR #501, opened 2021-03-07
Do not update Drupal email address.

KarinG:
Overview
----------------------------------------
https://www.drupal.org/project/webform_civicrm/issues/2621918

Before
----------------------------------------
Dubious code

After
----------------------------------------
Removed 🎉

Technical Details
----------------------------------------
Really not necessary - Drupal has its own forms to update user's email addresses 🙂

PR #645, opened 2021-11-18
Duplicate email addresses created when users with an account add an email via CiviCRM webform

JKingsnorth:
For full details, please see: https://www.drupal.org/project/webform_civicrm/issues/3250255


colemanw:
Ok. This keeps the block unlike #501 which removes it completely in the D9 branch. But that's probably a good call for the D7 branch which shouldn't be getting any major changes at this point.

PR #647, opened 2021-11-28
D8/9 - Enable waitlist registration via webforms

jitendrapurohit:
Overview
Enable waitlist registration via webforms

Before
Can't submit a waitlist registration

After
Participants can be submitted with status = `On Waitlist` via webforms.

1. Extended this setting to support waitlist registrations -

image
 
2. So if an event has space available = 2 and webform is registering 3 participants to an event, the first 2 are registered and the 3rd one is added to the waitlist (Test added to assert this behaviour).
3. Waitlist text from CiviCRM is displayed on the webform if event is full.

image

Comments
Drupal Ticket - https://www.drupal.org/project/webform_civicrm/issues/3248788

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -897,7 +897,7 @@ class AdminForm implements AdminFormInterface {
     $this->form['participant']['reg_options']['show_remaining'] = $field;
     $this->form['participant']['reg_options']['validate'] = [
       '#type' => 'checkbox',
-      '#title' => t('Prevent Registration for Past/Full Events'),
+      '#title' => t('Prevent Registration (enable waitlist) for Past/Full Events'),
       '#default_value' => (bool) wf_crm_aval($this->data, 'reg_options:validate'),
     ];
     $this->help($this->form['participant']['reg_options']['validate'], 'reg_options_validate');
--- a/src/AdminHelp.php
+++ b/src/AdminHelp.php
@@ -227,7 +227,10 @@ class AdminHelp implements AdminHelpInterface {
 
   protected function reg_options_validate() {
     return '<p>' .
-      t('Will not allow the form to be submitted if user registers for an event that is ended or full.') .
+      t('Will not allow the form to be submitted if user registers for an event that is ended or full.').
+      '</p>'.
+     '<p>' .
+      t('If an event offers a waitlist, the form will be submitted and the participant will be added to a waiting list.').
       '</p>';
   }
 
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -29,6 +29,7 @@ abstract class WebformCivicrmBase {
   protected $data = [];
   protected $ent = [];
   protected $events = [];
+  protected $waitlist_events = [];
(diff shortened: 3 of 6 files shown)


MegaphoneJon:
Hi Jitendra,

This looks like a step in the right direction.

We shouldn't be offering waitlists on events that have ended.  I'm also thinking of use cases where the event ID is set to "User Select", registration for multiple events when at least one is waitlisted and one isn't, and payment on waitlisted events.

I think the spec here is more complex than the implementation, and I'm not even sure how Civi core handles waitlists on paid events.

KarinG:
Hi Jon @MegaphoneJon - I can see this can get quite out of control :-) Are you ok with us merging this as a useful step to start providing some waitlist functionality?

jitendrapurohit:
I've rebased the PR.

> We shouldn't be offering waitlists on events that have ended

@MegaphoneJon Can you clarify pls? I get this if i try to register for past event -

image

So i think this is already handled? Do you get waitlisted for any of your past events? If yes, can you pls add steps to reproduce that usecase?


 Re multiple events, the current code in webform civicrm will submit the webform without any errors and add the participant to the waitlist.

**Case 1:** If there are multiple events and all support waitlist - No error is displayed and webform is submitted with all participants status = "On waitlist".

**Case 2:** Multiple Events with max participant reached + not all offer a waitlist - On submission, the webform displays a validation error as follows.

image


This was the same behaviour before this PR which i think is also correct.

**Case 3:** : Eg - Event 1 has 2 space remaining. Event 2 is full. Both offer a waitlist - In this case, the first participant registered on event 1 is set to `Registered` and all other participants are set to `On waitlist`.


Any other case that you want to support? I think its ok to get this merged since it provides the basic support of waitlisted participants?

MegaphoneJon:
I only evaluated one permutation, but it didn't pass, so I think the next webform_civicrm version should be released without this.

My form had 3 CiviCRM contacts, events enabled, all contacts register for same event, event is User Select, my test event had max participants of 1.
With this patch applied:
* I registered 3 people.
* I then registered a further 3 people.
All 6 registrations went through.

Without the patch:
* On the first submission  but after fixing #678, I got a proper error message that there weren't enough spaces.  Without #678 I got a WSOD.
* On the second submission, I got the proper message that the event was full.
I'm attaching an export of the form I used for testing.
[647 webform.yml.txt](https://github.com/colemanw/webform_civicrm/files/7781200/647.webform.yml.txt)
.

KarinG:
@MegaphoneJon - Thank you for your review Jon. I've tagged this PR as needs work and will go ahead and tag a release without it.

dptarrant:
D7 Webform event registrations do not respect the online registration end date.  Is this an opportunity to rectify that as well?  Assuming D8/9 have the same issue?
https://civicrm.stackexchange.com/questions/16889/does-webform-civicrm-event-registration-respect-the-registration-end-date-sett

MegaphoneJon:
@dptarrant That work would be unrelated, though it's not a bad idea.

PR #648, opened 2021-12-02
Remove fixed value that makes relationship always active.

KarinG:
Overview
When adding a Relationship -> the in_active field had a "value" = 1 (always) so you could never make a relationship inactive.

Removing the 1 from the Build tab -> Element config works. The is_active element can now be toggled to ON or OFF without it defaulting to 1 (ON) all the time.


jitendrapurohit:
seems fine. I've tested and confirmed both active and inactive submissions after the patch. merging.

PR #654, opened 2021-12-12
Webform use release instead of dev

KarinG:
Overview
We started noticing test failures a few days ago. The test that failed consistently across all Drupal and CiviCRM versions is one of our integration tests -> `testTokensInEmail`

Taking a screenshot during the test (Github actions) -> shows CKEditor does not appear to be loading properly:

image

Closer inspection -> we were testing against webform-dev (entire matrix). So I did:

Before
The build includes composer require 'drupal/webform:6.x-dev@dev' -> were ❌

After
The build now includes composer require 'drupal/webform:^6.1' ✅

Comments
There were just a handful of commits in the webform repo the past couple of days. Will open an issue on drupal.org to alert Jake.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,8 @@ jobs:
       - name: Ensure Webform ^6.0
         run: |
           cd ~/drupal
-          COMPOSER_MEMORY_LIMIT=-1 composer require 'drupal/webform:6.x-dev@dev'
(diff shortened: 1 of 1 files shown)


KarinG:
Ok confirming we are now using:
  - Installing drupal/webform (6.1.2): Extracting archive

KarinG:
Yep -> that's it 🎉 Going to file an issue in the webform queue. And will merge this for us (we can go back to webform -dev testing in future but for now this is ✅)

PR #655, opened 2021-12-13
Add AJAX contribution.

puresyntax71:
Overview
When AJAX is enabled for a webform and contribution is enabled, the contribution doesn't work. The line items are not being updated and the billing block is not loaded.

The AJAX contribution is found under Settings > General > AJAX settings. This has been only manually tested though.

I don't seem to find this on the d.o issue queue.

Before

Contribution not working when AJAX is enabled.

After

Contribution is working when AJAX is enabled.

Technical Details

The relevant JS in webform_civicrm_payment.js responsible for updating the line item total and loading billing block is wrapped inside Drupal.behaviors similar to webform_civicrm_forms.js.

`cj` is also replaced by just `jQuery` which seemed to work fine when I tested.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,9 +17,10 @@ jobs:
           - drupal: '9.2.*'
             civicrm: '5.39.*'
           - drupal: '9.2.*'
-            civicrm: '5.43.x-dev'
+            civicrm: '5.45.x-dev'
           - drupal: '9.2.*'
-            civicrm: 'dev-master'
+	    civicrm: '5.44.*'
+            #civicrm: 'dev-master'
     name: Drupal ${{ matrix.drupal }} | CiviCRM ${{ matrix.civicrm }}
     services:
       mysql:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
           - drupal: '9.2.*'
             civicrm: '5.45.x-dev'
           - drupal: '9.2.*'
-	    civicrm: '5.44.*'
+            civicrm: '5.44.*'
             #civicrm: 'dev-master'
     name: Drupal ${{ matrix.drupal }} | CiviCRM ${{ matrix.civicrm }}
     services:
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -144,4 +144,4 @@
     }
   }
 
-})(jQuery, Drupal, drupalSettings);
+})(cj, Drupal, drupalSettings);
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -1,11 +1,10 @@
 // Webform payment processing using CiviCRM's jQuery
-cj(function($) {
+(function ($, D, drupalSettings) {
   'use strict';
-  var
-    setting = drupalSettings.webform_civicrm,
-    $processorFields = $('[name$="civicrm_1_contribution_1_contribution_payment_processor_id"]');
+  var setting = drupalSettings.webform_civicrm;
 
   function getPaymentProcessor() {
+    var $processorFields = $('[name$="civicrm_1_contribution_1_contribution_payment_processor_id"]');
     if (!$processorFields.length) {
       return setting.paymentProcessor;
     }
@@ -43,16 +42,6 @@ cj(function($) {
       $('#billing-payment-block').html('');
     }
   }
-  $('fieldset.billing_name_address-group').remove();
-  $processorFields.on('change', function() {
-    setting.billingSubmission || (setting.billingSubmission = {});
-    $('#billing-payment-block').find('input:visible, select').each(function() {
-      var name = $(this).attr('name');
-      name && (setting.billingSubmission[name] = $(this).val());
-    });
-    loadBillingBlock();
-  });
-  loadBillingBlock();
 
   function getTotalAmount() {
     var totalAmount = 0.0;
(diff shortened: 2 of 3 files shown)


KarinG:
Thanks @puresyntax71 - what would be the use case/scenario for an administrator to want (or to need) AJAX enabled?

Please note that this PR causes existing `Drupal\Tests\webform_civicrm\FunctionalJavascript\StripeTest::testSubmitContribution` -> to fail.

puresyntax71:
Hi @KarinG,

Yeah, nothing in particular. Just noticed that it doesn't work with AJAX enabled.

Figured others might want AJAX for UI improvements in multi-step forms e.g. with contributions wherein the contribution is mostly set to another page.

Yeah, sorry, I haven't ran the test beforehand. I've marked it as "Draft" temporarily. I could probably investigate the tests if need be.

Could probably circle back in if this is a valid fix.

KarinG:
@puresyntax71 - no worries. Are you set up to run the Functional JS locally? We would be happy to show you / hop on a call with you - if you think that would be helpful. 

For this use case ->
"for UI improvements in multi-step forms e.g. with contributions wherein the contribution is mostly set to another page." 
I'd suggest start by writing a new test e.g. `StripeTestAjax.php` -> and then in the test select the AJAX checkbox and try to get it to process a Contribution - check that a transaction ID is stored etc. 

Once you have that working we can circle back to the `StripeTest.php` to see if the failure is something we can address in the existing test - or if the code edits really are preventing Stripe transactions when AJAX is -not- enabled. We would ping @jitendrapurohit in on that because he wrote the original Stripe test - it's good to touch base on any edits. 

Noting that both Dummy, iATS Legacy and iATS 1st tests appear to be ✅ with this PR. So it's probably close.

puresyntax71:
Hi @KarinG,

Yeah, was able to make it pass. Yeah, I'll try to add the `StripeTestAjax.php`. I think I might just copy the test from StripeTest.php itself.

What I currently have is add a test `StripeTest.php` and added `testAjaxSubmitContribution()` instead.

KarinG:
Awesome 🎉 Well done 🙂 

Re: tests -> what you could do is pass in `$ajax` into:  

```
public function testSubmitContribution() { 
in StripeTest.php

```
And then if `$ajax=1` -> navigate to the setting screen and switch on AJAX.
And `else` -> the original test will stay the same.

before
`$this->enableCivicrmOnWebform();`

puresyntax71:
Yeah, I wonder if this works. I had an issue with the test though wherein duplicate contributions are made. One is pending and the other is complete. Would be nice to get it fixed. This doesn't happen for manual tests though.

jitendrapurohit:
I've tested this manually with Membership, Participant and extra line item fields. Seems to be working absolutely fine. Thanks @puresyntax71. I think we can remove WIP from this PR?

KarinG:
What about the additional pending contribution? That would be really annoying for administrators.

The StripeTest (no AJAX) only generates one contribution so it must be related to AJAX.

KarinG:
@puresyntax71 - can you please take this out of Draft. I think it's close. I believe all that's left is to understand why (with AJAX) we get two Contribution records and have to specifically include -> completed in the test.

puresyntax71:
Hi @KarinG and @jitendrapurohit ,

> What about the additional pending contribution? That would be really annoying for administrators.

Yeah, it's only the test that seems to be producing pending contributions. Maybe my test is submitting the form twice. Manual test worked on my end though.

KarinG:
Ah ok -> I'll pull the test to my local and see if I can spot anything.

KarinG:
Ok looking at this now! 

1. I can run ✅ the iATS test `testSubmitContribution()` on my local -> 
image

2. But are getting an error message with the StripeTest `testSubmitContribution()` (before this PR).
`Behat\Mink\Exception\ElementNotFoundException : Form field with id|name|label|value "civicrm_1_lineitem_1_contribution_line_total" not found.`

I had never tried to run the Stripe test on my local yet. I wonder what I'm missing. I need to sort this out - and then can look into the test adjustments.

KarinG:
Ok -> indeed I get no line items for some reason:

image

Going to add in an additional wait()
`$this->getSession()->wait(5000);`

No joy.

Ah - ok got it. Adding a commit here that works to ensure the current test is more reliable.

KarinG:
image

KarinG:
If/when this passes the automated tests I'll be able to pull in the edited StripeTest to look at the second pending contribution issue.

KarinG:
Noting this -> dev-master only fail
image

KarinG:
Ok my findings:
1. non AJAX webform Stripe test -> 1 contribution (completed)
2. AJAX Stripe test -> 2 contributions -> 5 seconds apart - different invoice_id numbers -> one pending; one completed with trxn_id.

A
```
      'contribution_status_id' => '2',
      'receive_date' => '2021-12-22 03:01:17',
      'contribution_source' => 'CiviCRM Webform Test.5.41.0',
      'trxn_id' => '',
      'invoice_id' => '2996331eeefa7682affa5fbb99dae9d6',
      'id' => '1',

```
B
```
      'contribution_status_id' => '1',
      'receive_date' => '2021-12-22 03:01:22',
      'contribution_source' => 'CiviCRM Webform Test.5.41.0',
      'trxn_id' => 'ch_3K9AmAAwDzuDdbFC2y8pW5EA',
      'invoice_id' => '5afa72c2d0e6e07fe0faee1549d565c3',
      'id' => '2',

```

KarinG:
@mattwire -> Puresyntax71 has been working on 'if you enable AJAX in webform -> Stripe transaction fails'. That's fixed now 👍 but we're getting two contributions in the automated test -> 5 seconds apart (details see above). I'm going to merge this in because it's money in! And the pending transaction does not seem to surface on a live site? But I remain a bit concerned about generating extra contributions. If you have a chance to take a look I would appreciate it.

Also -> there is something with dev-master that I need to look at.

KarinG:
And we're green - it looks like there is an issue with CiviCRM master -> that somehow causes -> "You have Javascript disabled"

image

PR #659, opened 2021-12-15
Admin Form - Fix cross-theme compatability

colemanw:
Fixes the following to work on both Seven and Claro themes:

- Icons appear correctly
- Icons don't disappear on hover
- Icons change when changing contact type
- Form panel appears/disappears when checking "Enable CiviCRM Processing"
- Contact tab label updates as you type

Note: there doesn't appear to be a "standard" for marking up vertical tabs in Drupal - each theme gets to decide what markup to use. Which means it could be anything. That makes this script tricky to get right and probably there are themes out there which would appear broken. But at least it works on both of these now.

### DIFF
--- a/js/webform_civicrm_admin.js
+++ b/js/webform_civicrm_admin.js
@@ -180,8 +180,8 @@ var wfCiviAdmin = (function ($, D) {
   function changeContactLabel() {
     var c = $(this).attr('name').split('_')[0];
     var label = getContactLabel(c);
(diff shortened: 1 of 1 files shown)


KarinG:
Oh - thank you 🎉 I’m going to test that tonight.

KarinG:
Tested and works great on both seven and claro! And that's all we'll need well into Drupal 10 (and likely beyond). 

Thank you very much! 🥳

PR #660, opened 2021-12-15
Remove unused "clone contact" code

colemanw:
Overview
Removes unused code leftover from the D7 version.

Technical Details
This feature has not been implemented for D8/9 and if it is implemented it will basically need rewriting due to the differences between the old and new Webform modules.


KarinG:
Cool! In D9 you can copy paste and then search replace (_1 with _2) in the Source YAML for the new contact. That kind of works.

KarinG:
Tests ✅
Manual check ✅

petednz:
> Cool! In D9 you can copy paste and then search replace (_1 with _2) in the Source YAML for the new contact. That kind of works.

Good idea.

PR #661, opened 2021-12-15
Change package to "CiviCRM"

colemanw:
Changes the package so the module is grouped with other CiviCRM modules.

Historically, this module has been in the "CiviCRM" package: https://github.com/colemanw/webform_civicrm/blob/7.x-5.x/webform_civicrm.info#L4

I think this might have been an oversight during the porting process, as "Custom" is probably boilerplate code.

### DIFF
--- a/webform_civicrm.info.yml
+++ b/webform_civicrm.info.yml
@@ -2,7 +2,7 @@ name: 'Webform CiviCRM'
 type: module
 description: 'Webform CiviCRM Integration'
 core_version_requirement: ^8.8 || ^9
-package: 'Custom'
+package: 'CiviCRM'
 dependencies:
   - drupal:webform
   - civicrm


KarinG:
Maybe we should put it under Package: Webform :-)

Webform Mailchimp is under Webform

colemanw:
One or the other. Historical note, it's been under "CiviCRM" for the past 11 years :)

KarinG:
Ok points for that! 🙂

KarinG:
Funny how we've got 16 D9 sites/projects now and I never noticed it was under 'Custom'

colemanw:
I remember waffling between those two choices 11 years ago, and confusing @petednz by switching from "Webform" to "CiviCRM" when he was testing the 6.x-1.0.beta1 release! I made the switch because I noticed that `civicrm_views` module was under "CiviCRM", and that made sense, so I thought I'd be consistent.

PR #662, opened 2021-12-16
Fix updating records when editing a submission

colemanw:
Overview
Ensures that when editing an existing submission, the previously submitted contacts, etc will be updated.

Before

**Problem 1**
1. Make a "create new contact" style webform (without existing contact matching)
2. Submit the form. A new contact will be created.
3. Go back and edit the submission, then submit again.
4. Instead of updating the contact previously created, it creates a 2nd contact.

**Problem 2**
1. Add a multi-record custom group which contains a text field and a contact ref field
2. Expose both to a "create new contact" style webform, adding a 2nd contact to the form as the value of the contact reference field.
3. When saving the form, 2 records will be inserted, one for each field. Should have been a single record with both values.

**Problem 3**
1. Add a multi-record custom group with some fields
2. Make an "update current user" style webform (existing contact match to user)
3. Add one set of the multi-value fields, selecting "Create Only" for the create mode.
4. Submit the form. Note one custom record has been inserted.
5. Edit the form submission and save. Note a second record has been inserted instead of updating the previously created one.

After
All the above work as expected now.

### DIFF
--- a/tests/src/FunctionalJavascript/SubmissionEditTest.php
+++ b/tests/src/FunctionalJavascript/SubmissionEditTest.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Drupal\Tests\webform_civicrm\FunctionalJavascript;
+
+use Drupal\Core\Url;
+use Drupal\webform\Entity\WebformSubmission;
+
+/**
+ * Tests editing webform submissions.
+ *
+ * @group webform_civicrm
+ */
+final class SubmissionEditTest extends WebformCivicrmTestBase {
+
+  public function testEditSubmission() {
+    // Set up webform-civicrm with one contact, create-only
+    $this->drupalLogin($this->rootUser);
+    $this->drupalGet(Url::fromRoute('entity.webform.civicrm', [
+      'webform' => $this->webform->id(),
+    ]));
+    $this->enableCivicrmOnWebform();
+    $this->getSession()->getPage()->uncheckField("civicrm_1_contact_1_contact_existing");
+    $this->saveCiviCRMSettings();
+
+    $oldMax = $this->getMaxId();
+
+    // Submit form to create a contact
+    $this->drupalGet($this->webform->toUrl('canonical'));
+    $this->getSession()->getPage()->fillField('First Name', 'Dummy');
+    $this->getSession()->getPage()->fillField('Last Name', 'Tester');
+    $this->getSession()->getPage()->pressButton('Submit');
+
+    // Should have created one contact
+    $newMax = $this->getMaxId();
+    $this->assertEquals($oldMax + 1, $newMax);
+
+    $this->assertEquals('Dummy', civicrm_api3('Contact', 'get', ['id' => $newMax])['values'][$newMax]['first_name']);
+
+    // Get submission id
+    $database = \Drupal::database();
+    $query = $database->query("SELECT MAX(sid) AS maxsid FROM {webform_submission} WHERE webform_id = :wfid", [
+      ':wfid' => $this->webform->id(),
+    ]);
+    $sid = $query->fetchAll()[0]->maxsid;
+    $submission = WebformSubmission::load($sid);
+
+    // Edit submission and save
+    $this->drupalGet($submission->toUrl('edit-form'));
+    $this->getSession()->getPage()->fillField('First Name', 'Smarty');
+    $this->getSession()->getPage()->fillField('Last Name', 'Tester');
+    $this->getSession()->getPage()->pressButton('Save');
+
+    // Should have updated not created a new contact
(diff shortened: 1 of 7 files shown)


colemanw:
@KarinG I'm impressed at the existing test coverage! Tests caught several bugs in this PR, and after several rounds of fixes I got everything passing, and then added a unit test of my own (my very first :heart_eyes:). It tests the scenario described in problem 1, which I think is the most important. It would be nice to have tests for the other two but IMO shouldn't block merging this.

KarinG:
@colemanw - awesome on your first test 🎉 I love these functional tests - they are meaningful and easy to understand and fun to write. And have been able to identify regressions in CiviCRM (and recently in Webform). I’ll be reviewing this on Monday.

KarinG:
PS and yes I don’t have a must include tests for all scenarios policy so no worries.

KarinG:
Sorry @colemanw - this needs a rebase. 

PS - on testing - have you discovered the browser artifacts yet and the screenshots?

jitendrapurohit:
Thanks @colemanw. Saving/updating multi-record values in create/edit mode had some complexities and thanks for cleaning those up 👍 .  I'm glad the tests are passing with all these changes.

I've tested with 3-5 multi-valued custom fields (including contact refs which have bitten us in the past) and confirmed that the values are stored/updated as expected.

colemanw:
Thanks for the review @jitendrapurohit - I've rebased and tests are still passing so I think this is good to go :)

colemanw:
@KarinG no, what are the browser artifacts and screenshots?

KarinG:
Hang on there! I’m still going to review this tomorrow. And run my usual live site checks. And then I’ll show you what the artifacts and screenshots are as well.

KarinG:
Existing automated tests ✅
New test - local check ✅

Live site checks
Problem 1 (most important one) ✅
Screenshots

Edit submission -> add Middle Name -> Dog 
image

Still only one record in CiviCRM ✅ and also checked that contact record is updated to include Middle Name (Dog)
image

KarinG:
Browser artifacts: located at the bottom of this screen -> currently hundreds of HTML files + a handful of screenshots. 

image

HTML from your new test:
image

image

html is generated by the test on certain events e.g. when a button is hit - but you can always request one by injecting:

`$this->htmlOutput();`

e.g. https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/ContactSubmissionTest.php#L212

In addition there are screenshots PNG -> e.g. -> alaxvalues.png -> 

`$this->createScreenshot($this->htmlOutputDirectory . '/ajaxvalues.png');`

https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/CustomFieldSubmissionTest.php#L553

image

KarinG:
So that 👆 is a screenshot generated by Github Actions -> currently only the last pass (of the 4 Drupal x CiviCRM combinations in our testing matrix; which CiviCRM: see title of the webform -> 5.39.1 in this case) - it's super helpful to debug tests and see what's going on - especially if there is a mystery difference between your local and Github actions.

I also use screenshots on my local when writing tests. 

At some point I'll cull the html and screenshots a bit when we need to start speeding up the tests - but right now the tests are ok in terms of length and I think these visuals are an important part of testing.

KarinG:
Forgot to say THANK you for jumping in and getting this piece ported to D9. It is much appreciated. Hope you enjoyed it!

colemanw:
Thanks @KarinG for the thorough revew & explanation!

tests/src/FunctionalJavascript/SubmissionEditTest.php
```diff
@@ -0,0 +1,65 @@
...
+
+    $this->assertEquals('Smarty', civicrm_api3('Contact', 'get', ['id' => $newMax])['values'][$newMax]['first_name']);
+
+  }
+
+  protected function getMaxId($entity = 'Contact') {
+    return civicrm_api3($entity, 'get', [
+      'options' => ['sort' => "id DESC", 'limit' => 1]
+    ])['id'];
+  }
```
colemanw, line 64: FYI @KarinG in an earlier draft of this PR I was using APIv4 here. The test worked locally, but bombed on GitHub.
Ah well, someday we'll move this module up to v4.

Interesting historical note, APIv3 and this module are very nearly the same age; version 1.0 of Webform-CiviCRM was released just a month or so after APIv3 was introduced in CiviCRM core.

KarinG: @colemanw - the plan for that is to leave the tests as is (APIv3 - for now) - but to start with replacing all APIv3 calls with APIv4 calls in the code itself - first. I believe we already have 1 or 2 🙂

PR #663, opened 2021-12-16
Cleanup old hooks

colemanw:
Overview
Deletes one unused hook that no longer exists in D9, and fixes the path for `hook_help()` (the output is pretty lame, but at least it works now).


KarinG:
Nice cleanup 👍 - thank you @colemanw

PR #667, opened 2021-12-19
Add jquery.tokeninput.js to webform_civicrm

KarinG:
Overview
When using the autocomplete element - and when using a slightly older version of jquery.tokeninput.js (currently shipped with CiviCRM) releases the target area for the autocomplete element is quite small. We PR-ed an update to the jquery.tokeninput.js but it will be some time before sites will be using that release. 

Before
Using  jquery.tokeninput.js that ships with CiviCRM

After
Shipping  jquery.tokeninput.js with webform civicrm module

### DIFF
--- a/js/jquery.tokeninput.js
+++ b/js/jquery.tokeninput.js
@@ -1,0 +1,1106 @@
+/*
+ * jQuery Plugin: Tokenizing Autocomplete Text Entry
+ * Version 1.6.2
+ *
+ * Copyright (c) 2009 James Smith (http://loopj.com)
+ * Licensed jointly under the GPL and MIT licenses,
+ * choose which one suits your project best!
+ *
+ */
+;(function ($) {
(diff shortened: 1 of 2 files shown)


KarinG:
Alternatively -> I asked Dave and he said we can also do something like this:

You'd want https://github.com/civicrm/composer-downloads-plugin, i.e.:
"civicrm/composer-downloads-plugin": "^3.0"

and then in a section like this: https://github.com/civicrm/civicrm-core/blob/a597865cdda6c08e688ca40da3cf28d3e91b6831/composer.json#L129

you'd have the "* path" one and then your tokeninput one.

colemanw:
This is a perfectly fine low-tech solution IMO.

Honestly I think the TokenInput libary is on its way out anyway. The only reason I started using it with Webform-CiviCRM is "it was there" but in the intervening time CiviCRM has moved on to Select2 and has dropped TokenInput.

PR #669, opened 2021-12-21
don't disable state/province fields when existing contact exists

MegaphoneJon:
Overview
This is a regression of #561.  

If you have an autocomplete-select Existing Contact, and any fields are locked, **all** state/province fields on the form are locked.

I'm attaching a simple form that replicates the issue, but here are manual replication steps:
* Create a new Webform with 2 Civi contacts.  Both contacts should have first name, last name, and state/province enabled.  Contact 2 should also have "Existing Contact" enabled.
* Edit the Existing Contact *Form Widget* to **Autocomplete** and select **Name** under *Fields to Lock*.
* View the form.

Before
Your state/province fields will all be disabled.

After
Your state/province fields will be unlocked.

Comments
I don't know what the intent of these lines were - prior to #561, they were commented out. @jitendrapurohit do these lines serve a purpose?

My guess is that without this, you may not be able to lock state/province fields - but if that's the intent, that selector is way too broad, and was perhaps tested on a form with only a single contact.  If this code *is* necessary, I think it'll be necessary to further filter the DOM elements by CiviCRM contact, which should be possible because that value is part of the `n` array.

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -183,10 +183,6 @@ var wfCivi = (function ($, D, drupalSettings) {
           var fn = (op === 'hide' && (!showEmpty || !isFormItemBlank($el))) ? 'hide' : 'show';
           $(':input', $el).prop('disabled', fn === 'hide');
           $(':input', $el).prop('readonly', fn === 'hide');
-          $('select.civicrm-enabled[name*="_address_state_province_id"]').each(function() {
-            $(this).prop('disabled', fn === 'hide');
-            $(this).prop('readonly', fn === 'hide');
-          });
           if (hideOrDisable === 'hide') {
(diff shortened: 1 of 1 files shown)


KarinG:
Hi Jon -> that failure on dev-master is across the board. I'm looking into it. We're testing fine against the RC.

KarinG:
Ok the test failure for dev-master is re: https://github.com/civicrm/civicrm-core/pull/22282 - I've alerted Coleman. 

@jitendrapurohit - do you have any concerns about this PR? Jon's PR does not break functionality we're testing re: existing contacts and locked fields.

KarinG:
Hi - I'd like to tag a new WFC release today/tomorrow and agree w/Jon that the selector needs more work.

KarinG:
PS - The regression in CiviCRM Core is fixed.

PR #676, opened 2021-12-26
D8/9 - Fix recur payment via ipn processor

jitendrapurohit:
Overview
Fix recur payment via ipn processor

Before
Recur payment does not work with ipn processor eg `PayPal - Website Payments Standard`. To replicate -

- Create processor of type `PayPal - Website Payments Standard`.
- Create a webform and enable this processor + set Frequency of Installments to `year`
- Load the webform and submit. The submission takes you to Paypal site for payment.
- Check the civi records, the contribution created is not attached to any recur series.

After
Works. On submit, a recur contribution is created in civi.

Technical Details
Uses the same code written for deferred payment to create a recur contribution in civi and then sets the same to $params in `submitIPNPayment()` where the payment processor call to `doPayment()` is done.

Comments
This is also a bug in d7.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1983,7 +1983,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     ];
 
     $APIAction = 'transact';
-    if ($paymentType === 'deferred') {
+    if (in_array($paymentType, ['deferred', 'ipn'], TRUE)) {
       $APIAction = 'create';
     }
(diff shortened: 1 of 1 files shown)


mattwire:
Just to note I'm currently getting "Things don't appear to be working at the moment. Please try again later." from paypal when applying this PR

mattwire:
If `frequency interval` is not on the form we also need the following core PR: https://github.com/civicrm/civicrm-core/pull/24614

mattwire:
Have now tested this and confirmed it's ok to merge. See note about core PR

KarinG:
Thank you Matt! I've asked Dave to look at your CiviCRM Core PR.

PR #677, opened 2021-12-26
D8/9 - Fix case update when client != c1

jitendrapurohit:
Overview
Fix case update when client != c1.

Before
Case updated on c1 when client is set to c2. To replicate -

- Case tab - Select 1 in the Number of case with Case client = Contact 2.
-  Ensure `Update Existing Case` is set to Ongoing.
- Set the widget of Contact 2 to autocomplete.
- Load the webform and submit. Confirm a case is created on contact 2.
- Load the webform again and submit with same set of contacts.

Expected Result: Case on contact 2 should be updated.

Actual Result: New Case is created on contact 2. 

After
Fixed. Existing case on the contact is updated.

Technical Details
The previous variable - `$this->ent['contact'][$n]['id']` has $n iterating as 1, 2, 3... which considers c1, c2 to check for existing case. This should instead be checking for client id set on the webform setting which correctly arrives on `$data['case'][1]['client_id']`.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1358,8 +1358,8 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       if (is_array($data) && !empty($data['case'][1]['client_id'])) {
         $params = $data['case'][1];
         //Check if webform is set to update the existing case on contact.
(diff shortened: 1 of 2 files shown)


MegaphoneJon:
I felt bad about rejecting #647 so I tried testing this PR as well.

Unfortunately, there's a hard-coded assumption that [the manager is contact B in a relationship](https://github.com/colemanw/webform_civicrm/blob/55995a1856aed703531bd47731fef1e10f252d4f/src/AdminForm.php#L756).

This is out of scope for this PR, but is a blocker to my testing.

PR #680, opened 2022-01-12
fix formatting on 'no results' result for existing contact widget

MegaphoneJon:
Overview
On an "existing contact" widget. the "no results" option shows raw HTML where it previously rendered the HTML.

This site has the upgraded jquery.inputtoken (https://github.com/civicrm/civicrm-packages/pull/336) but I believe the bug exists regardless.

Before

Selection_1354

After
Selection_1353

Technical Details
Passing `enableHTML: TRUE` is a simpler alternative but I feel like this offers stronger protection against accidentally introducing an XSS vuln in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Webform CiviCRM is a powerful, flexible, user-friendly form builder for CiviCRM!
 Installation & Getting Started
 ------------------------------
 
-- Download and enable this module, plus CiviCRM, Webform.
+- Download and enable this module, plus CiviCRM, Webform!
 - Create a new webform (or go to edit an existing one).
 - Click on the Settings -> CiviCRM tab.
 - Enable the fields you like, and optionally choose introduction text and other settings.
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Documentation
 -------------
 
 Further instructions can be found at:
-https://docs.civicrm.org/webform-civicrm/en/latest/
+https://docs.civicrm.org/webform-civicrm/en/latest
 
 Issues
 ------
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -18,6 +18,7 @@
             toHide, {
             hintText: field.data('search-prompt'),
             noResultsText: field.data('none-prompt'),
+            resultsFormatter: formatChoices,
             searchingText: "Searching..."
           });
         }
@@ -37,6 +38,16 @@
         });
       });
 
+      /**
+       * Format the choices in the "Existing Contact widget", with a special format for the "No Results" item.
+       */
+      function formatChoices(item){
+        var string = item[this.propertyToSearch];
+        if (string == this.noResultsText) {
+          return "<li><em><i>" + string + "</i></em></li>";
+        }
+        return "<li>" + string + "</li>";
+      }
 
       function changeDefault() {
         var val = $(this).val().replace(/_/g, '-');
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -136,7 +136,7 @@ class ContactComponent implements ContactComponentInterface {
       }
       if (count($ret) < $limit && $element['#allow_create']) {
         // HTML hack to get prompt to show up different than search results
-        $ret[] = ['id' => "-$str", 'name' => '<em><i>' . Xss::filter($element['#none_prompt']) . '</i></em>'];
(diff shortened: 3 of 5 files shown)


colemanw:
Fix makes sense to me

jitendrapurohit:
When i apply this PR, the create new option is not displayed which i think might be the reason for the test failures. 

Console output -

image

Seems it affects only those setup which has older civi installed (i.e, without updated tokeninput PR). I think this is because wc still connects the civi tokeninput file via webform_civicrm_library_info_alter hook - https://github.com/colemanw/webform_civicrm/blob/6.x/webform_civicrm.module#L38.

Can you pls try removing this hook entirely since the latest tokeninput was added in the js directory of the module itself. See https://github.com/colemanw/webform_civicrm/pull/667/files#diff-39103a9714d984159d152853b569f1210884045767aff880a5e8818cafdc22de. I think we should remove these dependency calls so those with older civi versions are not affected.

KarinG:
🤔 I am able to reproduce this -> on [5.47.alpha1] - I upgraded my site to -dev last week
image

KarinG:
And this PR fixes it -> 
image

KarinG:
Going to push to kick off the tests again.

KarinG:
It's really only 1 error (after that we crash and the test is complaining about the database not torn down properly). 

1) Drupal\Tests\webform_civicrm\FunctionalJavascript\CaseSubmissionTest::testCaseSubmission
WebDriver\Exception\ElementNotVisible: element not interactable

Line 63 in the CaseSubmissionTest:
`$this->fillContactAutocomplete('token-input-edit-civicrm-1-contact-1-contact-existing', $this->_caseContact['first_name']);`

So that happens here ->
```
  protected function fillContactAutocomplete($id, $value) {
    $page = $this->getSession()->getPage();
    $driver = $this->getSession()->getDriver()->getWebDriverSession();
    $elementXpath = $page->findField($id)->getXpath();

    $this->assertSession()->elementExists('css', "#" . $id)->click();
    $driver->element('xpath', $elementXpath)->postValue(['value' => [$value]]);

    $this->assertSession()->waitForElementVisible('xpath', '//li[contains(@class, "token-input-dropdown")][1]');
    $this->createScreenshot($this->htmlOutputDirectory . '/autocomplete.png');

    $page->find('xpath', '//li[contains(@class, "token-input-dropdown")][1]')->click();
  }
```

This is the line that it chokes on:
`$page->find('xpath', '//li[contains(@class, "token-input-dropdown")][1]')->click();
`

KarinG:
@jitendrapurohit -> do you know what we need to do to adjust the test so that it can click on that element again?

Going to try one thing ->

KarinG:
So that test passes on my local but not on (older) 5.35.* (in github matrix).

MegaphoneJon:
I said earlier that I think this bug existed regardless of https://github.com/civicrm/civicrm-packages/pull/336 but my later testing showed that it was that PR which introduced the bug.  So that would explain it not working on 5.35.

KarinG:
Fails on 5.45.* as well (that's current ESR) -> 5.46.* and up are ok.

image

KarinG:
What if we just strip the <em> and <i> tags? Then it should work for all - (as in not show HTML tags) - later on we can re-introduce the tags?

KarinG:
Ok so we can't merge yet - it also fails on 5.45 and the ESR people will be upset.

KarinG:
Ok I just pushed a trivial commit to get the tests to re-run. I suspect 5.45 will fail (again) - but let's confirm.

KarinG:
The error in 5.45.* goes back to this line -> that just doesn't work in 5.45.* with this PR [but it works great without this PR].

`
$this->fillContactAutocomplete('token-input-edit-civicrm-1-contact-1-contact-existing', $this->_caseContact['first_name']);
`

I'm not sure how to solve this?

pradpnayak:
Will this be backported on Drupal 7?

KarinG:
If we can get it to pass 5.45.* then yes we can backport it.

KarinG:
Ok - closing in favour of #729 -> essentially: no longer depending on what civicrm core's version of jquery.tokeninput.js

jitendrapurohit:
@pradpnayak The D7 equivalent is raised at https://github.com/colemanw/webform_civicrm/pull/730. Can you pls confirm if it works for you?

PR #683, opened 2022-01-19
allow correct loading of 'Show Public Events'

MegaphoneJon:
Overview
If you go to the Events tab of WFC, and select **Private** for *Show Public Events*, then reload the page, your setting isn't respected.

Before
Selecting "Private" doesn't work.

After
Selecting "Private" works.

Technical Details
Unlike the other settings on this page, this setting can have a value of `0` and needs `wf_crm_aval`'s strict mode to avoid a truthiness bug.


jitendrapurohit:
Replicated the bug and confirmed that the fix here resolves it correctly. Thanks @MegaphoneJon

PR #684, opened 2022-01-19
Add option for hiding full events

MegaphoneJon:
Overview
Adds the ability to hide full events.

Before
We have functionality to prevent registering for past or full events, for showing warnings for past or full events, but can't hide full events, just past events.

After
Parity for past and full events.

Comments
I kept the default of showing full events a) for backwards compatibility, b) checking for full events adds an API call per event.


jitendrapurohit:
This looks good. I've tested and it works as expected. I'll add a test in another PR. merging.

jitendrapurohit:
Test added https://github.com/colemanw/webform_civicrm/pull/705

PR #685, opened 2022-01-22
D8/9 - Allow loc types to be submitted as non primary from the webform

jitendrapurohit:
Overview
Allow loc types to be submitted as non primary from the webform

Before
The first Address or other location types submitted va webform are forced to is_primary = 1. This doesn't allow admins to capture additional address values without overriding the primary record in civi contact.

After
Adds a setting on the form (default to 1). If selected as No - the location type will not be forced to be a primary record.

image

Comments
@KarinG Tagging as wip to make sure it passes the tests and will also add a new test to check the same.

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -372,6 +372,10 @@ class AdminForm implements AdminFormInterface {
               (isset($field['contact_type']) && $field['contact_type'] != $c['contact'][1]['contact_type'])) {
               continue;
             }
+            // Make sure Primary field is only displayed for the 1st set.
+            if (strpos($fid, 'is_primary') !== false && $i > 1) {
+              continue;
+            }
             $fid = 'civicrm_' . $n . '_contact_' . $i . '_' . $fid;
             $fieldset[$fid] = $this->addItem($fid, $field);
           }
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -354,6 +354,12 @@ class Fields implements FieldsInterface {
             'expose_list' => TRUE,
             'value' => '1',
           ];
+          $fields[$key . '_is_primary'] = [
+            'name' => 'Is Primary',
+            'type' => 'select',
+            'expose_list' => TRUE,
+            'value' => '1',
+          ];
         }
       }
       $fields['website_url'] = [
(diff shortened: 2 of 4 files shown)


MegaphoneJon:
I dealt with this once [with a hack](https://gist.githubusercontent.com/MegaphoneJon/c49857d9f68cf6292c3820be8a914213/raw/2606e87b7c3a809399f5e50e72985de6177008c7/gistfile1.txt). 

This PR is better, but I don't think we should be defaulting the first address to primary.  
* On a new contact, or a contact with no addresses, the first address will be set primary anyway.  
* If the contact already has addresses, we shouldn't override the existing primary address unless requested.

Additionally, "Is Primary" set to "No" doesn't mean the address won't be primary, which can be confusing.

I propose:
* The field be relabeled "Always Set as Primary" 
* It defaults to `FALSE`
* We remove the 2 lines I removed in my diff.  
* We invert the behavior of this new field.
 
That should remove any UX confusion.

KarinG:
Thank you for reviewing @MegaphoneJon

@jitendrapurohit - if you have time to work on this next weekend that would be great. I will then see this through and get this merged.

jitendrapurohit:
Hmm, just worried about the admins that expect the override to happen? i.e, 

>If the contact already has addresses, we shouldn't override the existing primary address unless requested.

This was the default behavior in d7 and d8. If we change this, users already expecting the above in the existing forms would need to manually update the setting `always set as primary` to 1. That sounds like unnecessary overhead. Not sure if we need to add an upgrade script to update the setting to 1 for existing webforms.

KarinG:
The primary goal is to give the admin an option to ensure that not every address (or email or phone) ever submitted via a webform ends up as is_primary. 

I'll live test this today and will add some final tweaks to the wording.

KarinG:
I've tested this locally - merging and then live testing on a project that has webforms for donations (with complete addresses that we want is_primary) and webforms for appeals (with partial addresses - postal code only - that we do not want to override is_primary).

KarinG:
Circling back reporting on live testing -> Other

image

Just the label is a bit confusing (Address 1 (primary)) - will think about how to address that -> 

image

KarinG:
Ok decided to change it to ->

**Address 1 (often primary)**

PR #686, opened 2022-01-26
D8/9 - Fix default load of custom value when it is set to null in civicrm

jitendrapurohit:
Overview
Fix default load of custom value when it is set to null in CiviCRM

Before
To replicate -

- Create a custom set with text, radio and checkbox custom fields.
- Create a contact in CiviCRM with a value in only the text custom field. Radio & checkbox value is empty/null.
- Enable all three fields on a webform.
- Set a default value on the radio and checkbox custom field.
- Load the webform with cid1=contact id from step 2.
- The value of radio and checkbox custom field is not loaded.

Expected Result - If there is no value set in CiviCRM, the webform defaults should be loaded on the element.

After
Fixed.


jitendrapurohit:
This PR is in production from last 1 year. Also have a test. Can we merge this @KarinG

KarinG:
Absolutely - thanks for flagging it 👍

PR #688, opened 2022-01-26
D8/9 - Fatal error on webform submission if it has a radio button element wi…

jitendrapurohit:
…th single option

Overview
Fatal error on webform submission if it has a radio button element with single static option

Before
To replicate

- Create a custom field of type radio and 1 option.
- Enable it on a webform.
- Edit the radio element from the build page and set the option to static.

image

- Load the webform. The radio element is rendered as a checkbox (Correct).
- Submit the webform without selecting the checkbox value. Fatal error is returned on the next page.

```
Drupal\Core\Entity\EntityStorageException: '' is not a valid option for field custom_3 
in Drupal\Core\Entity\Sql\SqlContentEntityStorage-
>save() (line 811 of /Users/jitendra/www/d9unit/web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).
```

From log

```
CiviCRM_API3_Exception: '' is not a valid option for field custom_3 
in civicrm_api3() (line 134 of /Users/jitendra/www/d9unit/vendor/civicrm/civicrm-core/api/api.php).
```

After
Fixed.

### DIFF
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -265,11 +265,7 @@ class CivicrmOptions extends OptionsBase {
    * {@inheritdoc}
    */
   public function hasMultipleValues(array $element) {
-    if (!empty($element['#extra']['multiple']) ||
-      (empty($element['#civicrm_live_options']) && !empty($element['#options']) && count($element['#options']) === 1)) {
-      return TRUE;
-    }
-    return FALSE;
+    return \Drupal::service('webform_civicrm.utils')->hasMultipleValues($element);
   }
 
   /**
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -1008,4 +1008,17 @@ class Utils implements UtilsInterface {
     return $params;
   }
 
+  /**
+   * Does an element support multiple values
+   *
+   * @param array $element
+   */
+  public function hasMultipleValues($element) {
+    if (!empty($element['#extra']['multiple']) ||
+      (empty($element['#civicrm_live_options']) && !empty($element['#options']) && count($element['#options']) === 1)) {
+      return TRUE;
+    }
+    return FALSE;
+  }
+
 }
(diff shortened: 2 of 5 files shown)


stesi561:
In addition (or perhaps further to?) the test failures above - hitting problems with this when submitting a webform that has multiple memberships when not all are required to be filled.

jitendrapurohit:
@stesi561 The latest update should fix the problem with multiple memberships. Can you confirm pls? It also passes the test now 🎉

stesi561:
I've just tested on the site we're having problems with. 

There were two webforms experiencing issues. Webform A was the one which this PR is supposed to fix. Webform B which earlier versions broke.

Testing: With the modified version of this PR in place - that reverted to the previous behaviour on Webform B. 
Webform A submits, Webform B submits.

Replacing the customised patch with this version.
Webform A failed to submit but Webform B submits.


Webform A error was the original:

`Drupal\Core\Entity\EntityStorageException: '' is not a valid option for field custom_734 in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 810 of /srv/www/production/d9_nzoa.org.nz/web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).
`

KarinG:
@stesi561 - can you please test this on a new form? Alternatively - removing and adding fields involved can work too (but can have negative side effects if you use the Results).  A third option is to add `custom_734` to a new form and then check the **YAML** -> manually update your existing form.

KarinG:
@stesi561 - did you have a change to look at testing this on a new form or - re-adding `custom_734`?

stesi561:
Was trying to test this against latest but was failing to apply via composer so have rerolled at -> https://github.com/colemanw/webform_civicrm/pull/734

jitendrapurohit:
#734 misses the comments for the fix in src/WebformCivicrmPostProcess.php so I've rebased this PR and closed that.

jitendrapurohit:
@stesi561 has tested this at https://github.com/colemanw/webform_civicrm/pull/734#issuecomment-1112820151

Merging.

PR #690, opened 2022-01-28
D8/9 - Added check to see if $elements['managed_file'] exists

kewljuice:
Fixes: Error: Call to a member function isDisabled() on null in Drupal\webform_civicrm\Fields->wf_crm_get_fields() (line 198 of /var/www/html/web/modules/contrib/webform_civicrm/src/Fields.php)

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -195,7 +195,7 @@ class Fields implements FieldsInterface {
         'type' => 'select',
         'value' => $this->utils->wf_crm_get_civi_setting('lcMessages', 'en_US'),
       ];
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you for the patch. 

Test failure is unrelated to this PR. I’m trying to figure out why one of our payment processor tests against CiviCRM dev-master started failing 6-9 days ago (it consistently runs fine against other CiviCRM versions).

KarinG:
We figured out the issue in CiviCRM Core (a recent change to billing params) - and have stripped dev-master from our WFC testing matrix until that is resolved - so this will test ✅ on merge.

PR #524, opened 2021-04-03
D8/9 - Add test for Stripe payment + tidy up iATS test

jitendrapurohit:
Overview
----------------------------------------
Add test for Stripe payment

Before
----------------------------------------
No test.

After
----------------------------------------
Test added.

Comments
----------------------------------------
@KarinG @mattwire

PR #692, opened 2022-01-28
D7 - Fix PHP notices when there are no billing address fields

mattwire:
Overview
Stripe does not use billing address fields but webform_civicrm expects them to be there so you get PHP notices if they are not.

Before
php notices for address fields

After
No php notices for address fields

Technical Details

Comments


jitendrapurohit:
D8/9 was handled via https://github.com/colemanw/webform_civicrm/pull/524. Seems of a port of that PR.

KarinG:
Confirming port  ✅

PR #695, opened 2022-02-02
Fix output in github actions so it shows you the actual errors in drupal error textboxes at the top of pages

demeritcowboy:
Overview
Wrong function used to get the contents of the drupal error textboxes that appear at top of pages.

Before
Just a limited stack trace that points to the generic error function, so you don't know what the actual error is without hunting down the screenshot.

After
Gives you the text contents of the error boxes in the github action output directly.

Technical Details
getValue() only works on form elements. The textbox is just a div.

Comments
Ahhhhhhhhh

See https://github.com/colemanw/webform_civicrm/actions/runs/1786370647 for an example where I fudged the code to make such an error.


demeritcowboy:
@KarinG @mattwire

mattwire:
Oh that's awesome @demeritcowboy thanks so much for doing this :-)

KarinG:
Awesome Dave! 👍

PR #697, opened 2022-02-04
Using API params only

KarinG:
Overview
We ran into an issue where one of the iATS tests had started failing due to a change in propertybag in CiviCRM Core
[Add billingStateProvince and standardized property names by mattwire · Pull Request #21583 · civicrm/civicrm-core](https://github.com/civicrm/civicrm-core/pull/21583)

Before
Contains propertyBag references

After
propertyBag references are removed - using API params only.

Technical Details
**Note:** All payment processor tests including -> Stripe FunctionalJavascript test pass ✅
I've only commented out //// the propertyBag bits for now to clearly show what can be removed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,8 +12,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - drupal: '9.1.*'
+          - drupal: '9.3.*'
             civicrm: '5.35.*'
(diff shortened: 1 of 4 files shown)


KarinG:
@mattwire - thank you for the call this morning! I've cleaned this up and have also adjusted the matrix so that we're now running D9.3.4 against 4 different CiviCRM versions. Will merge on pass.

PR #696, opened 2022-02-03
Adding test for ACHEFT - status should be Pending

KarinG:
Overview
----------------------------------------
Adding to our FunctionalJavascript testing

Before
----------------------------------------
No test for ACHEFT + Contribution status was incorrectly set to Completed

After
----------------------------------------
This PR adds a test for iATS Payments ACHEFT + Contribution status now correctly set to Pending

Technical Details
----------------------------------------
iATS Legacy Payment Processor

Comments
----------------------------------------

The test initially failed. But all good now.

Native Contribution pages -> also correctly put status = Pending (Incomplete) for ACHEFT.

Initial result before this PR:

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionIatsTest::testSubmitACHEFTContribution
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Pending'
+'Completed'
```

PR #698, opened 2022-02-05
Remove special handling for transact API and return the expected result

mattwire:
Overview
@KarinG This will fix one of the PHP notices you're seeing on #696

Before
transact not returning results in expected format. Special handling for that.

After
transact returning results in expected format. Remove special handling.

Technical Details
Note that this was an error when I wrote the contributiontransactlegacy extension where this code came from. The original core transact API returned an API3 Contribution.get result per this PR.

Comments

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -744,13 +744,9 @@ class Utils implements UtilsInterface {
       return ['error_message' => $e->getMessage()];
     }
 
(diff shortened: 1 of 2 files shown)


mattwire:
All tests passed :-) Merging

KarinG:
✅ Green is SO nice! I'll rebase my ACHEFT PR - after I work these two teens out the door (early swim). At least they can drive now!

PR #699, opened 2022-02-05
WIP - PHP 7.4

KarinG:
Overview

**5.35.***
```
There were 19 failures:
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContactSubmissionTest::testAutocompleteWithGroupFilter
Error message Deprecated function: Function money_format() is deprecated in
...
```

**5.45.*** + **5.46.*** + **dev-master**
```
There was 1 failure:
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\MultiCustomFieldsSubmissionTest::testAnonymousSubmitWithContribution
Error message Notice: Trying to access array offset on value of type null in Drupal\webform_civicrm\AdminForm->postProcess() (line 1896 of /home/runner/work/webform_civicrm/webform_civicrm/src/AdminForm.php).
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: shivammathur/setup-php@v2
         with:
-            php-version: 7.3
+            php-version: 7.4
             extensions: dom, curl, libxml, mbstring, zip, pdo, mysql, pdo_mysql, bcmath, soap, intl, gd, exif, iconv
             coverage: none
             tools: composer:v2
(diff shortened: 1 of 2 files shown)


KarinG:
It's not "type", it's that `$field` is null. It's not finding the field for `$key`.

```
    $i = 0;
    $created = [];
    foreach ($this->settings as $key => $val) {
      if (strpos($key, 'civicrm') === 0) {
        ++$i;
        $field = $this->utils->wf_crm_get_field($key);
        if (!isset($enabled[$key])) {
          $val = (array) $val;
          if (in_array('create_civicrm_webform_element', $val, TRUE) || (isset($val[0]) && $field['type'] == 'hidden')) {
```

KarinG:
Great 🎉 so we can switch to testing w/ PHP 7.4 when supporting 5.45.* and up.

image

PR #703, opened 2022-02-08
fix saving multi-record custom field create mode

MegaphoneJon:
Overview
Multi-record custom fields have a "create mode" which determines whether the webform submission can overwrite existing data or must always create a new record.

#### Steps to replicate
* Create a multi-record custom field set in CiviCRM.
* In Webform, enable that field set on a contact.
* Set the *Create Mode* to **Create Only**.
* Press **Save**.
Selection_1391

Before
When the form reloads, the **Create Only** setting is lost.

After
When the form reloads, the **Create Only** setting is respected.

Technical Details
On save, there's a giant `if..elseif..elseif...` etc. block that handles special cases.  However, the first `if` statement will always evaluate to `TRUE`, so the `elseif` that handles saving the `create_mode` setting is never evaluated.

I moved the `createmode` value into its own `if` statement so that it's certain to be evaluated.

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1946,10 +1946,6 @@ class AdminForm implements AdminFormInterface {
           $component['value'] = $val;
           $enabled[$key] = $component;
         }
-        elseif (substr($key, -11) === '_createmode') {
-          // Update webform's settings with 'Create mode' value for custom group.
-          $this->settings['data']['config']['create_mode'][$key] = $val;
-        }
         else {
           // Try to "update" options for existing fields via ::insertComponent
           // Always insert fieldsets, as there are checks to see if the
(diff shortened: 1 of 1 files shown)


colemanw:
Looks like the right fix. I'm not sure what's going on with the tests.

KarinG:
Moving those lines -> breaks this:

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\MultiCustomFieldsSubmissionTest::testSubmitWebform
Behat\Mink\Exception\ExpectationException: The field "civicrm_1_contact_1_cg1_custom_1" value is "", but "Feb" expected.
```

jitendrapurohit:
Looks like a similar change was merged along with https://github.com/colemanw/webform_civicrm/pull/725. I can no longer replicate this problem. Do you agree @MegaphoneJon ? If yes, pls close. Thanks.

jitendrapurohit:
Closing as the fix is already merged in webform civicrm as per the above comment. Pls re-open if this is still required.

Thanks @MegaphoneJon

PR #704, opened 2022-02-12
Add Test for Custom Dedupe Rule: 

KarinG:
Overview
After the regression we recently had re: dedupe rules -> time to add some coverage

Before
This test was just doing unsupervised rule (and that was working) - now it's set up to handle a custom rule.

Comments
We really do want a number of rules - so once this is working - we need to refactor it so that we can run the same test using 3 different rules: unsupervised, FirstPhone, FirstDOB


KarinG:
This is running against 5.46.2 so we're expecting this to ✅

image

KarinG:
😂 Test fails on 5.35.* b/c API to create the custom deduperule did not exist yet

KarinG:
Also confirming that this test secures what failed in 5.46.1 ->

image

PR #706, opened 2022-02-13
D8/9 - Add test for stripe recur payment

jitendrapurohit:
Overview
Add test for stripe recur payment

Before
No test.

After
Test Added. But it fails and seems to be a bug (not sure in wc or in stripe). The error is:

image

To replicate

- Create a webform with recur enabled (Frequency = month).
- Submit the webform.
- Error.

@mattwire Any quick thoughts on why is it failing?

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1966,7 +1966,9 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       foreach ($this->line_items as $key => $k) {
         $this->line_items[$key]['unit_price'] = round(($k['unit_price'] / $numInstallments), 2, PHP_ROUND_HALF_UP);
         $this->line_items[$key]['line_total'] = round(($k['line_total'] / $numInstallments), 2, PHP_ROUND_HALF_UP);
-        $this->line_items[$key]['tax_amount'] = round(($k['tax_amount'] / $numInstallments), 2, PHP_ROUND_HALF_UP);
+        if (isset($k['tax_amount'])) {
+          $this->line_items[$key]['tax_amount'] = round(($k['tax_amount'] / $numInstallments), 2, PHP_ROUND_HALF_UP);
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


mattwire:
@jitendrapurohit Can you check the configuration of the webform per https://docs.civicrm.org/stripe/en/latest/integration/#recurring-payments

jitendrapurohit:
The test here does not enable `Number of installments` on the webform. Not having this field should also be able to start monthly recur payments on a contact? @mattwire

mattwire:
@jitendrapurohit You *must* have one of  "Number of installments" or "Interval of installments" on the form. This is because we must be able to detect if the webform is submitting a recurring payment using javascript: https://lab.civicrm.org/extensions/mjwshared/-/blob/master/js/crm.payment.js#L175-181

It would be really nice to find a better way to detect this but currently this is the only way I could find.

KarinG:
Use `number of installments = 0`

Anything other than `number of installments = 1` will ensure WFC creates a recurring series. `number of installments = 0` means open ended recurring.

jitendrapurohit:
@KarinG With installment  = 0 or anything else (not 1) returns the same error. I think this is what Matt means when he says you "must" have installment field while taking recur payments through the webforms. Am i correct @mattwire. If yes, do you intend to support unending stripe recur payments via webform in the future?

KarinG:
@mattwire - I just added to this test -> https://github.com/colemanw/webform_civicrm/pull/717 - it now demonstrates WFC with iATS Payments for both instalments = 12 as well as instalments = 0 (which creates an open ended recurring series). 

I think this is then waiting on you to make edits in Stripe? If/when you do - we have the test ready to go - to lock it in.

mattwire:
Stripe does support open ended recurring payments. But it needs to be able to detect this directly on the form using jquery/javascript. On drupal 7 that means you cannot use a hidden field (for installments) in "secure value" mode. I don't have a drupal 9 site to test with right now. If you run `CRM.payment.getIsRecur()` in the browser console it will tell you if it's detected or not.

KarinG:
Installments is not hidden here - so not sure why Stripe is not detecting that we’re trying to create a recurring series?

mattwire:
test

PR #707, opened 2022-02-13
Prepare for deprecated payment_status_id

mattwire:
Overview
See https://lab.civicrm.org/dev/financial/-/issues/141

We should check value of `payment_status` on return from payment but it may not be defined.

Before
Checking deprecated `payment_status_id`

After
Checking `payment_status` if available.

Technical Details

Comments

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -719,11 +719,19 @@ class Utils implements UtilsInterface {
       // We allow this to be overridden here but default to FALSE.
       $params['is_email_receipt'] = $params['is_email_receipt'] ?? FALSE;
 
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you - as of which CiviCRM version is `payment_status_id` deprecated? If you could add that in the comments in the code here so that we know when we can safely remove this - that would be great.

mattwire:
> Thank you - as of which CiviCRM version is `payment_status_id` deprecated? If you could add that in the comments in the code here so that we know when we can safely remove this - that would be great.

Honestly I think it'll be a long time away - all processors will need to return both before we can do that and we don't have a clear way of getting paymentprocessor authors to update their code nor anyone reviewing payment stuff on core. But making this change means webform will be ready for that change when it happens.

PR #709, opened 2022-02-18
Handle phone_type_id if overwritten to NULL in civicrm_phone table

stesi561:
https://www.drupal.org/project/webform_civicrm/issues/3264156

Resolve e-notice for missing array element.

It seems from comments that this part of code shouldn't apply in
situation causing error however I'm unclear what it's actually doing
so this needs review/testing before merging.

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -380,10 +380,13 @@ abstract class WebformCivicrmBase {
       $valueFound = false;
       foreach($values as $key => $value){
         if ((in_array($ent, ['address', 'email']) && $value['location_type_id'] == $setting['location_type_id'])
-          || (
-             $value['location_type_id'] == $setting['location_type_id'] &&
-             (!isset($setting[$ent.'_type_id']) || $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])
-             )
+            || (
+              $value['location_type_id'] == $setting['location_type_id'] &&
+              (
+                !isset($setting[$ent.'_type_id']) ||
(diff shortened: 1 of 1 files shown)


KarinG:
@jitendrapurohit - if you could look at this (including why we're not hitting that notice in a test yet) that would be great.

jitendrapurohit:
Hmm, I wasn't able to replicate this on a normal workflow of saving phone numbers.

From the changes, it seems `phone_type_id` will only be missing if its set to NULL in civicrm_phone table. So I was able to replicate it by manually removing the value for `phone_type_id` column in the database.

Since this column is allowed to store NULL by CiviCRM, I think it's ok to avoid this with an if condition as done here.  @KarinG @stesi561

src/WebformCivicrmBase.php
```diff
@@ -380,10 +380,13 @@ protected function matchLocationTypes($c, $ent, &$values){
       $valueFound = false;
       foreach($values as $key => $value){
         if ((in_array($ent, ['address', 'email']) && $value['location_type_id'] == $setting['location_type_id'])
-          || (
-             $value['location_type_id'] == $setting['location_type_id'] &&
-             (!isset($setting[$ent.'_type_id']) || $value[$ent.'_type_id'] == $setting[$ent.'_type_id'])
-             )
+            || (
```
jitendrapurohit, line 383: Looks like 4 spaces instead of 2 @stesi561

jitendrapurohit:
Can just merge ignoring the above styling since this part of code is slightly messed up with spacing issues. But it says WIP in the title -  is there anything else that you plan to add in this PR @stesi561 ?

KarinG:
Thank you Jitendra and Luke. I've changed the title - happy to merge this.

PR #711, opened 2022-02-22
Allow sorting events by title

MegaphoneJon:
Overview
Currently, webforms with User Select for events always sorts by start date.  This allows sorting by title, and opens the door to trivially allowing sort by other fields.

Before
Events sorted by start date.

After
Events sorted by sort date by default; sorting by title is also available.

Comments
On the config form, sorting was reversed from the actual webform.  I've maintained this when sorting by start date.


jitendrapurohit:
Tested, works as expected 👍 . Thanks @MegaphoneJon

PR #712, opened 2022-02-22
Correctly handle custom fields with 'multi-select' checkbox checked

MegaphoneJon:
Overview
Custom fields with the "multi-select" checkbox aren't treated as multi-select fields.

### Steps to replicate
* Create a new custom field on a contact of data type "Alphanumeric", HTML type of "Autocomplete-Select", and the "Multi-Select" box checked.
* Create a new webform with this field.
* Submit this field with multiple values selected.

Before
Only the first value is saved.

After
All values are saved.

Comments
https://github.com/civicrm/civicrm-core/pull/16992 added a new field to `civicrm_custom_field` called `serialize`, and all fields with this set to `TRUE` are multi-select fields.  WFC was unaware of this field and didn't create the fields `extra` array accordingly.

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -1010,6 +1010,9 @@ class Fields implements FieldsInterface {
           $fields[$id]['extra']['description_above'] = (int) empty($custom_field['help_pre']);
           $fields[$id]['has_help'] = TRUE;
         }
+        if ($custom_field['serialize'] || FALSE) {
+          $fields[$id]['extra']['multiple'] = 1;
+        }
         // Conditional rule - todo: support additional entities
         if ($sets[$set]['entity_type'] == 'contact' && !empty($sets[$set]['sub_types'])) {
           $fields[$id]['civicrm_condition'] = [
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -1010,7 +1010,7 @@ class Fields implements FieldsInterface {
           $fields[$id]['extra']['description_above'] = (int) empty($custom_field['help_pre']);
           $fields[$id]['has_help'] = TRUE;
         }
-        if ($custom_field['serialize'] || FALSE) {
+        if (!empty($custom_field['serialize'])) {
           $fields[$id]['extra']['multiple'] = 1;
         }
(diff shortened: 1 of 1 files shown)


src/Fields.php
```diff
@@ -1010,6 +1010,9 @@ protected function wf_crm_get_fields($var = 'fields') {
           $fields[$id]['extra']['description_above'] = (int) empty($custom_field['help_pre']);
           $fields[$id]['has_help'] = TRUE;
         }
+        if ($custom_field['serialize'] || FALSE) {
```
jitendrapurohit: @MegaphoneJon can we get rid of extra `|| FALSE` condition?

MegaphoneJon: Are we certain that the `serialize` key will always be present?  This is meant to prevent an e-notice if it's not (comparable to using `isset()`.

jitendrapurohit: > Are we certain that the serialize key will always be present?

I do get `serialize` for every custom field `get` using API explorer (and on the webform). But i do agree a safe check is required. 

> This is meant to prevent an e-notice if it's not (comparable to using isset().

Ah, is it? It seems `$custom_field['serialize']` will still be checked and trigger e-notice if the key isn't present? 

I tried to unset the key just before this line to test and confirmed that the notice is displayed even with the extra `|| FALSE`. Can you pls clarify? or can we replace it with simple !empty() check since we only need to enter this when `serialize` is present and is equal to 1. Agree @MegaphoneJon ?

colemanw: Right, the addition of `|| FALSE` doesn't add anything to an if statement. It doesn't guard against e-notices and it doesn't affect the logic; it can be removed.
If there is a chance of the key not existing, you can change the line to:
```suggestion
        if (!empty($custom_field['serialize'])) {
```

MegaphoneJon:
Ugh, sorry, I didn't even register that I hadn't done what I intended, which is `?? FALSE`, which *does* guard against e-notices.  But I'll do `!empty()` here.

KarinG:
All ✅ Thank you Jon and Coleman

PR #715, opened 2022-02-25
Make TextFormat the default widget for Activity Details.

KarinG:
Before
Activity Details default widget is text area (no HTML)

After
Now it's a text format -> so with HTML!

image

Comments
It looks like this also removed some old code that was trying to do this - but didn't.

### DIFF
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -19,7 +19,7 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
     'webform_ui',
     'webform_civicrm',
     'token',
-    'CKEditor',
+    'ckeditor',
   ];
 
   /**
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -19,6 +19,7 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
     'webform_ui',
     'webform_civicrm',
     'token',
+    'CKEditor',
   ];
 
   /**
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -405,7 +405,8 @@ class Fields implements FieldsInterface {
       ];
       $fields['activity_details'] = [
         'name' => t('Activity # Details'),
-        'type' => \Drupal::moduleHandler()->moduleExists('webform_html_textarea') ? 'html_textarea' : 'textarea',
+        'type' => 'text_format',
+        'allowed_formats' => [],
       ];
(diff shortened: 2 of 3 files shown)


KarinG:
@jitendrapurohit - will leave this for you to review/merge!

KarinG:
The error is -> 

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ActivitySubmissionTest::testSubmitWebform
Behat\Mink\Exception\ElementNotFoundException: Form field with id|name|label|value|placeholder "Activity Details" not found.
```

KarinG:
Yep - field is not there. Ah perhaps we need to enable CKEditor...

image

KarinG:
Hold on - CKEditor is already installed/enabled - we're using it to test tokens in Email.

KarinG:
@jitendrapurohit - can you spot what I'm missing here? This PR is live on two sites now and Activity Details field is showing up. Why not in the tests here... Do we need a CKEditor config?

tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
```diff
@@ -19,6 +19,7 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
     'webform_ui',
     'webform_civicrm',
     'token',
+    'CKEditor',
```
jitendrapurohit: Dont think we need this. We created a helper function to fill value in ckeditor. See - https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php#L663

So maybe, 

    $this->fillCKEditor('civicrm_1_activity_1_activity_details[value]', activity_details);

should be able to fill values on the webform? @KarinG

KarinG: If I look at the screenshots - the field Activity Details is not on the View of the Webform (with this type change). But on my 2 live sites with this PR it is. Are we missing a Webform submodule in the test?

jitendrapurohit: ah ok - i see you just need to replace `CKEditor` with `ckeditor` the fix the error @KarinG. Can you push that update pls?

KarinG: Ah yes l - will do in a few min - just getting set up in the Aquaplex at the Sport Centre

KarinG: Ok running that on my local now (w/ '`ckeditor`')

KarinG: Passed on my local test env -> pushing now. I think that was it! Thank you for spotting that 👍

KarinG:
Thank you @jitendrapurohit 🎉

PR #717, opened 2022-02-26
Add open-ended monthly donation test using iATS Payments.

KarinG:
Overview
Expanding testing of recurring contributions.

Before
iATS Payments testing included 12 instalments option (an option typically used for e.g. membership payments).

After
Added test for open-ended monthly contribution (an option typically used for e.g. monthly donations).

Same test - same webform - just filled out differently by the user.
Demonstrating also the difference of instalments = 0 vs instalments > 1

PS for a real use case one would not have the donor type in 0 - but for this test it's nice to have that in plain sight.


KarinG:
80% success - I think this well pass on merge (that was a lot of requests to iATS test account push and pull_request) within a short time.

PR #718, opened 2022-02-27
D8/9 - Make Amount Requested default + required on Grant section

jitendrapurohit:
Overview
Make Amount Requested field as default + required on Grant section

Before
Amount Requested is a required field for CiviCRM to create Grant, but is not selected by default on the webform settings page.

Results in a fatal error while submitting the webform. Related - https://www.drupal.org/project/webform_civicrm/issues/3212184

After
Amount Requested field is enabled by default and is also made a required field.

image


KarinG:
🎉

PR #719, opened 2022-03-04
Civigrant

KarinG:
Overview
---------------------------------------

CiviGrant is no longer a component as of 5.47 but an extension.

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -69,13 +69,48 @@ class Fields implements FieldsInterface {
         'CiviEvent' => ['entity_type' => 'participant', 'label' => t('Participant'), 'max_instances' => 9],
         'CiviContribute' => ['entity_type' => 'contribution', 'label' => t('Contribution')],
         'CiviMember' => ['entity_type' => 'membership', 'label' => t('Membership'), 'custom_fields' => 'combined'],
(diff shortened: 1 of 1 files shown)


src/Fields.php
```diff
@@ -69,13 +69,48 @@ protected function getSets(array $components): array {
         'CiviEvent' => ['entity_type' => 'participant', 'label' => t('Participant'), 'max_instances' => 9],
         'CiviContribute' => ['entity_type' => 'contribution', 'label' => t('Contribution')],
         'CiviMember' => ['entity_type' => 'membership', 'label' => t('Membership'), 'custom_fields' => 'combined'],
-        'CiviGrant' => ['entity_type' => 'grant', 'label' => t('Grant'), 'max_instances' => 30, 'attachments' => TRUE],
       ];
       foreach ($conditional_sets as $component => $set) {
         if (in_array($component, $components, TRUE)) {
           $sets[$set['entity_type']] = $set;
         }
       }
```
colemanw: If this was below the new code, then you wouldn't need to repeat the loop. The new code (pre 5.47) could just add to the `$conditional_sets` array and let this loop do its thing.

KarinG: Right - thank you. This was pretty late while helping my 15y old with her trig math :-)

PR #720, opened 2022-03-06
D8/9 - Store free membership in line item table

jitendrapurohit:
Overview
Store free membership in line item table

Before
0 membership amount is not stored in line items. To replicate

- Create a free membership type in civicrm.
- Add this membership to contactA.
- Create a webform with this membership type and a line item.
- Load the webform with cid1 = contactA.
- The 0 membership is loaded as default on the webform.
- Submit the webform with line item.

image

- The contribution created has only single line item and no mention of free membership.

image

After
Free membership is stored in line item table and is also displayed on View contribution page -

image


KarinG:
Thank you Jitendra -> Should this be done for Free Event registrations as well then?

KarinG:
Tested! We can revisit Free Event registrations at a later date.

PR #722, opened 2022-03-07
once function not existing

puresyntax71:
Overview
This is in related to #710. Most of the time the `jquery.once` library is being loaded although there was an instance where it wasn't attached as a library and the contribution page shown an error:

> TypeError: $(...).once is not a function

Before
Error saying:

> TypeError: $(...).once is not a function

After
Added `jquery.once` as a dependency.


KarinG:
Thank you! 🎉

PR #723, opened 2022-03-16
Add maximum field length where applicable

MegaphoneJon:
Overview
If a user submits a value that is longer than the CiviCRM maximum field length (e.g. a first name longer than 64 characters) the form submission fails with an error.

This PR adds the maximum field length to the form's YAML source as a default validation.  Arguably we should also do a server-side check but this is a good first step.

Before
Error when field value is greater than CiviCRM's maximum length.

After
Field can not be longer than CiviCRM's maximum length.

Comments
To date, `src/Fields.php` doesn't really pull data from the Civi API to build the `$fields` array, except on custom fields.  I wrote this with the idea that we *should* be pulling more data from the Civi API's field metadata - but in such a way that we can still override it.

If we want to add more data dynamically we should write a function for it, then call it from `addFieldMetadata()`.

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -6,6 +6,10 @@ class Fields implements FieldsInterface {
 
   protected $components = [];
   protected $sets = [];
+  /**
+   * Store data from CiviCRM API getfields action.
+   */
+  protected $fieldMetadata = [];
 
   public function __construct(UtilsInterface $utils) {
     $this->utils = $utils;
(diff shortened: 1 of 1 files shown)


KarinG:
This is super useful. Thank you for this Jon! @jitendrapurohit could you please bump this to the top of your list for review? I'll run a live test as well this weekend - if I survive security update week 🙂

src/Fields.php
```diff
@@ -1070,6 +1120,12 @@ protected function wf_crm_get_fields($var = 'fields') {
           $fields[$id]['extra']['cols'] = $custom_field['note_columns'] ?? 60;
           $fields[$id]['extra']['rows'] = $custom_field['note_rows'] ?? 4;
         }
+        // Set maximum field length.
+        if (isset($custom_field['text_length'])) {
```
jitendrapurohit: Thanks @MegaphoneJon I've tested this and it works as expected. One minor correction though -

It seems `counter_type`, `counter_maximum`, etc is a property specific to textfields and textarea but with this PR, it seems it adds the counter_type to all the custom fields irrespective of its type? Can we add an additional check for type = text at this point?

`text_length` in civi is a Database field length and is set for checkbox, radio custom fields as well. Setting counter_type for these elements on the webform doesn't fix anything. So seems like adding unnecessary lines to the yaml source.

MegaphoneJon:
Hi Jitendra,

That's a good catch, and I've updated my code to reflect your comments.

I considered leaving out the `textarea` check and just checking for `textfield` but I decided to keep it in as a protection against a user changing the field type of a Webform element, not realizing that CiviCRM won't store the extra characters.  I've seen similar mistakes with other form-building tools when integrated with Civi.

KarinG:
Thank you Jon and Jitendra 🎉 - When tests pass - I'll merge this.

KarinG:
And we are ✅ (testing a lot - a lot - of fields). Have added 'cool stuff' label.

PR #726, opened 2022-03-22
Remove older workarounds from unit test config

demeritcowboy:
Overview
These workarounds _shouldn't_ be needed if only testing 5.45+, and without them will be closer to what a "normal" install would have.


KarinG:
Awesome - thank you Dave!

PR #733, opened 2022-04-26
Issue 3257696 fix undefined index

herbdool:
Fixes https://www.drupal.org/project/webform_civicrm/issues/3257696


jitendrapurohit:
Replicated -

image

Confirmed that the patch fixes the issue.

PR #735, opened 2022-04-29
[Unit tests] "branches" no longer valid syntax

demeritcowboy:
Overview
As of less than a couple hours ago this no longer seems to be a valid parameter to the actions workflow script.


KarinG:
Awesome - thank you Dave! Will merge on pass.

PR #737, opened 2022-05-06
[REF] remove duplicate check for default country

MegaphoneJon:
Overview
A minor refactor I did while troubleshooting an issue.  I saw this code to handle default country, but that same code already exists in `wf_crm_state_abbr`.

Before
Works

After
Still works


mattwire:
@MegaphoneJon Needs rebase - conflicts

PR #739, opened 2022-05-07
D8/9 - Fix reqd error on File upload field

jitendrapurohit:
Overview
Fix Require error on File field 

Before
If File upload field is pre-loaded with a default value, submitting the webform leads to a validation error of "field is required"

https://www.drupal.org/project/webform_civicrm/issues/3277834

image

After
Fixed. No error is displayed on submission.

Technical Details
Webform CiviCRM has a special handling with File upload fields. Looks like when a user uploads a file, wc copies the file to civicrm directory and copies/creates a new file to store in civi. So there are technically 2 files with separate ids.

When the file is pre-loaded with a default civi file id, webform civi simply hides the upload field using js and displays the civi file name to let user know that the file for this entity is already present in civi.

The problem arises when the field is configured as required in webform. Since the file is already present in civi, i think removing the required field attribute from the field makes sense. 

Thoughts @KarinG 

Comments
I've done the following testing and the results seems to be as expected.

- Submit the form with default value selected. The field in civicrm is unaffected.
- Click "Change" and submit the form. The field in civicrm is unaffected.
- Click "Change" and upload a new file. The field in civicrm is updated with the new uploaded file.
- Click "Change" => upload a new file => Remove the file => Submit. The webform gives a validation error.

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -116,8 +116,8 @@ var wfCivi = (function ($, D, drupalSettings) {
         }
       }
       else {
-        $(':visible', container).hide();
-        container.append('<input type="submit" class="button form-submit ajax-processed civicrm-remove-file" value="' + Drupal.t('Remove') + '" onclick="wfCivi.clearFileField(\'' + field + '\'); return false;">');
+        $(container).children().hide();
+        container.append('<input type="submit" class="button form-submit ajax-processed civicrm-remove-file" value="' + Drupal.t('Change') + '" onclick="wfCivi.clearFileField(\'' + field + '\'); return false;">');
       }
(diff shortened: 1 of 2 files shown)


stesi561:
Okay - so when we upload a file via CiviCRM Webform to a civicrm field it gets stored in two places in the file system? I was going to say this seems problematic - however when I tested deleting the activity the file didn't get removed anyway! Furthermore testing a purely CiviCRM workflow - create activity add file to custom field delete yields the same. I think ideally we would only have one copy of the file on the server to make managing deletions and updates easier - and we want to allow deleting the file for a number of reasons including GDPR compliance. https://lab.civicrm.org/dev/core/-/issues/3130 however this is probably well outside of the scope of this issue!

stesi561:
We have the above solution on a site and can confirm it works.

PR #740, opened 2022-05-11
D8/D9 - Javascript error causing form to not load properly: webformId.replace is not a function

deepak-srivastava:
Overview
A js error causing form to not load properly.

Before
JS error: Uncaught TypeError: webformId.replace is not a function.
image

After
No error after fixing the responsbile js code.

Technical Details
webformId.replace throws a js error when webformId is a number, and works fine on a string.


jitendrapurohit:
was able to replicate the error on a numeric webform id

image

Confirmed that the change here fixes the problem.

jitendrapurohit (approved):

PR #736, opened 2022-04-30
D8/9 - Allow -User Select- on Contribution Financial type field

jitendrapurohit:
Overview
----------------------------------------
Allow `-User Select-` on contribution financial type field.

Before
----------------------------------------
Admins need to set Financial Type on the settings form (Contribution Tab).

image

After
----------------------------------------
Can be set to `-User Select-` -

Screenshot 2022-04-30 at 8 33 52 PM

Comments
----------------------------------------
FYI @KarinG @MegaphoneJon

PR #741, opened 2022-05-14
Fix Exception when Donation financial type isn't available in civicrm

jitendrapurohit:
Overview
Fixes https://github.com/colemanw/webform_civicrm/pull/736#issuecomment-1121275626

Before
Exception when CiviCRM doesn't have `Donation` FT available.

`CiviCRM_API3_Exception: Expected one FinancialType but found 0 in civicrm_api3() (line 134 of /var/www/drupal/vendor/civicrm/civicrm-core/api/api.php).`

After

Looks like Donation FTs are pre-existing data in civicrm but are also not reserved ones so anyone can delete them 

image

Fixed the exception by replacing the `getvalue` by a simple `get` api.

Comments
@KarinG


KarinG:
Fails on dev-master - 1stPay timeout. Merging!

KarinG:
Test failure - looking into it:

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ExistingContactElementTest::testTokensInEmail
Undefined offset: 0
```

PR #742, opened 2022-05-21
D8/9 - Fix source field to display on the contact widget using v4 api

jitendrapurohit:
Overview
Fix source field to display on the contact widget using v4 api

Before
Source field fails to load on the contact widget. To replicate -

- Add Existing contact element to the webform.
- Update widget to Autocomplete. Select Source in the `Contact Display Field(s)` input.
- Load the webform and search for a contact.
- No source is displayed

image

After
Source value is displayed -

image

Technical Details
It looks like a core bug in civicrm api v3. It does not return the source value when contact get api is called. But we do get the source using v4. This PR uses the v4 api instead and displays the values on the contact widget. 

Comments
@KarinG

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -92,26 +92,33 @@ class ContactComponent implements ContactComponentInterface {
     $limit = $str ? 12 : 500;
     $ret = [];
     $display_fields = array_values($element['#results_display']);
+    $fieldMappings = [
+      'current_employer' => ['employer_id', 'display_name'],
+      'email' => ['email', 'email'],
+      'phone' => ['phone', 'phone'],
+      'city' => ['address', 'city'],
+      'state_province' => ['address', 'state_province_id:label'],
+      'country' => ['address', 'country_id:label'],
+      'county' => ['address', 'county_id:label'],
+      'postal_code' => ['address', 'postal_code'],
+    ];
+    $joinedTables = [];
+    foreach ($fieldMappings as $field => $type) {
+      if ($key = array_search($field, $display_fields)) {
+        unset($display_fields[$key]);
+        [$table, $fieldName]= $type;
+        $display_fields[] = "{$table}.{$fieldName}";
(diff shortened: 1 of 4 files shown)


jitendrapurohit:
Added another commit to allow searching on all display fields. So eg, if source is one of the display fields on the autocomplete => users can simply enter the source value to get all the matching contacts @KarinG  

image

KarinG:
Great 🎉 - live testing this.

src/ContactComponent.php
```diff
@@ -92,44 +92,63 @@ function wf_crm_contact_search($node, $element, $params, $contacts, $str = NULL)
...
-      else {
-        $params[$search_field] = ['LIKE' => "%$str%"];
+      $searchFields = [];
+      foreach ($display_fields as $fld) {
+        $searchFields[] = [$fld, 'CONTAINS', $str];
       }
+      $params['where'][] = ['OR', $searchFields];
     }
-    $result = $this->utils->wf_civicrm_api('contact', 'get', $params);
+    $result = $this->utils->wf_civicrm_api4('Contact', 'get', $params);
```
demeritcowboy, line 148: The recent test fails seems to be from here, since api4 checks perms by default, so the existing contact dropdown doesn't have Default Org in it since the anon user doesn't have permission.

And then lower down in wf_crm_contact_access where the other api4 replacement was made, the logic will need to change to account for the change in default. And as written currently it's assuming all callers to wf_crm_contact_access will have been updated to pass in filters['checkPermissions'] instead of filters['check_permissions']

KarinG:
I'm going to revert this PR - as it will help us debug the current test failures.

KarinG:
@jitendrapurohit - could you please have another pass at this when you have a chance?

jitendrapurohit:
Resubmitted with the test fix - https://github.com/colemanw/webform_civicrm/pull/785

PR #744, opened 2022-05-21
D8/9 - Fix error on current employer submission

jitendrapurohit:
Overview
Fix error on current employer submission

Before
Current employer field leads to an error if submitted anonymously. Replicate in the test -

- Create a webform with 2 contacts. Individual & Organization
- Enable Contribution tab and select dummy processor.
- Enable current employer field on the 1st contact and set it to "Contact 2".
- Edit the existing contact field for second contact and make it a select widget + disable "Enforce permissions" field.
- Load the webform as an anonymous user.
- Select an org in the Contact 2 field and submit the form.
- Error.

After
Fixed.

Technical Details
Current employer is a contact ref and it needs to be saved after ref fields are updated with actual contact ids. Billing Contact is saved during the validation phase where the ref ids aren't created yet. 

We do the same thing during a normal contact submission - https://github.com/colemanw/webform_civicrm/blob/6.x/src/WebformCivicrmPostProcess.php#L210

Adding the same statement while creating the billing contact as well.

Comments
@KarinG @MegaphoneJon

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1875,6 +1875,8 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       'location_type_id' => 'Billing',
     ];
     if (!$cid) {
+      // Current employer must wait for ContactRef ids to be filled
+      unset($contact['contact'][1]['employer_id']);
       $cid = $this->createContact($contact);
       $this->billing_contact = $cid;
     }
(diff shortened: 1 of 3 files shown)


KarinG:
@MegaphoneJon - can you please confirm this fixes the issue you pointed out?

MegaphoneJon:
My client confirms this is fixed.

KarinG:
Thank you Jon!

PR #745, opened 2022-05-22
D8/9 - Fix contact subtype for dedupe check

jitendrapurohit:
Overview
Fix contact subtype for dedupe check

Before
Dedupe test fails with api error.

After
Pass.

Technical Details
Looks like WC has `strtolower()` done for subtypes all over the place. Not sure why it was done that way (maybe for consistency?). But if we change that now, i think we'll also need to handle the existing webforms via upgrade code.
How about just fix it with something like this? We also do this for contact type (just above the line i added). The commit adds a similar check for contact sub type.

Comments
@KarinG


KarinG:
Agree on the save fix. The strtolower are dating back to D7WFC and there are so man of them. 

I think this may require a D7 backport?

PR #749, opened 2022-05-29
D8/9 - Add hash to file name before saving the uploaded attachment

jitendrapurohit:
Overview
File filename before saving to civicrm

Before
Filename is not hashed before saving to civicrm database.

After
Filename is hashed.

Technical Details
Call to core function CRM_Utils_File::makeFileName().

Alternative fix could have been the usage of Attachment api but it requires an entity id to be passed in params. We dont necessarily have entity id at this point in every case so can't use this.
 
Api4 doesn't have attachment api and `File` api doesn't really changes the uri param as of now.

Seems like we still need to use the CRM function.

Comments
@KarinG @stesi561 fyi

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -784,7 +784,8 @@ abstract class WebformCivicrmBase {
     $file = File::load($id);
     if ($file) {
       $config = \CRM_Core_Config::singleton();
-      $path = \Drupal::service('file_system')->copy($file->getFileUri(), $config->customFileUploadDir);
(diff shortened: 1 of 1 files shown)


stesi561:
This is now conflicting. 

I'm unclear from above what the goal of hashing is here. Is it to prevent collisions? I got distracted testing and found -> https://www.drupal.org/project/webform_civicrm/issues/3493702

jitendrapurohit:
looks like a 2 year older patch. I cant seem to find a reason on our local issue queue that led to this PR. Closing for now.

PR #755, opened 2022-06-16
Update some deprecated phpunit syntax

demeritcowboy:
Overview
* void return typehint on setUp/tearDown
* assertContains on string


KarinG:
😍😍😍

PR #756, opened 2022-06-18
D8/9 - Add test to capture case update error

jitendrapurohit:
Overview
Test to verify case update using authenticated user.

Before
No test for updating case with non admin user.

After
Test added. Case update results into the following error -

image

Comments
This fails the build due to a bug in civicrm. Probably related to this line in the commit - https://github.com/civicrm/civicrm-core/commit/0501e0eac68d304debeb723da96a96165f272d61#diff-316f8ac3ece43544ca077df1239657e89d5b67919831549e396312179bb4bddbR182, where a api4 call is made without `checkPermissions`?

Related SE - https://civicrm.stackexchange.com/questions/42109/cannot-change-cases-via-drupal-webform-after-civicrm-update

### DIFF
--- a/tests/src/FunctionalJavascript/CaseSubmissionTest.php
+++ b/tests/src/FunctionalJavascript/CaseSubmissionTest.php
@@ -11,13 +11,10 @@ use Drupal\Core\Url;
  */
 final class CaseSubmissionTest extends WebformCivicrmTestBase {
 
-  /**
-   * Test Case Submission.
-   */
-  public function testCaseSubmission() {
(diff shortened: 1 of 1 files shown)


KarinG:
Hi Luke @stesi561 - here's an example of a regression reported by a user on SE on Jun14 and now captured in a WFC test last night. I'll post a screenshot in PM Channel and will ping you. Since this involves API i.e. Coleman I don't think you'll have to negotiate warranty for WFC users :-)

KarinG:
Merging this to lock it in.

PR #757, opened 2022-06-27
Skip test for 5.49 since will never pass

demeritcowboy:
Overview
While there'll probably be a 5.50 release with a fix for #756, I don't think there'll be a 5.49 version, so this test will never pass for 5.49. Next ESR should be 5.51 next month.


KarinG:
I actually prefer it will continue to fail this test because it does: 5.49.x does not allow certain WFC users to update cases. At some point 5.49 will rotate out of the test matrix and then we’ll be testing all greens again.

demeritcowboy:
Ok.

PR #360, opened 2020-09-22
D7: make Contact selectable for Contributions

sluc23:
Overview
----------------------------------------
In `webform_civicrm` when creating a Contribution, it's assigned to **Contact 1** by default.  
There are some use cases where we need to create several Contacts in the same webform or having conditional contacts creation (Individual or Organization in the same webform selectable by the user, for instance) and we don't want to assign the new Contribution (one off and/or recurring) to **Contact 1**.

This PR adds the possibility to select to which contact the contribution will be assigned, or selectable by the User as well, that could be manual assigned, or use in different scenarios (using webform_conditional to assing the contribution to Contact X depending on criteria)  

webform_civicrm

D7 or D8?
----------------------------------------
Drupal 7

Before
----------------------------------------
The Contribution is always assigned to **Contact 1**

After
----------------------------------------
The Contribution can be assigned to any Contact available in the webform, even *selectable by user*, adding more webforms' use cases 

Technical Details
----------------------------------------
This PR adds `contribution_contact_id` field as part of the Contribution fields

Comments
----------------------------------------
I'm adding this to `7.x-5.x` branch because it's the version we are still using in Drupal 7. I won't be ready to port this to Drupal 8 at the moment

PR #758, opened 2022-06-28
D8/9: make Contact selectable for Contributions

sluc23:
Overview
Adds a new feature for D9 version,  included in iXiam's D7 fork, to be able to select for the Contribution, which Contact in the webform (if there are more than 1) will the owner of it.
This opens many new use cases that can be used in Webforms with multiple contacts (father/son, individual/company, etc.. where not always the contribution must be assigned to Contact 1)

webform_civicrm

Before
If the Webform contains Contribution and several Contacts, **ALWAYS** the contribution is created for **Contact 1**

After
User can select which Contact added in the Webform would be the one assigned to the contribution, or even could be selected by the **End User** in the same Webform with the option `- User Select -`

Technical Details
This feature was offered to be included in D7 main repo in [PR #360](https://github.com/colemanw/webform_civicrm/pull/360) , but we didn't succeed, so we started using [iXiam's fork](https://github.com/ixiam/webform_civicrm/pull/1) with this feature included

Comments
This new feature takes in account default historically `webform_civicrm` behavior.. so by design always **Contact 1** is the selected by default in the Contribution Tab

Following the new validation added of checking that Email field is mandatory for the **Contact 1**, if Contribution is enabled, this new feature checks that Email field is mandatory for the **Contact x** selected as the assigned.
If the option selected is `- User Select -`, there's no email checking

### DIFF
--- a/js/webform_civicrm_admin.js
+++ b/js/webform_civicrm_admin.js
@@ -542,15 +542,19 @@ var wfCiviAdmin = (function (D, $, once) {
       }).change();
 
       function billingMessages() {
+        var contactId = $('[name=civicrm_1_contribution_1_contribution_contact_id]').val();
+        if (typeof contactId == 'undefined') {
+          contactId = 1;
+        }
         var $pageSelect = $('[name=civicrm_1_contribution_1_contribution_enable_contribution]');
         // Warning about contribution page with no email
-        if ($pageSelect.val() !== '0' && ($('[name=civicrm_1_contact_1_email_email]:checked').length < 1 || $('[name=contact_1_number_of_email]').val() == '0')) {
-          var msg = Drupal.t('You must enable an email field for :contact in order to process transactions.', {':contact': getContactLabel(1)});
+        if ($pageSelect.val() !== '0' && ($('[name=civicrm_' + contactId + '_contact_1_email_email]:checked').length < 1 || $('[name=contact_' + contactId + '_number_of_email]').val() == '0')) {
+          var msg = Drupal.t('You must enable an email field for :contact in order to process transactions.', {':contact': getContactLabel(contactId)});
           if (!$('.wf-crm-billing-email-alert').length) {
             $pageSelect.after('<div class="messages error wf-crm-billing-email-alert">' + msg + ' <button>' + Drupal.t('Enable It') + '</button></div>');
             $('.wf-crm-billing-email-alert button').click(function() {
-              $('input[name=civicrm_1_contact_1_email_email]').prop('checked', true).change();
-              $('select[name=contact_1_number_of_email]').val('1').change();
+              $('input[name=civicrm_' + contactId + '_contact_1_email_email]').prop('checked', true).change();
+              $('select[name=contact_' + contactId + '_number_of_email]').val('1').change();
               return false;
             });
             if ($('.wf-crm-billing-email-alert').is(':hidden')) {
@@ -569,8 +573,12 @@ var wfCiviAdmin = (function (D, $, once) {
           $('#edit-participant').prepend('<div class="wf-crm-paid-entities-info messages status">' + Drupal.t('Configure the Contribution settings to enable paid events.') + '</div>');
         }
       }
-      $(once('email-alert', '[name=civicrm_1_contribution_1_contribution_enable_contribution], [name=civicrm_1_contact_1_email_email]', context)).change(billingMessages);
-      billingMessages();
+      // Get contact assigned to contribution
+      var contactIdForContrib = $('[name=civicrm_1_contribution_1_contribution_contact_id]').val();
+      if (contactIdForContrib !== "create_civicrm_webform_element") {
+        $(once('email-alert', '[name=civicrm_1_contribution_1_contribution_enable_contribution], [name=civicrm_' + contactIdForContrib + '_contact_1_email_email]', context)).change(billingMessages);
+        billingMessages();
+      }
 
       // Handlers for submit-limit & tracking-mode mini-forms
       $(once('wf-civi', '#configure-submit-limit', context)).click(function() {
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -672,6 +672,16 @@ class Fields implements FieldsInterface {
           'parent' => 'contribution_pagebreak',
         ];
         // @todo moved in order since we can't pass `weight`.
+        $fields['contribution_contact_id'] = [
+          'name' => t('Contact'),
+          'type' => 'select',
+          'expose_list' => TRUE,
+          'data_type' => 'ContactReference',
+          'extra' => [
+            'aslist' => 1,
+            'required' => TRUE
+          ],
+        ];
         $fields['contribution_total_amount'] = [
             'name' => 'Contribution Amount',
             'parent' => 'contribution_pagebreak',
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -84,6 +84,10 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
 
     $this->settings = $handler->getConfiguration()['settings'];
     $this->data = $this->settings['data'];
+
+    // Decouple of the configuration of the selected contact in admin page with the contact assigned to the contribution
+    unset($this->data['contribution'][1]['contribution'][1]['contact_id']);
+
     $this->enabled = $this->utils->wf_crm_enabled_fields($this->node);
     $this->all_fields = $this->utils->wf_crm_get_fields();
     $this->all_sets = $this->utils->wf_crm_get_fields('sets');
@@ -1872,16 +1876,17 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
    */
   private function createBillingContact() {
     $cid = wf_crm_aval($this->existing_contacts, 1);
+    $contact_ref_contribution = $this->settings['data']['contribution'][1]['contribution'][1]['contact_id'] ?? $this->rawValues["civicrm_1_contribution_1_contribution_contact_id"];
     if (!$cid) {
-      $contact = $this->data['contact'][1];
+
+      $contact = $this->data['contact'][$contact_ref_contribution];
       // Only use middle name from billing if we are using the rest of the billing name as well
-      if (empty($contact['contact'][1]['first_name']) && !empty($this->billing_params['middle_name'])) {
-        $contact['contact'][1]['middle_name'] = $this->billing_params['middle_name'];
+      if (empty($contact['contact'][$contact_ref_contribution]['first_name']) && !empty($this->billing_params['middle_name'])) {
+        $contact['contact'][$contact_ref_contribution]['middle_name'] = $this->billing_params['middle_name'];
       }
-      $contact['contact'][1] += [
-        'first_name' => $this->billing_params['first_name'] ?? NULL,
-        'last_name' => $this->billing_params['last_name'] ?? NULL,
-      ];
+      $contact['contact'][1]['first_name'] = $this->billing_params['first_name'] ?? NULL;
+      $contact['contact'][1]['last_name'] = $this->billing_params['last_name'] ?? NULL;
+
       $cid = $this->findDuplicateContact($contact);
     }
     $address = [
@@ -1898,7 +1903,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     ];
     if (!$cid) {
       // Current employer must wait for ContactRef ids to be filled
-      unset($contact['contact'][1]['employer_id']);
+      unset($contact['contact'][$contact_ref_contribution]['employer_id']);
       $cid = $this->createContact($contact);
       $this->billing_contact = $cid;
     }
@@ -1916,7 +1921,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       }
     }
     if ($cid) {
-      $address['contact_id'] = $email['contact_id'] = $this->ent['contact'][1]['id'] = $cid;
+      $address['contact_id'] = $email['contact_id'] = $this->ent['contact'][$contact_ref_contribution]['id'] = $cid;
       // Don't create a blank billing address.
       if ($address['street_address'] || $address['city'] || $address['country_id'] || $address['state_province_id'] || $address['postal_code']) {
         $this->utils->wf_civicrm_api('address', 'create', $address);
@@ -2182,7 +2187,10 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     $params['financial_type_id'] = wf_crm_aval($this->data, 'contribution:1:contribution:1:financial_type_id');
     $params['currency'] = $params['currencyID'] = wf_crm_aval($this->data, "contribution:1:currency");
     $params['skipRecentView'] = $params['skipLineItem'] = 1;
-    $params['contact_id'] = $this->ent['contact'][1]['id'];
+    ## Deb
+    // For the "civicrm_1_contribution_1_contribution_contact_id", if webform civicrm admin is configured($this->settings), get this value, if not the user selected value will be used ($this->submission).
+    $contact_ref_contribution = $this->settings['data']['contribution'][1]['contribution'][1]['contact_id'] ?? $this->rawValues["civicrm_1_contribution_1_contribution_contact_id"];
+    $this->data['contribution'][1]['contribution'][1]['contact_id'] = $params['contact_id'] = $this->ent['contact'][$contact_ref_contribution]['id'];
     $params['total_amount'] = round($this->totalContribution, 2);
 
     // Most payment processors expect this (normally be set by contribution page processConfirm)
--- a/webform_civicrm.install
+++ b/webform_civicrm.install
@@ -336,3 +336,27 @@ function webform_civicrm_update_8004() {
     $webform->save();
   }
 }
+
+/**
+ * Set Contact 1 as the assigned for Contributions in all webforms
+ * (It keeps default behavior <= 6.2.0)
+ */
+function webform_civicrm_update_8005() {
+  \Drupal::service('civicrm')->initialize();
+  $webforms = Webform::loadMultiple();
+  foreach ($webforms as $webform) {
+    $handler = $webform->getHandlers('webform_civicrm');
+    $config = $handler->getConfiguration();
+    if (empty($config['webform_civicrm'])) {
+      continue;
+    }
+    $settings = &$config['webform_civicrm']['settings'];
+    // If Contribution enabled, set Contact 1 assigned as default behavior <= 6.2.0
+    if ($settings['civicrm_1_contribution_1_contribution_enable_contribution'] == "1") {
+      $settings['civicrm_1_contribution_1_contribution_contact_id'] = 1;
+      $settings['data']['contribution'][1]['contribution'][1]['contact_id'] = 1;
+      $handler->setConfiguration($config);
+      $webform->save();
+    }
+  }
+}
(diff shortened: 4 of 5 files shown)


sluc23:
@KarinG I'm struggling a little bit with the Unit Tests.. this PR won't pass some of the existing tests (I guess it's a matter of string comparisons..) But I'm not able to setup a proper Unit Test environment (without PHPStorm) to be able the re-do all those tests, and add some specific tests for this new feature.

I think I'll need a hand to sort out the tests part of this PR.
tx

demeritcowboy:
@sluc23 My own environment doesn't have phpstorm. I can probably help out. There are these docs but they are a bit out of date now: https://docs.civicrm.org/dev/en/latest/testing/mink/#installation. I'll try to set one up from scratch and see what's new.

demeritcowboy:
@sluc23 Those mink docs still worked pretty well for me. I just had one real issue and then some optional stuff. And then there's obviously some extra steps to get webform installed. So

1. Follow the docs at https://docs.civicrm.org/dev/en/latest/testing/mink
   * I had phpcs installed globally which conflicted with installing `drupal/core-dev`. Most people probably won't have that. It just involves appending another entry to the repositories section in composer.json first: `{"type": "composer", "url": "https://packagist.org"}`
1. In step 5 there where you set BROWSERTEST_OUTPUT_DIRECTORY, also make sure to create that folder in the filesystem, otherwise you'll get a scary warning and screenshots won't get created.
   * Note though that no matter what folder you specify, the screenshots end up in web/sites/simpletest/browser_output (known issue)
1. (Optional) Uncomment this line in phpunit.xml and set its value to 999999: `<env name="SYMFONY_DEPRECATIONS_HELPER" value="999999"/>`

Then to get webform installed:
1. composer require drupal/token:'^1.9' drupal/webform:6.x-dev drupal/webform_civicrm:6.x-dev
   * but note what you get depends on when it was last sync'd to drupal's repo (https://git.drupalcode.org/project/webform_civicrm/-/commits/6.x/) which might be a little older. If you need the latest-latest you can: `cd web/modules/contrib/webform_civicrm; git remote add ghub https://github.com/colemanw/webform_civicrm.git; git pull ghub 6.x`
1. To run the tests it's the same as in the mink docs, just specifying appropriate paths. So if you are in the top-level folder, above web and vendor, and that's where your phpunit.xml is, it would be, e.g.
   * `vendor/bin/phpunit -c phpunit.xml --filter testSelectContactElement web/modules/contrib/webform_civicrm/tests/src/FunctionalJavascript/ContactSubmissionTest.php`
   * The `--filter` is just to pick one test. If you want the whole suite (not recommended unless you are going out for a few hours), leave out the filter and just specify web/modules/contrib/webform_civicrm/tests/src/FunctionalJavascript as the path.

sluc23:
@demeritcowboy thanks a lot for your help.. I'm almost there !!
I was able to setup a `docker-compose` stack  with all needed containers to run the tests, including image [selenium/standalone-chrome](https://github.com/SeleniumHQ/docker-selenium), which is very handy to have the selenium in a docker, with a VNC connection to see what's going on Chrome when running a test in **real-time**. :sunglasses: 

The last issue I've found, is when I'm running the tests, they failed, because Drupal is not installed, and something must be wrong, because the test do not install Drupal at bootstrap (which I understood that's what should happen, right?)

For instance, I got this response running this test:
`phpunit -c ./phpunit.xml web/modules/contrib/webform_civicrm/tests/src/FunctionalJavascript/ContributionDummyTest.php`

```bash
HTML output directory web/sites/default/browsertests is not a writable directory.
PHPUnit 9.5.21 #StandWithUkraine

Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

Testing Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest
EEEE                                                                4 / 4 (100%)

Time: 01:04.160, Memory: 98.00 MB

There were 4 errors:

1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest::testOnePageDonation
Behat\Mink\Exception\ElementNotFoundException: Button with id|name|label|value "Log in" not found.

/var/www/html/web/core/tests/Drupal/Tests/WebAssert.php:148
/var/www/html/web/core/tests/Drupal/Tests/UiHelperTrait.php:77
/var/www/html/web/core/tests/Drupal/Tests/UiHelperTrait.php:257
/var/www/html/web/modules/contrib/webform_civicrm/tests/src/FunctionalJavascript/ContributionDummyTest.php:20
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestResult.php:726

```

and Selenium output is this one:
selenium


I'll keep on investigating what could be wrong..

Once I have this working, I'll wrap it out everything to share how to setup this with docker
cheers!

demeritcowboy:
Cool!
In phpunit.xml, there's a line near the top that has `bootstrap=`. It should point to web/core/tests/bootstrap.php since you are in the top level folder.
Also make sure the SIMPLETEST_DB setting points to an existing empty database with proper credentials.
And I see you have a separate error: You need to create the folder web/sites/default/browsertests.

KarinG:
Test failures are in the Contributions tests -> 

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest::testOnePageDonation
Behat\Mink\Exception\ResponseTextException: The text "You must enable an email field for Contact 1 in order to process transactions." was not found anywhere in the text of the current page.
```

If we do select 
`    $this->getSession()->getPage()->selectFieldOption('civicrm_1_contribution_1_contribution_enable_contribution', 1);`
and if we only have one Contact - then we still need 
`    $this->assertSession()->pageTextContains('You must enable an email field for Contact 1 in order to process transactions.');`

We've started consolidating this bit of the tests in WebformTestBase.php - function configureContributionTab(). 

Also we really do need a test for a Contribution assigned to Contact 2. I think you got really close to getting Mink tests to work on your dev! If needed we can help with that.

sluc23:
yeap @KarinG .. we are moving slowly (but forwards) to make tests working on.. It's useful for us as well to understand all the unit tests Eco-system and develop our own tests for this.
I agree with your comments.. so let give us some more time to adjust them and I get back to you if we have any blocking issue.

KarinG:
We're happy to help you see this through + help get you set up with Mink testing.

Have you seen the cool screenshots your PR run produced?

For example:
image

That screenshot comes from here -> 
https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/CustomFieldSubmissionTest.php#L571
`$this->createScreenshot($this->htmlOutputDirectory . '/ajaxvalues.png');`

Also tons of HTML output -> (this happens on button pressing - but can also be requested via:
$this->htmlOutput();

https://github.com/colemanw/webform_civicrm/actions/runs/2910091254 -> go down to ->
image

KarinG:
Oh - I see at Test 🎉

KarinG:
Test fail:

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest::testAssignContributionSecondContact
WebDriver\Exception\UnexpectedAlertOpen: unexpected alert open: {Alert text : Email Required
You must enable an email field for Contact 2 in order to process transactions.}
  (Session info: headless chrome=106.0.5249.91)
  (Driver info: chromedriver=106.0.5249.61 (511755355844955cd3e264779baf0dd38212a4d0-refs/branch-heads/5249@{#569}),platform=Linux 5.15.0-1020-azure x86_64)
```

rubofvil:
@KarinG thx for the revision, we fixed the test in the last commit

KarinG:
Excellent - thank you for seeing this through! I'll ask Jitendra to (final) review this when he has a next block of time [later this week].

jitendrapurohit:
Existing webforms using contributions return a fatal error after this PR (even after drush updb) -

image

webform_civicrm.install
```diff
@@ -302,3 +302,27 @@ function webform_civicrm_update_8003() {
...
+  foreach ($webforms as $webform) {
+    $handler = $webform->getHandlers('webform_civicrm');
+    $config = $handler->getConfiguration();
+    if (empty($config['webform_civicrm'])) {
+      continue;
+    }
+    $settings = &$config['webform_civicrm']['settings'];
+    // If Contribution enabled, set Contact 1 assigned as default behavior <= 6.2.0
+    if ($settings['civicrm_1_contribution_1_contribution_enable_contribution'] == "1") {
+      $settings['civicrm_1_contribution_1_contribution_contact_id'] = 1;
```
jitendrapurohit, line 356: @sluc23 I think you also need to add it to the $settings['data']? Something like -

    $settings['data']['contribution'][1]['contribution'][1]['contact_id'] = 1;

rubofvil: Thx @jitendrapurohit we are testing this solution, and we will give feedback this week

rubofvil: @jitendrapurohit  added the missing params and tested
The test:
- Installed webform_civicrm with a previous release. Enabled contribution tab.
- Updated the webform_civicrm with ixiam:6.x#DevAssingContribs, executed the updb. Made a submit into webform. Verified that the contribution was assigned correctly to the contact.

Thx for the previous advice!

rubofvil:
@KarinG @jitendrapurohit thank you for everything, 
Can we do something on our side to make progress on this issue?

KarinG:
@rubofvil - sorry for the delay -> it's on our list the review - for this weekend.

src/WebformCivicrmPostProcess.php
```diff
@@ -2143,7 +2147,8 @@ private function contributionParams() {
     $params['financial_type_id'] = wf_crm_aval($this->data, 'contribution:1:contribution:1:financial_type_id');
     $params['currency'] = $params['currencyID'] = wf_crm_aval($this->data, "contribution:1:currency");
     $params['skipRecentView'] = $params['skipLineItem'] = 1;
-    $params['contact_id'] = $this->ent['contact'][1]['id'];
+    $contact_ref_contribution = $this->settings['data']['contribution'][1]['contribution'][1]['contact_id'];
+    $this->data['contribution'][1]['contribution'][1]['contact_id'] = $params['contact_id'] = $this->ent['contact'][$contact_ref_contribution]['id'];
```
jitendrapurohit, line 2193: I think there's still some issues at this point of the code. `$contact_ref_contribution` is not set when Contribution contact is set to `-User select-`.

image

Since no contact is found, civicrm tries to create a contribution without a contact and webform submission eventually fails with the following error -

image

To replicate: 

Create a webform with 2 contacts and set contribution contact to `-User select-`. Submit the webform => Error.

Fix:

I think we should look for the `$contact_ref_contribution` only when `$this->settings['data']['contribution'][1]['contribution'][1]['contact_id']` is NOT set to `-User select-`. Eg

````
$contact_ref_contribution = wf_crm_aval($this->settings, 'data:contribution:1:contribution:1:contact_id');
    if (!empty($contact_ref_contribution)) {
      $this->data['contribution'][1]['contribution'][1]['contact_id'] = $params['contact_id'] = $this->ent['contact'][$contact_ref_contribution]['id'];
    }
    else {
      // Contribution contact is set to -User select-. At this point, if the fillContactRefs() function is executed,
      // the contact id is already set in $this->data['contribution']. But, if this is called during validation phase (eg submit live payment),
      // contact id is still a reference and need to be replaced with the billing contact id.
      $contact_ref_contribution = wf_crm_aval($this->data, 'contribution:1:contribution:1:contact_id');
      if (!empty($contact_ref_contribution)
        && !(isset($this->ent['contact'][$contact_ref_contribution]) && $contact_ref_contribution == $this->ent['contact'][$contact_ref_contribution]['id'])) {
          $this->data['contribution'][1]['contribution'][1]['contact_id'] = $params['contact_id'] = $this->ent['contact'][$contact_ref_contribution]['id'];
      }
    }
```` 

Does that help @sluc23 @rubofvil ?

File: src/Fields.php [hunk 3]
```diff
@@ -661,6 +661,13 @@ protected function wf_crm_get_fields($var = 'fields') {
           'parent' => 'contribution_pagebreak',
         ];
         // @todo moved in order since we can't pass `weight`.
+        $fields['contribution_contact_id'] = [
+          'name' => t('Contact'),
+          'type' => 'select',
+          'expose_list' => TRUE,
+          'value' => 1,
```
jitendrapurohit: I think this line `'value' => 1,` should be removed? It forces the submission of 1 when contribution contact is set to `-user select-` in civicrm settings.

Do you agree @sluc23 @rubofvil ? To replicate the related error here: 

- Enable -User select- for contribution contact field.
- Submit the webform as anonymous and select Contact 2 on the second page as contribution contact.
- After submit - Contribution is created on first contact (Error).

jitendrapurohit (changes requested):
Thanks @sluc23 @rubofvil. The upgrade statement is working now. 

But, I think there is still problems with this PR. Eg createBillingContact is always executed for the 1st contact in the webform https://github.com/colemanw/webform_civicrm/blob/6.x/src/WebformCivicrmPostProcess.php#L1869

So, submitting live payments are still recorded on contact 1.

src/Fields.php
rubofvil: @jitendrapurohit  thx for the feedback and clues!

We added a fix and a test to validate "when contribution contact is set to -user select- in civicrm settings."
Could you try again?

It's important for test again, remove the value into config of the webform. Like in the attached image(remove the value in red). **If you create a new webform, after pull, is not required this step.**
Ref https://github.com/colemanw/webform_civicrm/pull/758/commits/760a7a9b976b377763d7c919e896d1bf50a50161#diff-5b105b68fdf928d8ddea9d9f72602752c175e66553b49fd839181692f96d9fd0L668


image

src/WebformCivicrmPostProcess.php
rubofvil: Yes, @jitendrapurohit. 

Fixed in the last commit. Seems ok for you this solve?

Ref: https://github.com/colemanw/webform_civicrm/commit/760a7a9b976b377763d7c919e896d1bf50a50161#diff-c5d6d9f4f2a6106846c858b8e3b25ee6ecb15384b855ecdcbfe52533a39ad877R2150

jitendrapurohit:
Thanks for the update @sluc23. Before retesting this - do you have any thoughts on -

> But, I think there are still problems with this PR. Eg createBillingContact is always executed for the 1st contact in the webform https://github.com/colemanw/webform_civicrm/blob/6.x/src/WebformCivicrmPostProcess.php#L1869
> 
> So, submitting live payments are still recorded on contact 1.

rubofvil:
@jitendrapurohit thx, we made a change to assign the billing to the contributor. We don't use this option(billing), Could you take a look if we need to make some extra test?
https://github.com/colemanw/webform_civicrm/pull/758/commits/0d38ccc17951d6df31a366d67ac748455f763089

> Thanks for the update @sluc23. Before retesting this - do you have any thoughts on -
> 
> > But, I think there are still problems with this PR. Eg createBillingContact is always executed for the 1st contact in the webform https://github.com/colemanw/webform_civicrm/blob/6.x/src/WebformCivicrmPostProcess.php#L1869
> > So, submitting live payments are still recorded on contact 1.

KarinG:
@rubofvil - could you please rebase so that we can run the tests?

rubofvil:
@KarinG We made the rebase

* Seems there is a error in branch 6.x with, but no related with this dev.
(Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionIatsTest::testSubmit1stPayContribution Cryptogram)

> @rubofvil - could you please rebase so that we can run the tests?

KarinG:
@rubofvil - agreed -> it's unrelated. A change in the most recent D9.4 release (that's why the D9.3 test passes for you) is somehow causing the card number to be invalid shortly after it has been validated:

image

I'm working on getting that sorted.

jitendrapurohit:
Hmm, not yet working @rubofvil sorry.

I did another test and get the following fatal error with the webform + dummy processor payment.

To replicate -

- Create a webform with 2 contacts -> Contribution contact = user select and enable dummy processor with billing fields.
- Go anonymous and Submit the webform with contribution contact = Contact 2.
- Error.

image

If this is turning out to be too-complicated, I'm wondering if we really need `-User select-` option on the contribution contact? Maybe, we can limit the admins to select which contact gets the contribution and just get rid of `-user select-` option (Unless we really want users to select the contacts while filling the form)

KarinG:
I do like the idea of simplifying this to -not- making this a user select -> but letting the admin decide which Contact # the Contribution gets assigned to. That will still accomplish the primary purpose of this PR which is to -not- automatically make Contact 1 the contact that gets the Contribution attached to their record.

francescbassas:
I have an eye on here. If I have not misunderstood, I think that if letting the admin decide which Contact the Contribution gets assigned to, then the use case individual/org unique webform will not be fulfilled. See https://civicrm.stackexchange.com/questions/21173/how-to-build-a-drupal-webform-for-individual-organization-contributors.

rubofvil:
Thx for review this case @jitendrapurohit . It's fixed and added a test to verify that works correctly https://github.com/colemanw/webform_civicrm/pull/758/commits/eee42435165c987b2bf4f66f94654ce2149846dd

> I have an eye on here. If I have not misunderstood, I think that if letting the admin decide which Contact the Contribution gets assigned to, then the use case individual/org unique webform will not be fulfilled. See https://civicrm.stackexchange.com/questions/21173/how-to-build-a-drupal-webform-for-individual-organization-contributors.

sluc23:
@KarinG we do really need the feature to `- User Select -` contact for contribution in the webform. This adds the possibility to have only 1 one webform for Organizations or Individuals in the same form where the user can choose what type of contact creates (and later assign the contribution to it), instead of having the very same 2 webforms duplicated because we cannot create one or the other.. This is actually one of the requests we received the most from our customers..

I think we are very close to get it... @rubofvil has submitted a fix on the latest comment.
We appreciate a lot your feedback and effort putting on this PR.. I'm confident we will make it :muscle: 

cheers!

KarinG:
@sluc23 and @rubofvil - thank you for sticking with this. Yes I think we're close. I'll ask @jitendrapurohit to have another review of it. 

I'm still fighting with the cryptogram (the test failure). It's the result of a change in one of the test functions. If I can't figure it out I'm going to remove that test for now.

jitendrapurohit:
ok - retesting with the same usecase as mentioned in my previous comment - https://github.com/colemanw/webform_civicrm/pull/758#issuecomment-1356150644

This time - Contact 1 was not created in civicrm at all. The webform submission only creates Contact 2 and assigns the contribution to it. @rubofvil @sluc23 Seems like the values filled for contact 1 does not get evaluated with contact create api?

js/webform_civicrm_admin.js
```diff
@@ -541,16 +541,20 @@ var wfCiviAdmin = (function (D, $, once) {
         }
       }).change();
 
-      function billingMessages() {
+      function billingMessages(contactId) {
```
jitendrapurohit: Is this param required @rubofvil ? I dont see it being used on any call.

jitendrapurohit:
I'm happy to work on this @rubofvil @sluc23. I think it also requires some work on admin js file as the alert message isn't correct for every option selected in the contribution contact field.

Can you please squash all the commits so its easy to cherrypick them?

Thanks.

js/webform_civicrm_admin.js
rubofvil: This isn't required, we cleaned in the last rebase, thx

rubofvil:
@jitendrapurohit  we made the squash, thx for the work. If we can made some adjust in the alert settings tell us

> I'm happy to work on this @rubofvil @sluc23. I think it also requires some work on admin js file as the alert message isn't correct for every option selected in the contribution contact field.
> 
> Can you please squash all the commits so its easy to cherrypick them?
> 
> Thanks.

jitendrapurohit:
Have raised https://github.com/colemanw/webform_civicrm/pull/821 to fix outstanding issues @rubofvil @sluc23

sluc23:
merged on https://github.com/colemanw/webform_civicrm/pull/821
closing!

PR #759, opened 2022-06-29
Update data-amount with .attr() function

erawat:
Overview
Webform CiviCRM uses HTML attribute `data-amount` to store the amounts (line items amount, total amount). `webform_civcrm_payment.js ` does use` jQuery.data() ` function, which only stores the associated new value in the memory. It does not change the attribute in the DOM when the price is changing, i.e. dynamic contribution amount which is causing the issue when the form is relying on the `data-amount`  attribute 

Changed `data()` function to `attr()` function, addressed this issue. 

See https://api.jquery.com/data/#data-html5

Before
webform_issue

After
webform_issue_fixed

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -57,7 +57,7 @@ cj(function($) {
   function getTotalAmount() {
     var totalAmount = 0.0;
     $('.line-item:visible', '#wf-crm-billing-items').each(function() {
-      totalAmount += parseFloat($(this).data('amount'));
+      totalAmount += parseFloat($(this).attr('data-amount'));
     });
     return totalAmount;
   }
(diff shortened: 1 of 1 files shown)


stesi561:
Thanks heaps for the PR. 

Would it be possible to resubmit this against the Drupal 9 branch -> 6.x - that way we can look at the result of the automated testing, and have more confidence that this change won't break existing functionality. Or is the problem you are referencing only present on the 7.x codebase?

Can you explain a bit better what is breaking in the existing treatment? 

If targeting the 6.x branch - Are you able to write a test that fails - that your changes fixes.

I'm also wondering if either safer to add the attr method instead of replacing, or think about what happens when .data() is used elsewhere. My brief testing suggests that if you set via attr you don't get that value back when accessing via "data". I do see other instances in the code where the first approach is being used for example here:

https://github.com/colemanw/webform_civicrm/blob/b46aca3d4b04df3f996d42965ae079b467b3fff0/js/webform_civicrm_payment.js#L77

PR #761, opened 2022-07-05
Revert "Test change - retrigger tests"

stesi561:
Add back in test for 5.49.x

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,8 @@ jobs:
             php-version: 7.4
             extensions: dom, curl, libxml, mbstring, zip, pdo, mysql, pdo_mysql, bcmath, soap, intl, gd, exif, iconv
             coverage: none
-            tools: composer:v2
+            tools: composer:2.3.8
+
       - name: Get composer cache directory
         id: composercache
         run: echo "::set-output name=dir::$(composer config cache-files-dir)"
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Sigh. Composer just earlier today made a release where it now throws exceptions instead of warning: https://github.com/composer/composer/pull/10920

stesi561:
Aha. I wonder if a better path into this is to specify a composer version that works! 

https://github.com/shivammathur/setup-php suggests we can do this using the following format:

To set up a particular version of a tool, specify it in the form tool:version.

Version can be in the following format:

    Semver. For example tool:1.2.3 or tool:1.2.3-beta1.
    Major version. For example tool:1 or tool:1.x.
    Major and minor version. For example tool:1.2 or tool:1.2.x.

stesi561:
Trying initially with 2.3.8 - however we might want to go for an earlier version so we don't need to load -dev Drupal? However I'm not clear how early we'd need to go in that instance?

demeritcowboy:
Yeah that could work. I'll make a counter-argument though:
* In some ways though it's good to be ahead of the curve and catch these things before the end users start running into them.
* It would end up missing other new composer issues.
* And the plugin issue needs to be solved eventually somehow since eventually 2.3 will be unsupported.

For drupal-9.4.x-dev, looks like there should be a point release tomorrow, so that should solve that.

stesi561:
Agree - your solution is much better!

demeritcowboy:
Thanks!
I'll add 5.49 back in there.

stesi561:
yes sorry about that! - mucked up my remotes was trying to create a new PR to trigger a retest. Not used to having permission on where I've forked from as well as my repo.

PR #762, opened 2022-07-05
Work around composer plugin problem

demeritcowboy:
Overview
Let's see if this works for https://github.com/colemanw/webform_civicrm/pull/761

@stesi561


stesi561:
I like this solution better! :)

PR #763, opened 2022-07-07
Tests only: Increment versions on test matrix

stesi561:
Overview
Increment CiviCRM text matrix following CiviCRM release


stesi561:
I think needs to wait for 5.52.x to drop.

demeritcowboy:
Yes that happens every time there's a delay between release and RC branching.

Also drupal 9.4.2 is now out so you could put 9.4.* and not need -dev for drupal anymore, but there's not too much going on in 9.4 so 50/50.

PR #766, opened 2022-07-17
D8/9 - Fix fatal error if subtype has more than one word

jitendrapurohit:
Overview
Fix fatal error if subtype has more than one word

Before
See issue described at https://github.com/colemanw/webform_civicrm/pull/765

After
Fixed.

Comments
cc @MattTrim1

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -331,7 +331,7 @@ class Utils implements UtilsInterface {
           $contact_types[strtolower($type['name'])] = $type['label'];
           continue;
         }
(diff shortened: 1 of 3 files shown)


MattTrim1:
Hi @jitendrapurohit thanks very much for the alternate solution, which looks cleaner and more widely applicable (ie this will also handle the case where a subtype name is in all caps). 

I also appreciate the addition of the functional test; that's not something I have much experience with in Drupal modules and I'm grateful for the knowledge you've shared.

I can confirm applying this patch has the desired effect.

jitendrapurohit:
Merging based on the above review. Thanks @MattTrim1

PR #769, opened 2022-07-29
In test runs use `git am` not `apply` if applying an optional core patch

demeritcowboy:
Overview
I learned something yesterday about `git apply` vs `git am`. See https://stackoverflow.com/questions/12240154/what-is-the-difference-between-git-am-and-git-apply. If you use apply on a .patch file, it will almost always work, but might end up with the wrong result in some situations. A .patch is the full sequence of commits in the PR, as compared to a .diff which is just the final result (what you see on the changes tab of a PR).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,12 +100,17 @@ jobs:
         run: |
           cd ~/drupal
           COMPOSER_MEMORY_LIMIT=-1 composer require drupal/token:^1.9
+      - name: Set identity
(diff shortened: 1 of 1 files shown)


KarinG:
Where does the mailbox come in? If it's more reliable than git apply - for sure we can use it - but mailbox?

`git-am - Apply a series of patches from a mailbox`

KarinG:
Thanks for explaining this Dave!

PR #770, opened 2022-07-31
D10 - First pass -rector changes for drupal 10 compatibility

jitendrapurohit:
Overview
Rector changes for drupal 10 compatibility

Before
Deprecated code.

After
Contains changes generated by - `vendor/bin/rector process web/modules/contrib/webform_civicrm`

Comments
@KarinG

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -2,9 +2,7 @@
 
 namespace Drupal\webform_civicrm;
 
-use Drupal\webform\WebformInterface;
 use Drupal\Core\Render\Markup;
-use Drupal\webform\Plugin\WebformElementManagerInterface;
 use Drupal\Component\Utility\Html;
 use Drupal\Component\Utility\Xss;
 
@@ -74,7 +72,7 @@ class ContactComponent implements ContactComponentInterface {
   /**
    * Returns a list of contacts based on component settings.
    *
-   * @param WebformInterface $node
+   * @param \Drupal\webform\WebformInterface $node
    *   Node object
    * @param array $element
    *   Webform element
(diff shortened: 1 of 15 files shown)


KarinG:
Awesome 🎉 That's a great first pass.

KarinG:
@demeritcowboy - I'm wondering if it's useful at this point to add D10 to the test matrix [with patches for webform and civicrm]. Or is it too early? (esp. civicrm wise).

demeritcowboy:
One issue was webform (the real webform) won't install inside drupal 10. There's also still the fatal deprecations (jquery.once and `:tabbable`), which I don't know offhand if they're coming directly from webform_civicrm or indirectly from webform.

src/ContactComponent.php
```diff
@@ -71,7 +74,7 @@ function _webform_csv_data_civicrm_contact($component, $export_options, $value)
   /**
    * Returns a list of contacts based on component settings.
    *
-   * @param \Drupal\webform\WebformInterface $node
+   * @param WebformInterface $node
```
demeritcowboy: It's interesting that rector would suggest this since phpcs with the drupal rules says `77 | ERROR   | [x] Data types in @param tags need to be fully namespaced`. None of the files pass linting anyway it just seems odd that rector doesn't use the drupal rules?

KarinG:
> 

@jitendrapurohit -> you initially ran into problems applying rector patch https://www.drupal.org/project/webform/issues/3290604 & https://www.drupal.org/project/webform/issues/3262067 - how did you end up getting webform installed on D10?

src/ContactComponent.php
jitendrapurohit: Yes, seems like a known bug in rector - https://www.drupal.org/project/rector/issues/3269329. A fix was recently merged at https://github.com/palantirnet/drupal-rector/pull/209. I just re-executed the rector with this PR included and have updated the PR now. Thanks @demeritcowboy

PR #771, opened 2022-08-04
Small ad-hoc commit needed to allow negative amounts on the same page.

KarinG:
Necessary to allow negative amounts (like e.g. line items representing discounts) to update if the checkout is on the same webform page as the discount is added [e.g. a one page checkout].

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -92,7 +92,7 @@
         }
         total += isNaN(amount) ? 0 : amount;
       });
-    return total < 0 ? 0 : total;
+    return total;
   }
 
   function calculateLineItemAmount() {


KarinG:
@jitendrapurohit

KarinG:
Going to merge this. It's low risk, because if admins want to constrain a number to only be > 0, they can do so in the config of the element (min / max / step settings).

KarinG:
Ok merging this produced ❌ on ALL versions of CiviCRM so must be related to 
'9.4.x-dev'

@demeritcowboy - why are we testing against 9.4.x-dev? Perhaps we can default back to a 9.4 release? I'm sure there was a reason to address something but I prefer to go back to an actual release. Going to make some yml edits. 

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest::testCurrentEmployer
Behat\Mink\Exception\ElementNotFoundException: Select option with value|text "Default Organization" not found.
```

PR #775, opened 2022-08-08
translate "Existing Contact" element's properties

sluc23:
Overview
Apply `Drupal.t()` function to some properties from **Existing Contact** element, to be able to translate them in the Drupal translation strings

Before
Following **Existing Contact** element are not translated:
- `$element['#attributes']['data-search-prompt']` (used by Autocomplete widget)
- `$element['#attributes']['data-none-prompt']` (used by Autocomplete widget)
- `$element['#search_prompt']` (used by Select widget)
- `$element['#none_prompt']` (used by Selecte widget)

After
Drupal `t()` function is applied to those properties

Technical Details
It's tricky to find exactly where to apply `t()` function for **Existing Contact** element, due to different widgets that could be used, and Drupal / Webform implementation for those elements
Proper parse of string empty or NULL values is important since you cannot send empty or NULL string to `t()` function, to avoid Exceptions.

### DIFF
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -130,8 +130,17 @@ class CivicrmContact extends WebformElementBase {
       $element['#options'] = [];
       $element['#attributes']['data-is-select'] = 1;
     }
-    $element['#attributes']['data-search-prompt'] = $this->getElementProperty($element, 'search_prompt') ?? '- Choose existing -';
-    $element['#attributes']['data-none-prompt'] = $this->getElementProperty($element, 'none_prompt') ?? '+ Create new +';
+    // get translated properties used for autocomplete widget
+    $element['#attributes']['data-search-prompt'] = $this->getTranslatedElementProperty($element, 'search_prompt') ?? t('- Choose existing -');
+    $element['#attributes']['data-none-prompt'] = $this->getTranslatedElementProperty($element, 'none_prompt') ?? t('+ Create new +');
+    // translate search_prompt, which later is copied as #empty_value, and then created as first <option> if element is select widget
+    if(!empty($element['#search_prompt'])){
+      $element['#search_prompt'] = t($element['#search_prompt']);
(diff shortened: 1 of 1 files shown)


KarinG:
@sluc23 - thank you! We've run into a recent regression in what looks like webform 6.x-dev so let me wait until we've resolved that and then approve your PR for a test run.

src/Plugin/WebformElement/CivicrmContact.php
```diff
@@ -643,4 +652,26 @@ public static function validateRequired(&$element, FormStateInterface &$form_sta
...
+   *   property does not exist.
+   */
+  public function getTranslatedElementProperty(array $element, $property_name) {
+    $prop = $this->getElementProperty($element, $property_name);
+    if (!empty($prop) && is_string($prop)) {
+      return t($prop);
+    }
+    else {
+      return $prop;
+    }
```
jitendrapurohit: @sluc23 Can we eliminate the usage of `else` here? and simply `return $prop;`?

sluc23: @jitendrapurohit the problem is that `t($value)` function returns an Exception if the `$value` is empty. `t()` it's intended to be used with literal strings, not variables.. so we need to be sure the variable has no empty string.

sluc23: re-thinking your question, the answer is yes, it could be eliminated the else and return `$prop`.. do you want me to rework it?

KarinG:
Hi @sluc23 - yes please rework.

src/Plugin/WebformElement/CivicrmContact.php
sluc23: reworked and rebased to latest master commit @KarinG @jitendrapurohit

KarinG:
Great and all green on 

image

KarinG:
Merging this now 🎉

I would love your thoughts though on how to get some translation testing into our Matrix.

PR #784, opened 2022-08-10
[do not merge] drupal 10 test runs

demeritcowboy:
Overview
One issue is the parent webform doesn't run in drupal 10, and among other things I had to remove the dependence on webform from composer.json. This is just to get to the point where tests actually start running.

Because merging this would break stuff, to see the test run output you can see it in my fork: https://github.com/demeritcowboy/webform_civicrm/runs/7766348267

### DIFF
--- a/.github/workflows/d10.yml
+++ b/.github/workflows/d10.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         include:
           - drupal: '10.0.x-dev'
-            civicrm: '5.58.alpha1'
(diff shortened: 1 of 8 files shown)


KarinG:
🤞 I like experiments!

demeritcowboy:
I don't need this PR open - it was mostly to track progress and I thought I did need a PR to trigger test runs but (a) I don't, and (b) it runs the wrong yml file so I have to keep cancelling it. I need to keep the branch though.

PR #785, opened 2022-08-14
D8/9 - Fix source field to display on the contact widget using v4 api

jitendrapurohit:
Overview
Resubmission of https://github.com/colemanw/webform_civicrm/pull/742

Fix source field to display on the contact widget using v4 api

Before
Source field fails to load on the contact widget. To replicate -

- Add Existing contact element to the webform.
- Update widget to Autocomplete. Select Source in the `Contact Display Field(s)` input.
- Load the webform and search for a contact.
- No source is displayed

image

After
Source value is displayed -

image

Technical Details
It looks like a core bug in civicrm api v3. It does not return the source value when contact get api is called. But we do get the source using v4. This PR uses the v4 api instead and displays the values on the contact widget. 

Comments
@KarinG

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -93,23 +93,46 @@ class ContactComponent implements ContactComponentInterface {
     $limit = $str ? 12 : 500;
     $ret = [];
     $display_fields = array_values($element['#results_display']);
-    $search_field = 'display_name';
+    $fieldMappings = [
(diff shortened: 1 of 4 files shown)


KarinG:
Thank you @jitendrapurohit - Test fails on a curl timeout, so that's not an issue. I have this live working on a site. Merging!

PR #788, opened 2022-08-18
Add sorting when finding memberships.

puresyntax71:
Overview
If you have a multiple memberships with the same status for a contact, the displayed membership when viewing the webform can be unpredictable (or perhaps it's being loaded by ID in descending order). An example is when you have a membership 1 and 2 wherein membership 1 has the latest end date instead of membership 2. This can happen if membership 1 was overridden at some point.

The PR loads the membership in descending order by "end_date" to have the membership loaded with the latest end date.

Before
Wrong membership is being displayed on the page.

After
Membership with the latest end date is loaded.

### DIFF
--- a/tests/src/FunctionalJavascript/MembershipSubmissionTest.php
+++ b/tests/src/FunctionalJavascript/MembershipSubmissionTest.php
@@ -135,7 +135,7 @@ final class MembershipSubmissionTest extends WebformCivicrmTestBase {
 
     // Create two memberships with the same status with the first membership
     // having an end date after the second membership's end date.
-    \Drupal::service('webform_civicrm.utils')->wf_civicrm_api('membership', 'create', [
+    $this->utils->wf_civicrm_api('membership', 'create', [
       'membership_type_id' => 'Basic',
       'contact_id' => 2,
       'join_date' => '08/10/21',
@@ -145,7 +145,7 @@ final class MembershipSubmissionTest extends WebformCivicrmTestBase {
       'status_id' => 'Expired',
     ]);
 
-    \Drupal::service('webform_civicrm.utils')->wf_civicrm_api('membership', 'create', [
+    $this->utils->wf_civicrm_api('membership', 'create', [
       'membership_type_id' => 'Basic',
       'contact_id' => 2,
       'join_date' => '01/01/21',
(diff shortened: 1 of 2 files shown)


KarinG:
That one error -> Github Actions was having issues earlier today! Adding to our list to test/review. Thank you!

tests/src/FunctionalJavascript/MembershipSubmissionTest.php
```diff
@@ -133,6 +133,31 @@ public function testSubmitWebform() {
     $this->getSession()->getPage()->pressButton('Save Settings');
     $this->assertSession()->pageTextContains('Saved CiviCRM settings');
 
+    // Create two memberships with the same status with the first membership
+    // having an end date after the second membership's end date.
+    \Drupal::service('webform_civicrm.utils')->wf_civicrm_api('membership', 'create', [
```
jitendrapurohit: @puresyntax71 Can we use `$this->utils` instead of `\Drupal::service('webform_civicrm.utils')`?

puresyntax71: I have updated the test.

tests/src/FunctionalJavascript/MembershipSubmissionTest.php
```diff
@@ -133,6 +133,31 @@ public function testSubmitWebform() {
...
+      'contact_id' => 2,
+      'join_date' => '01/01/21',
+      'start_date' => '01/01/21',
+      'end_date' => '01/01/22',
+      'is_override' => 1,
+      'status_id' => 'Expired',
+    ]);
+
+    $this->drupalGet($this->webform->toUrl('canonical'));
+    $this->assertSession()->statusMessageContains('Basic membership for ' . mb_strtolower($this->adminUser->getEmail()) . ' has a status of "Expired". Expired ' . \CRM_Utils_Date::customFormat('2022-08-10'));
```
jitendrapurohit: Test fail seems to be related to this function `statusMessageContains()`? It was introduced in [9.4.x version](https://www.drupal.org/node/3270424). @KarinG Do we still need 9.3.x in the test matrix?

KarinG: Yes - we do still need 9.3.x - it's a supported Drupal version so sites are using it. 9.3.21 was released Aug 03 (together with 9.4.5).

KarinG:
Thank you! 👍

PR #789, opened 2022-08-23
Suppress htmloutput links in actions run output

demeritcowboy:
Overview
There's soooo maaaaanny.

Before
Massive scrolling whenever you open an actions run log. Useless links to screenshots that give 404s.

After
No links displayed. Can still download the artifacts file if desired.

Technical Details
Could maybe also subclass the trait, but this seems easier. It's not meant to be a part of webform itself.

Comments
The links are sometimes useful locally if only running one or two tests, but they're not useful here.


KarinG:
😂

demeritcowboy:
Ooh that worked.

PR #790, opened 2022-08-29
Call to undefined function wf_crm_field_options() in webform_civicrm_webform_submission_render_alter()

mattwire:
Overview
See https://www.drupal.org/project/webform_civicrm/issues/3177210

This fix is required when "existing contact" is not selected for first contact.

Before
PHP error undefined function.

After
Submission renders.

Technical Details

Comments

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -331,6 +331,8 @@ function webform_civicrm_webform_submission_render_alter(&$renderable) {
     return;
   }
   civicrm_initialize();
+  // See https://www.drupal.org/project/webform_civicrm/issues/3177210
+  module_load_include('inc', 'webform_civicrm', 'includes/utils');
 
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you. We have many forms in use where Existing Contact is not selected for the first contact but I don't recall seeing a PHP error. Can you give me exact steps to reproduce?

KarinG:
According to the issue description:

2.0 Detailed steps to reproduce (embed screenshots)
Create a simple Drupal webform without any CiviCRM components.
Leave "Enable CiviCRM Processing" unchecked on the Webform CiviCRM tab.
Create a new submission.
View the submission at /node/[nid]/submission/[sid]

I just did this on a live site ->

image

Not getting any error (also not seeing a notice) in the dblog.

PR #794, opened 2022-09-04
Update paylater test

demeritcowboy:
Overview
Some recent fails because a couple wrongs used to make a right.

Before
Short version: For the second email on completetransaction, it was being sent out by php mail not redirect-to-database, so when the test checked the "most recent email" in the db it was checking the first email again not the second.

After
A sequence of changes in core means the email is now being sent to redirect-to-database. The email from completetransaction is not the same email as the first one.

Technical Details
Short version: Somewhere in here it "fixes" something about how the mailer picks which mailer in the situation in the test: https://github.com/civicrm/civicrm-core/compare/22a3e45ba2..d1f11af

It doesn't seem like it should but git bisect says so. It's difficult to pin down the exact commit because some of the intermediate commits introduce other bugs which are fixed in later commits that prevent getting far enough in the test to tell.

Comments

### DIFF
--- a/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
+++ b/tests/src/FunctionalJavascript/ContributionPayLaterTest.php
@@ -16,6 +16,7 @@ final class ContributionPayLaterTest extends WebformCivicrmTestBase {
   public function testReceiptParams() {
     $this->drupalLogin($this->rootUser);
     $this->redirectEmailsToDB();
+    $this->setOrgInfo();
 
     $this->drupalGet(Url::fromRoute('entity.webform.civicrm', [
       'webform' => $this->webform->id(),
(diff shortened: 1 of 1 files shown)


tests/src/FunctionalJavascript/ContributionPayLaterTest.php
```diff
@@ -74,15 +75,19 @@ public function testReceiptParams() {
     $this->assertStringContainsString('Payment by Direct Credit to: ABC. Please quote invoice number and name.', $sent_email);
     $this->assertStringContainsString('Thank you for your contribution', $sent_email);
 
+    // "Hack" for versions before 5.54 - it doesn't pick up the setting we set earlier.
+    $mb = \Civi::settings()->get('mailing_backend');
```
KarinG: Do we need to wrap this in a version check then?

demeritcowboy: makes sense

KarinG:
Awesome - thank you so much Dave for getting to the bottom of this one!

PR #796, opened 2022-09-18
D9 - Avoid sending recur fields for one time payments

jitendrapurohit:
Overview
Related to https://www.drupal.org/project/webform_civicrm/issues/3231305

Before
image

Error if frequency unit is set to 0 on the form.

After
Submits the webform without any error and a normal one time payment is processed.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -1927,6 +1927,8 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       $result = $this->contributionRecur($contributionParams);
     }
     else {
+      // Not a recur payment, remove any recur fields if present.
+      unset($contributionParams['frequency_unit'], $contributionParams['frequency_interval'], $contributionParams['is_recur']);
       $result = $this->utils->wf_civicrm_api('contribution', 'transact', $contributionParams);
     }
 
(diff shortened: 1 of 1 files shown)


KarinG:
This looks good - thanks Jitendra. I have asked Dave to look into why there are issues with CiviCRM dev-master - we noticed this a few days ago w/ CDN TaxReceipts extension as well.

jitendrapurohit:
Tested by [masetto](https://www.drupal.org/u/masetto) at https://www.drupal.org/project/webform_civicrm/issues/3231305#comment-14705665

src/WebformCivicrmPostProcess.php
```diff
@@ -2083,6 +2085,10 @@ private function submitIPNPayment() {
     $params['total_amount'] = round($this->totalContribution, 2);
     // Some processors want this one way, some want it the other
     $params['amount'] = $params['total_amount'];
+    $frequencyUnit = wf_crm_aval($params, 'frequency_unit');
+    if (empty($frequencyUnit)) {
+      unset($params['frequency_unit'], $params['frequency_interval'], $params['is_recur']);
```
KarinG, line 2091: @jitendrapurohit -> in the LIVE payment section, we use this to determine if it's a recurring contribution:

`if ($numInstallments != 1 && !empty($frequencyInterval)) {`

I think it's best to have the same IF in the IPN section? So use `$params['frequency_interval']` instead of `$frequencyUnit`. Also I'm not even sure if recurring works for IPN Payment Processors.

jitendrapurohit: Sounds good to me. Have updated. Thanks @KarinG

PR #798, opened 2022-09-22
New contact not attached to contact reference custom field

pradpnayak:
Overview
Contact reference custom field not saved for new creation

Steps to replicate:

1. Create custom field of type contact reference
2. include it in a webform

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2822,7 +2822,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     $updateParams = [
       'id' => $cid,
     ];
-    $skipKeys = [];
+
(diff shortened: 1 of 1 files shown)


KarinG:
Can you please generalize the comment (re: `custom_18`)?

jitendrapurohit:
I think i replicated a problem with contact reference field, but not sure if it matches your usecase. Can you confirm pls @pradpnayak 

- Create a custom group with multivalue = no.
- Add a contact custom ref field to the group.
- Create a webform and include that field as `User-select`  

image

- Visit the webform as anonymous user and enter the following.

image

**Expected Result** - A new contact is created with custom field value stored as the same cid.

**Actual Result** - A new contact is created in civicrm with custom field value set to cid=1.

Looks like it sets the option id instead of the new contact id created during the webform submission.

**After this patch** - it is still unresolved. So not sure if this is the exact problem you were trying to solve @pradpnayak ?

PR #804, opened 2022-10-13
Fix authnet payment failed.

puresyntax71:
Overview
When using contributions with authorize.net, webform keeps on failing submission. On later versions of CiviCRM, the following error pop ups:

> It appears that this transaction is a duplicate. Have you already submitted the form once? If so there may have been a connection problem. Check your email for a receipt from Authorize.net. If you do not receive a receipt within 2 hours you can try your transaction again. If you continue to have problems please contact the site administrator.

On lower versions of CiviCRM, it only says "Payment failed".

Before
Error when using contributions with auth.net.

After
No more error.

Technical Details
The error seem to have started on webform_civicrm 6.1.0.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -736,7 +736,7 @@ class Utils implements UtilsInterface {
     $order = civicrm_api3('Order', 'create', $params);
     try {
       $params['amount'] = $params['total_amount'];
-      $params['contribution_id'] = $order['id'];
+      $params['contribution_id'] = $params['contributionID'] = $order['id'];
       $payParams = $params;
       $payResult = reset(civicrm_api3('PaymentProcessor', 'pay', $payParams)['values']);
 
--- a/tests/src/FunctionalJavascript/AuthorizeNetTest.php
+++ b/tests/src/FunctionalJavascript/AuthorizeNetTest.php
@@ -1,0 +1,168 @@
+<?php
+
+namespace Drupal\Tests\webform_civicrm\FunctionalJavascript;
+
+use Drupal\Core\Url;
+
+/**
+ * Tests submitting a Webform with CiviCRM: Contribution using AuthorizeNet.
+ *
+ * @group webform_civicrm
+ */
+final class AuthorizeNetTest extends WebformCivicrmTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    $result = $this->utils->wf_civicrm_api('PaymentProcessor', 'create', [
+      'payment_processor_type_id' => 'AuthNet',
+      'financial_account_id' => 'Payment Processor Account',
+      'name' => 'Credit Card',
+      // 'user_name' => 'xxxx',
+      // 'password' => 'xxxx',
+      'url_site' => 'https://test.authorize.net/gateway/transact.dll',
+      'accepted_credit_cards' => [
+        'Visa' => 'Visa',
+        'Mastercard' => 'Mastercard',
+        'Amex' => 'Amex',
+        'Discover' => 'Discover',
+      ],
+    ]);
+
+    $this->paymentProcessorID = $result['id'];
+
+    drupal_flush_all_caches();
+  }
+
+  /**
+   * Test webform submission using AuthorizeNet processor.
+   */
+  public function testSubmitContribution() {
+    $this->drupalLogin($this->adminUser);
+    $this->drupalGet(Url::fromRoute('entity.webform.civicrm', [
+      'webform' => $this->webform->id(),
+    ]));
+    $this->setUpSettings();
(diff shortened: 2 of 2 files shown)


KarinG:
Can you please clarify what you mean by 
"The error seem to have started on webform_civicrm 6.1.0"? Any idea which PR?

If this is ensuring webform civicrm works with Authorize.net then that's great - but it will require a test to ensure param confusion gets identified quickly in case this happens again.

KarinG:
Also which authorize.net is this? Core or extension -and why are there two authorize.net options? @mattwire - I'm sure you know the answer to this one! :-)

puresyntax71:
Yeah, seems to have started from this PR https://github.com/colemanw/webform_civicrm/pull/698 although I haven't checked on why since I just went through the merge commits between 6.1.0 and 6.2.1.

> Also which authorize.net is this? 

This is the core one. I tested it with vanilla Drupal 9.4.5 with webform_civicrm 6.2.1 and CiviCRM 5.53.

Yeah, definiitely need tests.

puresyntax71:
I have added a test that is a stripped out version of the StripeTest (no verification of contribution). I'm not entirely sure how to set the [username and password though](https://github.com/colemanw/webform_civicrm/pull/804/files#diff-e4f9e43cb9d695648266a456114979b5d70c09c89969ab952e7d9a75cdf94d82R24-R25).

In, gitlab, we normally used environment variables in the CI/CD although I'm not sure how it works on github.

KarinG:
Great - thank you!

authorize.net must have test accounts that you can use with fake credit cards - that's what you'll want to use for:

```
      // 'user_name' => 'xxxx',
      // 'password' => 'xxxx',

```

Also, I do think we need a // Load the Contribution and check the values to evaluate if Status = Completed, if the Amount is correct, etc.

puresyntax71:
> authorize.net must have test accounts that you can use with fake credit cards - that's what you'll want to use for

Yeah, not sure if there exists one although I can't find any apart from creating sandbox accounts.

I went ahead and refactored the tests. I'm not sure if it's a bit too much change though. I just copied StripeTests to PaymentProcessorBaseTest and used that as a base class for both Stripe and AuthorizeNet to avoid duplicates.

From https://www.drupal.org/project/webform_civicrm/issues/3312157#comment-14747198:

>  Such tests already exist for iATS Payments, and also for Stripe.

Oh, I didn't realize there were already tests for other payment processors. I thought it was only Stripe, Pay Later, and Dummy.

KarinG:
Hi - I strongly prefer just a new test for authorize.net [that does not touch/modify any other test] so that we can be 100% sure there is no breakage. In future work we'll consolidate and stream line code [remove duplications]. 

However if authorize.net does not have any standard sandbox we can use then a functional end-to-end test is not possible. In these tests we really need to hear back from the payment processor. 

@mattwire - do you happen to have a set of test credentials we can use for authorize.net?

puresyntax71:
Yeah, I've updated the test with verification based on StripeTest.

src/Utils.php
```diff
@@ -736,7 +736,7 @@ function wf_civicrm_api3_contribution_transact($params) {
     $order = civicrm_api3('Order', 'create', $params);
     try {
       $params['amount'] = $params['total_amount'];
-      $params['contribution_id'] = $order['id'];
+      $params['contribution_id'] = $params['contributionID'] = $order['id'];
```
mattwire, line 739: Why is this change needed?

KarinG:
```
Error message Notice: Undefined index: user_name in CRM_Core_Payment_AuthorizeNet->__construct() (line 68 of /home/runner/drupal/vendor/civicrm/civicrm-core/CRM/Core/Payment/AuthorizeNet.php). 
```

Without dedicated Auth.net test credentials we can't do end to end testing. Does Skvare have dedicated Sandbox credentials you can put in? If not - then I'm happy to merge just the one-line param variation fix - but of course there would be no future guarantees that Core Auth.net works with WFC.

src/Utils.php
mattwire: Since CiviCRM 5.39 this should not be required because of https://github.com/civicrm/civicrm-core/commit/5d8d2f382f3f6662c1e789c76a3d5033ce01c9d6 and core authnet should automatically convert `contribution_id` to `contributionID` internally.

puresyntax71: Yeah, I'm not entirely sure but this is where it's happening - https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Payment/AuthorizeNet.php#L170.

mattwire: ok, so the fix should be in CRM/Core/Payment/AuthorizeNet.php and not here. Change the $params['contributionID'] to $propertyBag['contributionID']

PR #805, opened 2022-10-16
Add CRM resources more often - specifically to every webform that has a webform_civicrm handler

KarinG:
Overview
We discovered that if you attach a webform to a node (e.g. with an entity reference, such as paragraphs) that there are moments where CiviCRM CRM JS are not loaded. This only happened for anonymous and was fixed consistently but only temporarily with a cache clear.

Technical Details
Add CRM resources more often - specifically to every webform that has a webform_civicrm handler

Symptom -> the tally at times does not work for anonymous users ->
This may well be specific to one page checkout forms (contact paging disabled)
Even if the tally does not work on the webform that is attached to the node, it always does work on the /form itself

image

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -86,11 +86,13 @@ function webform_civicrm_datetime_set_format($element) {
  * Implements hook_form_alter().
  */
 function webform_civicrm_form_alter(&$form, &$form_state, $form_id) {
-  if ($form_id == 'webform_edit_form' && $form_state->getFormObject()->getEntity()->getHandlers()->has('webform_civicrm')) {
-    \Drupal::service('civicrm')->initialize();
-    \CRM_Core_Resources::singleton()->addCoreResources();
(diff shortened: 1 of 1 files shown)


KarinG:
@jitendrapurohit - this is the PR I mentioned to you earlier today.

KarinG:
So this did not solve the problem. Even with this branch in place we saw a resurfacing of the JS error. CRM undefined...

So to summarize:

- /form itself - always works fine
- when you add that webform to a node (either using paragraphs or using the field -> Webform) - you end up with the JS error after some cache rebuild condition (not sure what that is).
- but only if that webform is a one page checkout (so Contact paging disabled).

jitendrapurohit:
This looks like a very safe fix. Is it worth adding a static var so `addCoreResources()` is not called more than once?

KarinG:
So the problem is that this PR did **_-not-_** solve the problem. Live testing this -> the CRM resources still intermittently go missing.

PR #806, opened 2022-10-30
github actions deprecations

demeritcowboy:
Overview
nodejs v12 and set-output are deprecated for github actions

Before
On any actions run, just above where you see the artifacts, there's a bunch of repeated warnings:
* Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/cache, actions/upload-artifact
* The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

After

Technical Details

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
           - 3306
         options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


KarinG:
Ok this is cool stuff and test only. Thank you for keeping us up to date Dave!

PR #809, opened 2022-11-27
D9 - Fix ipn processor payment with billing fields

jitendrapurohit:
Overview
Webform with billing fields + ipn processor (eg eway) fails for anonymous users.

Before
Webform does not submit for anonymous user with billing fields + ipn processor. To replicate

- Create a webform with contribution and Processor = Eway.
- Enable billing fields.
- Submit the form.
- User is NOT redirected to the eway page for payment. Instead it redirects you to the webform front page. 

After
Fixed.

Technical Details
Include billing fields expected by the payment processor. We do this for the live payment processor, but is missing for ipn submissions.


petednz:
Can confirm that on the staging site with this patch installed I was able to reach the eway interface as anonymous user

PR #813, opened 2022-12-03
D9 - Replace state abbreviation with id on normal address state field

jitendrapurohit:
Overview
State dropdown has odd behaviour based on if its included in billing or normal address fields. Fix it to have consistent behaviour.

Before
State dropdown in normal address fields has abbreviation in values.
 

whereas billing populated ids in the select values.

After
Consistent behaviour between billing and normal address state field - ID is populated in values for both the fields.

Comments
@KarinG @MegaphoneJon You might be interested in reviewing this?


KarinG:
Awesome and thank you 👍

PR #815, opened 2022-12-04
D8/9 - Add same as checkbox for billing address

jitendrapurohit:
Overview
Add `Same As` checkbox to copy values from normal address fields to billing fields.

Before
If both normal & Billing address fields are enabled - user has to enter the address & name information twice to submit the form. There's no way to copy the field values to the billing fields.

After
- A new checkbox `Same As` is added under billing section.

image

- Enable this to copy values from Contact 1 address fields to billing fields.

- Displays an alert message is Address on Contact 1 is not enabled.

image

- On the webform - enable this checkbox to copy the value from Contact 1 Address

image

Comments
@KarinG


KarinG:
Testing this!

Contact 1 ->  on page 1 (this is all good) ->
image

Onto page 2 -> on load ->
image

When checking the box 'Same as" -> unfortunately both Country and State/Province -> go blank

image

Any idea as to what may be happening (or not) here?

jitendrapurohit:
Sorry to take this long in updating this PR, I think i overlooked your comment @KarinG.

From the screenshot, it seems you do not have country field enabled on the first page so the code tries to copy the empty value to the billing country. 

Have fixed this so the "Same as" checkbox only copies the existing field values from normal -> billing address fields. Looks fine now?

KarinG:
Ok this is now live on a new D9 site we're working on 👍

PR #816, opened 2022-12-10
D8/9 - Custom fields for participant role not available at page load

jitendrapurohit:
Overview
Custom fields specific to a certain participant role are only available for selection after selecting a different participant role.

Before
Given an existing webform with one contact that has event registration enabled with Participant Role Attendee. Now if I navigate to the CiviCRM settings of this webform and want to select additional fields that are specific to the Attendee role, I cannot see them. Only if I select - User Select - first and then Attendee again they become available dynamically.

After
Custom set is loaded on page load.

Technical Details
The js to trigger the change event was not called on page load.

Comments
Drupal Ticket - https://www.drupal.org/project/webform_civicrm/issues/3323528

### DIFF
--- a/js/webform_civicrm_admin.js
+++ b/js/webform_civicrm_admin.js
@@ -88,6 +88,17 @@ var wfCiviAdmin = (function (D, $, once) {
   };
 
   /**
+   * Load custom fields related to participant fields.
+   */
+  $('.cf-participant-fields').each(function() {
+    $(this).on("change", function(event) {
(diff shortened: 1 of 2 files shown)


KarinG:
We have an issue with the 1stpay cryptogram widget - (which is the test that fails).
[fkohrt] has confirmed the patch resolves the issue, so I'm merging this and will make a note to circle back to the D9.4 vs cryptogram issue later this weekend.

PR #817, opened 2022-12-12
fix deprecated function warnings on wf_crm_get_relationship_types

MegaphoneJon:
Overview
When loading a webform_civicrm page, I get a lot of "Deprecated function" warnings in PHP 8 because I'm passing `NULL` to `strlower()`.  You can replicate this by creating a new relationship type which has "All Contacts" as its contact type (which is stored in the db as `NULL`).

Before
Watchdog noise.

After
Less noise.

Technical Details
`wf_crm_aval` lets us specify the default value, so we can just return an empty string instead of `NULL`.

Comments
There was already handling for this for contact subtypes but this is more concise, so I also updated the subtype code.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -361,16 +361,10 @@ class Utils implements UtilsInterface {
     static $types = [];
     if (!$types) {
       foreach ($this->wf_crm_apivalues('relationship_type', 'get', ['is_active' => 1]) as $r) {
-        $r['type_a'] = strtolower(wf_crm_aval($r, 'contact_type_a'));
-        $r['type_b'] = strtolower(wf_crm_aval($r, 'contact_type_b'));
-        $r['sub_type_a'] = wf_crm_aval($r, 'contact_sub_type_a');
-        if (!is_null($r['sub_type_a'])) {
-          $r['sub_type_a'] = $r['sub_type_a'];
-        }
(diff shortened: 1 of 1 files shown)


KarinG:
Thanks Jon. There is some noise from a failing iATS test on D9, but this one is real:

```
Testing modules/contrib/webform_civicrm
.......E....................................................S.    62 / 62 (100%)

Time: 1.09 hours, Memory: 300.00 MB

There was 1 error:

1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContactRelationshipTest::testSubmitWebform
Behat\Mink\Exception\ElementNotFoundException: Select option with value|text "School is" not found.

/home/runner/drupal/vendor/behat/mink/src/Element/NodeElement.php:234
/home/runner/drupal/vendor/behat/mink/src/Element/TraversableElement.php:262
/home/runner/work/webform_civicrm/webform_civicrm/tests/src/FunctionalJavascript/ContactRelationshipTest.php:194
/home/runner/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:703
```

MegaphoneJon:
Thanks for interpreting that for me @KarinG - I wasn't sure how many of those to pay attention to.  I'll take a look at this.

PR #823, opened 2023-01-03
Themes used in unit test are deprecated

demeritcowboy:
Overview
I'm cherry-picking a few commits from my drupal 10 WIP. These themes don't exist in drupal 10 but their replacements do exist in 9.x so should be ok to do this now.


KarinG:
Righto 🍾

PR #824, opened 2023-01-03
missing `use` statement

demeritcowboy:
Overview
In 9.x ckeditor is always present, so the error never triggers. In drupal 10 the error tries to happen on line 678, but it can't find the class because the `use` statement is missing.


KarinG:
Awesome 🎉

PR #825, opened 2023-01-03
test improvement - avoid having to hunt in screenshots for the error

demeritcowboy:
Overview
When there's a fatal error and the browser goes to the stack trace page, the actions log just says "element missing" or something when the test tries to execute the next statement. Then you have to download all 100MB of screenshots and guess which screenshot contains the stack trace. This outputs the trace to the actions log directly.


KarinG:
Yeah - I often just end up adding an actual screenshot -> (faster than sorting through the HTML) :-)

`$this->createScreenshot($this->htmlOutputDirectory . '/legacy289.png');`

demeritcowboy:
Thanks Karin!

PR #826, opened 2023-01-03
test improvement - store civi logs in artifacts

demeritcowboy:
Overview
The ConfigAndLog folder is inaccessible after a test finishes. This copies any log files that were created during the test into the artifacts folder so that they get included in the downloadable artifacts. Can be useful sometimes. Or you can use it directly in a test you're debugging by sprinkling in some `\Civi::log()->info('something');` and then looking at it after.


KarinG:
Ah cool 🥳

PR #828, opened 2023-01-05
Update test matrix - Jan 2023

KarinG:
Post the holiday security update.


KarinG:
Will merge this as I'm not expecting any troubles. Will get a notification if it does fail.

PR #833, opened 2023-01-26
Improve speed of getting Civi version

MegaphoneJon:
Overview
A speed boost that loads pages anywhere from 280ms to 700ms faster.

Before
Slower page load.

After
Faster page load.

Technical Details
The `System.get` API does a lot more than just calculate the version - enough so that when I ran XDebug profiling on a page submission, the API call stood out as the lowest-hanging fruit.

You can test this by running these on your command line and observing the time on the last line:
```shell
time cv api System.get 
time cv ev '$t = CRM_Utils_System::version(); var_dump($t)'
```

Comments
On a pretty bare site, the speed difference is 280-300ms.  On a heavy site, I've sped this up by as much as 700ms.  This function runs on every webform page load and submission.

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -68,7 +68,8 @@ class Fields implements FieldsInterface {
         'activity' => ['entity_type' => 'activity', 'label' => t('Activity'), 'max_instances' => 99,  'attachments' => TRUE],
         'relationship' => ['entity_type' => 'contact', 'label' => t('Relationship'), 'help_text' => TRUE, 'custom_fields' => 'combined'],
       ];
-      $civicrm_version = $this->utils->wf_crm_apivalues('System', 'get')[0]['version'];
+      
+      $civicrm_version = \CRM_Utils_System::version();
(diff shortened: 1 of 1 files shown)


KarinG:
Jon: some time ago I spent a lot of time replacing CiviCRM function calls with API calls as (Coleman told me to do so :-)) - but I clearly see your point re: speed. Dave commented: "CRM_Utils is not technically a public api, but that distinction is a bit blurry in civi" - so I'll rerun the one failed test (CURL timeout error) - and then will merge 👍

PS - Dave: we're getting timeout errors re: curl - I think this was merged to help prevent download timeouts from civicrm.org :-)

2) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionIatsTest::testSubmitContribution
CiviCRM_API3_Exception: cURL error 28: Operation timed out after 5000 milliseconds with 21062 out of 331765 bytes received (see https://curl.haxx.se/libcurl/c/libcurl-errors.html)

PR #839, opened 2023-02-04
D9 - Fix fatal error on single static option submission for Activity Type …

jitendrapurohit:
…field

Overview
Fix fatal error on single static option submission for Activity Type field.

Before
Fatal Error if Activity Type field has a single option enabled. To replicate -

- Create a webform with Activity Type = User select
- Edit this element and enable single static option 
- Submit the webform. Error

image

After
Fixed.

Comments
Drupal Ticket - https://www.drupal.org/project/webform_civicrm/issues/3336056

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Webform CiviCRM
 ===============
 
-Webform CiviCRM is a powerful, flexible, user-friendly form builder for CiviCRM.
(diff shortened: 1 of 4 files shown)


mikantchap:
Is this actually in 6.2.5 release? Thanks

MegaphoneJon:
@mikantchap Yes.

PR #841, opened 2023-02-09
Fix submitting state/province with IDs over 10,000

MegaphoneJon:
Overview
See https://www.drupal.org/project/webform_civicrm/issues/3340681.

Before
Crash when submitting an address with a state/province with id >= 10,000.

After
No crash

Technical Details
#813 replaced the state abbreviation on this field with an ID.  The code I'm removing is trying to convert it from abbreviation to ID.

Comments
This is the only use of `wf_crm_state_abbr` in the codebase.  Is there any reason not to delete the entire function and associated tests?  If not, I'll do so in a follow-up PR.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -684,10 +684,6 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
           }
         }
       }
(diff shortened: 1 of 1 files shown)


jitendrapurohit:
I concur with the removal of `wf_crm_state_abbr()`. Thanks for the PR @MegaphoneJon. I've tested and confirmed that it fixes the problem with state id > 10,000. Failing test is not related to this change. Merging.

PR #842, opened 2023-02-10
use form's currency, not system default, for displaying line items/total

MegaphoneJon:
Overview
See https://www.drupal.org/project/webform_civicrm/issues/3340955.

Before
The line items and total on the payment page display in the system default currency.

After
The line items and total on the payment page display in the webform's currency.

Comments
The PHP fix is the primary fix.  The JS fix works because the PHP fix is providing the correct format.

### DIFF
--- a/js/webform_civicrm_payment.js
+++ b/js/webform_civicrm_payment.js
@@ -54,8 +54,33 @@
   function tally() {
     var total = 0;
     total = getTotalAmount();
-    $('#wf-crm-billing-total').find('td+td').html(CRM.formatMoney(total));
+    // Match the existing format for the total.
+    var moneyFormat = getCurrencyFormat($('#wf-crm-billing-total').find('td+td').html());
+    $('#wf-crm-billing-total').find('td+td').html(CRM.formatMoney(total, false, moneyFormat));
     $('#billing-payment-block').toggle(total > 0);
+  }
+
+  // Given an existing amount, generate a valid format for CRM.formatMoney.
+  function getCurrencyFormat(currentAmount) {
+    if (!currentAmount) {
+      return null;
+    }
+    var moneyFormat = null;
+    var noSymbolsFormat = null;
+    // comma or decimal point followed by exactly two digits optionally followed by non-digits.
+    var decimalSeparator = (currentAmount.match(/([\.|,])\d{2}\D*$/) ?? [])[1] ?? '';
+    // comma, period, space, apostrophe or underscore followed by 3 digits.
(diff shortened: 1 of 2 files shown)


jitendrapurohit:
@MegaphoneJon The test failure seems to be related. There seems to be js error when the default currency is selected on the Contribution tab.

image


The console error is probably due to the usage of uppercase FALSE in js. Should be fixed by replacing it to `false`.

KarinG:
Hi @MegaphoneJon - that looks like a real failure:

```
5) Drupal\Tests\webform_civicrm\FunctionalJavascript\ContributionDummyTest::testAssignContributionSecondContactSelectByUserPaymentProcessor
Behat\Mink\Exception\ElementTextException: The text "10.00" was not found in the text of the element matching css "#wf-crm-billing-total".
```

MegaphoneJon:
A question (unrelated to the currently failing test):

I've written a new JavaScript function that really needs unit testing with a large number of inputs.  FunctionalJavascript seems way too heavy, since I want to do a pure unit test.  Is there existing support for JS unit tests?

The values I want to test:
```
€275.00 EUR
$4,000.00 EUR
€275 EUR
$4,000.00
$4,000,000.00
$4 000 000.00
275€ EUR
275.00€ EUR
1'275.00€ EUR
$4_000_000.00
$1
$1.00
£9,500
$2.000,75
```

KarinG:
Hi Jon -> if I understand your questions correctly -> yes we support non Functional JS tests -> 

image

MegaphoneJon:
test this please

KarinG:
I’ll hit re-run tests

VangelisP:
I got stuck with the same issue/situation as @MegaphoneJon : I have a couple of available currencies but the payment screen shows the default site currency.

I can confirm that with this patch, my issue got resolved. @KarinG could we merge this one ?

Thanks for the work @MegaphoneJon !!

KarinG:
It really needs to pass the existing battery of tests before we can merge this.

VangelisP:
Coming back to this PR as I've recently upgraded `webform_civicrm` to version 6.2.5. I have a different behaviour with 6.2.5 vs 6.2.4:

I believe [this line](https://github.com/colemanw/webform_civicrm/pull/842/files#diff-46ba4a148afe3b668cdaf3f11847502a31cd9f0a046fbd78b86f0464bee97d41R60) should be taken out.
It replaces the line item amount with the total amount

Before removing/commenting the pointed line:

image

After commenting it:

image

PS. Please ignore the total amount which is 26 in one screenshot and 21 in the other..

PR #844, opened 2023-02-15
[REF] Remove references to no longer found function civicrm_initialize()

seamuslee001:
Overview
civicrm_initialize does not exist in Drupal 8/9/10 version of CiviCRM integration module instead we use initialize() on the CiviCRM service

ping @KarinG @colemanw


KarinG:
Thank you @seamuslee001 👍

PR #851, opened 2023-02-27
[test only] Need dev version of parent webform and other updates to github d10 action yml

demeritcowboy:
Overview
Even with everything pending merged, the d10 yml won't pass if run in the main tree. This just brings it up to date with closer to what it will end up with since shortly we should be able to work on d10 directly in the main tree.

### DIFF
--- a/.github/workflows/d10.yml
+++ b/.github/workflows/d10.yml
@@ -10,11 +10,11 @@ jobs:
   phpunit:
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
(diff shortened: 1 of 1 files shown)


demeritcowboy:
The second commit now pulls in the extension-related changes from https://github.com/colemanw/webform_civicrm/pull/855 and then uses them to get the latest git versions of extensions. So now when the other test PRs are merged this yml action _should_ work to run in the main tree.

KarinG:
Ok here we go!

PR #852, opened 2023-02-27
Update info.yml to support drupal 10

demeritcowboy:
Overview
Some things to note:
* If you install it on drupal 10, you'll need the 6.2.x-dev version of the parent webform, but there isn't a compelling reason to up the version required in composer.json to `^6.2` since you don't need that for drupal 9. But composer should figure it out based on the parent webform's branch's composer.jsons you just might need to put `"minimum-stability": "dev"` in your root composer.json.
* Stripe and firewall would need to be master versions from git if using those.
* Iats needs to be patched if using that.


demeritcowboy:
I just updated the description - I think I was wrong about needing 2 branches.

KarinG:
Saving the best for last :-)

PR #855, opened 2023-03-01
[test only] Switch to downloading extensions in the yml

demeritcowboy:
Overview

Before
Each individual test for iats and stripe downloads the extensions in its setUp function, so besides lab.civicrm.org being hit about 4x4x4 times, it's difficult to do things like run against a specific version of an extension or patch it.

After
Download the extensions once in the yml. Can apply patches easily in the yml.

Technical Details
The install still happens in setUp() the same as before, just instead of the extension being in the temporary simpletest subfolder that gets created for each test, it's in a central more permanent location.

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,6 +106,10 @@ jobs:
         run: |
           cd ~/drupal
           COMPOSER_MEMORY_LIMIT=-1 composer require drupal/token:^1.9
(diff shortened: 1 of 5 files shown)


KarinG:
Ok this is going in and should be all green now for 1y (until we get passed Feb 29, 2024).

demeritcowboy:
Thanks!

PR #857, opened 2023-03-07
[test only] drupal 10 patch no longer needed

demeritcowboy:
Overview
No longer needed, and in fact would fail to apply now.


KarinG:
Right and yeah!

PR #862, opened 2023-03-19
D9 - Fix the submission of default custom field value

jitendrapurohit:
Overview
Fix the submission of default custom field value

Before
When submitting a webform which updates a custom field, the value is always set to the default value (if there is one) when "Option type" is set to "Reuse an existing set".

After
Fixed.

Technical Details
Update the `#value` with `#default_value` property.

Comments
Drupal Ticket: https://www.drupal.org/project/webform_civicrm/issues/3326150


KarinG:
👍 I ran into this issue myself last week. Did you scan the code to check if we accidentally use ‘value’ instead of ‘default_value’ for other elements perhaps as well?

jitendrapurohit:
yes, there are more than 10 such assignments. I think we should do a follow up PR after confirming the same issue on each of them. Hope thats fine?

PR #821, opened 2022-12-28
D8/9: make Contact selectable for Contributions Extend #758

jitendrapurohit:
Overview
----------------------------------------
Extend #758 and fix the following issues with the PR

- Contact 1 is not created in civicrm when 2 contacts are submitted via webform.
- Console error on settings page when Contribution contact = User select.

Tested the following usecases -

- Contribution Contact = User Select; Payment Processor = Dummy. Submit the webform as anonymous user with payment contact = 2. 
**Result Obtained** = Payment record & Billing Address created on 2nd contact (Correct).

- Contribution Contact = User Select; Paayment processor = Pay later. Submit the webform as anonymous user with payment contact = 2. 
**Result Obtained** = Payment record created on 2nd contact (Correct).

Repeated the above test cases by submitting the webform as logged in user.

This seems to work fine for me now @rubofvil @sluc23. Can you please confirm if it works for you as well?

There are still some issues on admin js, eg there is no alert to enable email if Contribution Contact = `-User select-`. I've added a help text to ensure admins enable email fields for the contribution contact.

image

Thanks.

PR #863, opened 2023-03-28
Fix bug assigning contribution second contact

rubofvil:
Overview

If you have two contacts to assign a contribution, and you expose to the user that can choose which contact will be assigned the contribuion, via component `civicrm_1_contribution_1_contribution_contact_id`.

If for any reason the **contact 1** (`civicrm_1_contact_1_contact_*`) don't have filled the info, and isn't created.
*We use conditional for not create always the contact 1*

With the previous conditions a fatal error is produced. The contribution can't be assigned to a contact.

image

Before

The contribution can't be assigned to a contact **1** and a fatal error is triggered.

After

The contribution  is assigned correctly to the contact selected and not error is triggered.

Technical Details

Comments

Ref to PR https://github.com/colemanw/webform_civicrm/pull/821

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2117,7 +2117,8 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       $paymentProcessor = \Civi\Payment\System::singleton()->getByName($processor_type['name'], TRUE);
     }
     // Add contact details to params (most processors want a first_name and last_name)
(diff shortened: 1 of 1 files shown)


rubofvil:
@jitendrapurohit this PR fix a case over the PR https://github.com/colemanw/webform_civicrm/pull/821

KarinG:
👍 I’ll ask Jitendra to add this to his list for review later this week.

jitendrapurohit:
Looks like this was only affecting IPN payments? The change looks fine and ok to merge.  Thanks for the fix @rubofvil

PR #864, opened 2023-04-08
D9 - Update value to default_value to avoid fixed value submission

jitendrapurohit:
Overview
Fix submission of fixed values on webform submission.

Before
Some fields submits fixed values on webform submission if `-User Select-` is enabled on them. Eg:

- Preferred Language
- Address, Email, Phone, IM Location Type & Is Primary.
- Case Client is always set to first contact if `-User Select-` is enabled.
- Number of Terms field for membership.
- Grant Status
- Participant Role
- Participant Status

After
The expectation here is to load default value on them, not submit the fixed values. Updating 'value' to `default_value` fixe the usecase.

Comments
@KarinG This is a followup for https://github.com/colemanw/webform_civicrm/pull/862

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -253,7 +253,7 @@ class Fields implements FieldsInterface {
       $fields['contact_preferred_language'] = [
         'name' => t('Preferred Language'),
         'type' => 'select',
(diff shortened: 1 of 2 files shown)


KarinG:
Thank you @jitendrapurohit - I have come across these `defaults` but it's always been ease to just clear them. It's about time we fix this :-)

PR #866, opened 2023-04-08
Issue #3353053: Skip case types that are inactive and with no roles

herbdool:
Fixes https://www.drupal.org/project/webform_civicrm/issues/3353053

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -572,11 +572,8 @@ class Fields implements FieldsInterface {
         ];
         // Fetch case roles
         $sets['caseRoles'] = ['entity_type' => 'case', 'label' => t('Case Roles')];
-        foreach ($this->utils->wf_crm_apivalues('case_type', 'get') as $case_type) {
-          if (!$case_type['is_active'] || !isset($case_type['definition']['caseRoles'])) {
-            continue;
-          }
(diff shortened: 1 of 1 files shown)


src/Fields.php
```diff
@@ -573,6 +573,9 @@ protected function wf_crm_get_fields($var = 'fields') {
         // Fetch case roles
         $sets['caseRoles'] = ['entity_type' => 'case', 'label' => t('Case Roles')];
         foreach ($this->utils->wf_crm_apivalues('case_type', 'get') as $case_type) {
+          if (!$case_type['is_active'] || !isset($case_type['definition']['caseRoles'])) {
+            continue;
+          }
           foreach ($case_type['definition']['caseRoles'] as $role) {
```
jitendrapurohit: @herbdool Can we check for active condition in the API instead? and maybe, optimise the above changes to the following?

````
foreach ($this->utils->wf_crm_apivalues('case_type', 'get', ['is_active' => 1]) as $case_type) {
  foreach ($case_type['definition']['caseRoles'] ?? [] as $role) {
````

herbdool: I've changed the PR with your suggestion @jitendrapurohit

KarinG:
@jitendrapurohit - that looks good - do you want to check it? Test fails are unrelated.

PR #867, opened 2023-04-08
Issue #3353056: skip NULL elements

herbdool:
Fixes https://www.drupal.org/project/webform_civicrm/issues/3353056

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -194,6 +194,10 @@ function _fillCiviCRMData($data, $webformSubmission) {
   $webform = $webformSubmission->getWebform();
   foreach ($data as $key => $val) {
     $element = $webform->getElement($key);
+    // Element no longer exists so skip.
+    if (!$element) {
+      continue;
+    }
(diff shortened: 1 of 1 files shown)


webform_civicrm.module
```diff
@@ -194,6 +194,10 @@ function _fillCiviCRMData($data, $webformSubmission) {
   $webform = $webformSubmission->getWebform();
   foreach ($data as $key => $val) {
     $element = $webform->getElement($key);
+    // Element no longer exists so skip.
+    if (!$element) {
+      continue;
+    }
```
jitendrapurohit: Can't imagine a scenario where $key would be present in $data but not as a webform element. However, it seems like a kind precautionary measure. @herbdool Can we just include this check in the same if statement (line 201) ?

    if ($element && !empty($val) && $element['#type'] == 'civicrm_options') {

herbdool: @jitendrapurohit yes, I can do that.

I'm not sure of the scenario too, but it did happen to a client so it must be possible. This client uses a lot of webforms so we've encountered a bunch of edge cases with them.

PR #868, opened 2023-04-12
Do not display "Existing contact" fields as autocomplete when not configured as such

jensschuppe:
drupal.org issue: [#3353716](https://www.drupal.org/project/webform_civicrm/issues/3353716)

Overview
Existing Contact fields can be configured to display as either an autocomplete, or a select field, or a textfield ("Enter Contact ID"), or be hidden. When selecting the textfield ("Enter Contact ID") option, the field will nevertheless be displayed as an autocomplete field. That is due to JS code only checking for whether the field ist "not a select field".

Steps to reproduce
1. Create a Webform with CiviCRM processing enabled
2. Enable a contact and add the "Existing Contact" field
3. Go to the "Build" tab and edit the "Existing Contact" field to display as "Enter Contact ID" for form display and disable prepopulation from URL or current contact
4. View the Webform, the "Existing Contact" field is being displayed as if it were an autocomplete field, while it should be just a simple textfield. It's basically just as if it were configured to be an "autocomplete" field, which is a separate form display option.

Proposed resolution
Add an `is-autocomplete` attribute for the webform field and let the JS code depend on that rather than the field not being a select field for initializing autcomplete behavior.

### DIFF
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -9,19 +9,21 @@
         }
         var autocompleteUrl = D.url('webform-civicrm/js/' + field.data('form-id') + '/' + field.data('civicrm-field-key'));
         var isSelect = field.data('is-select');
-        var isAutocomplete = field.data('is-autocomplete');
-        if (isAutocomplete) {
+        if (!isSelect) {
+          var tokenValues = field.data('is-contactid') ? false : {
+            hintText: field.data('search-prompt'),
+            noResultsText: field.data('none-prompt'),
+            resultsFormatter: formatChoices,
+            searchingText: "Searching..."
+          };
           wfCivi.existingInit(
             field,
             field.data('civicrm-contact'),
             field.data('form-id'),
             autocompleteUrl,
-            toHide, {
-            hintText: field.data('search-prompt'),
-            noResultsText: field.data('none-prompt'),
-            resultsFormatter: formatChoices,
-            searchingText: "Searching..."
-          });
+            toHide,
+            tokenValues
+          );
         }
 
         field.change(function () {
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -9,7 +9,8 @@
         }
         var autocompleteUrl = D.url('webform-civicrm/js/' + field.data('form-id') + '/' + field.data('civicrm-field-key'));
         var isSelect = field.data('is-select');
-        if (!isSelect) {
+        var isAutocomplete = field.data('is-autocomplete');
+        if (isAutocomplete) {
           wfCivi.existingInit(
             field,
             field.data('civicrm-contact'),
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -130,6 +130,9 @@ class CivicrmContact extends WebformElementBase {
       $element['#options'] = [];
       $element['#attributes']['data-is-select'] = 1;
     }
+    elseif ($element['#widget'] === 'autocomplete') {
+      $element['#attributes']['data-is-autocomplete'] = 1;
+    }
     // get translated properties used for autocomplete widget
     $element['#attributes']['data-search-prompt'] = $this->getTranslatedElementProperty($element, 'search_prompt') ?? t('- Choose existing -');
     $element['#attributes']['data-none-prompt'] = $this->getTranslatedElementProperty($element, 'none_prompt') ?? t('+ Create new +');
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -130,8 +130,10 @@ class CivicrmContact extends WebformElementBase {
       $element['#options'] = [];
       $element['#attributes']['data-is-select'] = 1;
     }
-    elseif ($element['#widget'] === 'autocomplete') {
-      $element['#attributes']['data-is-autocomplete'] = 1;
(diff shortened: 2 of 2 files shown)


KarinG:
Thank you for the PR! I have approved your user to kick off our Functional JS tests - they are running now.

KarinG:
Test failure is:

image

jensschuppe:
Seems I'll need some help understanding what exactly this code is supposed to do:

https://github.com/colemanw/webform_civicrm/blob/baa1832845f3ea5837a9b6087d77bf5d4832bea4/js/webform_civicrm_contact.js#L13-L23

I assumed this takes care of the existing contact field to be an autocomplete field, but it seems to be doing much more, like hiding fields; the test failure was at

https://github.com/colemanw/webform_civicrm/blob/baa1832845f3ea5837a9b6087d77bf5d4832bea4/tests/src/FunctionalJavascript/ExistingContactElementTest.php#L156-L157

Can someone help me out with this? Maybe the tests should be adjusted, too ...

KarinG:
Hi @jensschuppe - happy to help you out. I'll try to find some time tomorrow to dig up the screenshots of the test (to see what the fail looks like). If needed I will ask @jitendrapurohit to help you out [when he had time next].

js/webform_civicrm_contact.js
```diff
@@ -9,7 +9,8 @@
         }
         var autocompleteUrl = D.url('webform-civicrm/js/' + field.data('form-id') + '/' + field.data('civicrm-field-key'));
         var isSelect = field.data('is-select');
-        if (!isSelect) {
+        var isAutocomplete = field.data('is-autocomplete');
+        if (isAutocomplete) {
```
jitendrapurohit: @jensschuppe I think its just the last argument that needs to be avoided in case of contact id widget. Since the ajax still need to be called for a user that enters contact id in the text field. Can you please confirm if the following update to this PR works for you:

```
diff --git a/js/webform_civicrm_contact.js b/js/webform_civicrm_contact.js
index 5eca49c..7143419 100644
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -10,17 +10,20 @@
         var autocompleteUrl = D.url('webform-civicrm/js/' + field.data('form-id') + '/' + field.data('civicrm-field-key'));
         var isSelect = field.data('is-select');
         if (!isSelect) {
+          var tokenValues = field.data('is-contactid') ? false : {
+            hintText: field.data('search-prompt'),
+            noResultsText: field.data('none-prompt'),
+            resultsFormatter: formatChoices,
+            searchingText: "Searching..."
+          };
           wfCivi.existingInit(
             field,
             field.data('civicrm-contact'),
             field.data('form-id'),
             autocompleteUrl,
-            toHide, {
-            hintText: field.data('search-prompt'),
-            noResultsText: field.data('none-prompt'),
-            resultsFormatter: formatChoices,
-            searchingText: "Searching..."
-          });
+            toHide,
+            tokenValues
+          );
         }
 
         field.change(function () {

```

src/Plugin/WebformElement/CivicrmContact.php
```diff
@@ -130,6 +130,9 @@ public function prepare(array &$element, WebformSubmissionInterface $webform_sub
       $element['#options'] = [];
       $element['#attributes']['data-is-select'] = 1;
     }
+    elseif ($element['#widget'] === 'autocomplete') {
+      $element['#attributes']['data-is-autocomplete'] = 1;
+    }
```
jitendrapurohit, line 137: Also, I think it would be nice to check for textfield instead of autocomplete since we do have some description and size attribute to attach to the contact id field. Eg:

```
diff --git a/src/Plugin/WebformElement/CivicrmContact.php b/src/Plugin/WebformElement/CivicrmContact.php
index 0dbf37a..e3d8a24 100644
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -130,6 +130,11 @@ class CivicrmContact extends WebformElementBase {
       $element['#options'] = [];
       $element['#attributes']['data-is-select'] = 1;
     }
+    elseif ($element['#widget'] === 'textfield') {
+      $element['#attributes']['data-is-contactid'] = 1;
+      $element['#description'] = 'Lookup by Contact ID.';
+      $element['#size'] = 8;
+    }
     // get translated properties used for autocomplete widget
     $element['#attributes']['data-search-prompt'] = $this->getTranslatedElementProperty($element, 'search_prompt') ?? t('- Choose existing -');
     $element['#attributes']['data-none-prompt'] = $this->getTranslatedElementProperty($element, 'none_prompt') ?? t('+ Create new +');
```

jitendrapurohit (changes requested):
@jensschuppe I've made some comments to the changes required. Can we do that update please to check if it passes the test results?

Thanks.

js/webform_civicrm_contact.js
jensschuppe: Thanks @jitendrapurohit, this seems to work, I added this with 42e1fcc3ef594648252eb75e986deff6f5a1b821.

src/Plugin/WebformElement/CivicrmContact.php
jensschuppe: Added with 42e1fcc3ef594648252eb75e986deff6f5a1b821.

jitendrapurohit (approved):
Working as expected.

image

Thanks for the update @jensschuppe

KarinG:
Thank you @jensschuppe for making the changes 👍

PR #869, opened 2023-04-15
D9 - Fix js break on admin screen

jitendrapurohit:
Overview
Fix test failures caused by js break on admin screen.

Before
Replicates on Drupal > 9.5.

Admin Screen has console errors 

image

After
Fixed.

Comments
@KarinG


KarinG:
Can't believe we have not noticed this before :-)

I spotted an iATS related failure (transaction not authorized) - that's very likely the test account. So re-running two of the Integration Tests.

KarinG:
All green now 👍

PR #865, opened 2023-04-08
initialize CiviCRM before calling civicrm_api

herbdool:
Fixes https://www.drupal.org/project/webform_civicrm/issues/3353043

PR #871, opened 2023-04-20
Initialize CiviCRM in _fillCiviCRMData()

jensschuppe:
drupal.org issue: [#3355339](https://www.drupal.org/node/3355339)

Overview
In some cases viewing results of a CiviCRM webform leads to a `Call to undefined function Drupal\webform_civicrm\civicrm_api3() in Drupal\webform_civicrm\Utils->wf_civicrm_api()` error due to CiviCRM not being initialized yet.

Steps to reproduce
Create a Webform with CiviCRM processing enabled, submit it and try to view results. I'm not sure about the specific scenario and don't have the time to investigate as the Webform this appears on is quite large.

Proposed resolution
In `_fillCiviCRMData()` there should be a `\Drupal::service('civicrm')->initialize();`, which makes sure we have our global functions available. Re-initializing does not hurt as it early-returns in that case.

### DIFF
--- a/webform_civicrm.module
+++ b/webform_civicrm.module
@@ -190,6 +190,7 @@ function _fillCiviCRMData($data, $webformSubmission) {
     return $data;
   }
 
+  \Drupal::service('civicrm')->initialize();
   $utils = \Drupal::service('webform_civicrm.utils');
   $webform = $webformSubmission->getWebform();
   foreach ($data as $key => $val) {


jitendrapurohit:
Looks similar to https://github.com/colemanw/webform_civicrm/pull/865. 

@jensschuppe  CiviCRM should be initialized using the CiviCRM Webform Handler plugin. See https://github.com/colemanw/webform_civicrm/blob/6.x/src/Plugin/WebformHandler/CivicrmWebformHandler.php#L110

Can you please confirm if webform_civicrm handler is present when you visit - /admin/structure/webform/manage/<webform_id>/handlers in the URL?

image

If not - then the setup seems to be wrong. Is this webform migrated? If yes, the migration code need to create this handler in D9 to function correctly.

Overall, I think if multiple users are encountering this, merging this change isn't harmful though. Thoughts @KarinG ?

KarinG:
Yeah - I'm not sure why people run into this. I've not - but then we have not migrated webforms yet - we rebuild to use the new functionality. 

I've triggered the tests and then will merge this.

jensschuppe:
> Can you please confirm if webform_civicrm handler is present when you visit - /admin/structure/webform/manage/<webform_id>/handlers in the URL?
> 
> image
> 
> If not - then the setup seems to be wrong. Is this webform migrated? If yes, the migration code need to create this handler in D9 to function correctly.

Thank you, @colemanw - you are right, this is a migrated webform and the handler is indeed missing. I don't know why, but during migration neither the CiviCRM handler nor the "Enable CiviCRM Processing" setting were there. All other settings have been migrated correctly. We used a patched version of `webform_migrate` for migrating CiviCRM-integrated webforms, as there's no native support for this module in `webform_migrate`. Other webforms have been migrated correctly with this code, so there must have been something wrong with this webform anyway.

Sorry for that, but I think trying to re-init CiviCRM doesn't hurt either, so thanks for merging! Maybe users should be informed in some way when a CiviCRM-processed webform is misconfigured?

PS. Maybe CiviCRM processing has been disabled after migrating the Webform correctly - in that case this PR would still be useful, as viewing results should be independent of whether CiviCRM processing is currently enabled for the Webform or not.

PR #876, opened 2023-05-13
D9 - Fix Membership test failure on drupal v9.4

jitendrapurohit:
Overview
Fix Membership test failure on drupal v9.4

Before
Membership test fails

After
Pass.


KarinG:
Super 👍

PR #878, opened 2023-05-25
Adjust test for receipt subject line fix in core 5.63

demeritcowboy:
Overview
In core 5.63 the subject no longer says Invoice when a pay later is completed and the receipt is sent. That sounds correct, so adjust the test to match.


KarinG:
Thank you 🙂

demeritcowboy:
Thanks!

PR #879, opened 2023-06-01
When searching for a contact subtype we have to use the 'CONTAINS' op…

mattwire:
…erator or it will not find contacts with multiple subtypes

Overview
See https://github.com/civicrm/civicrm-core/pull/25931

Before
Using "existing contact" filter with a contact subtype filter does not find contacts with multiple subtypes.

After
Using "existing contact" filter with a contact subtype filter does find contacts with multiple subtypes.

Technical Details

Comments


jitendrapurohit:
Thanks @mattwire

PR #880, opened 2023-06-03
D9 - WSOD when editing results and pseudoconstant labels don't match Civi

jitendrapurohit:
Overview
WSOD when editing results and pseudoconstant labels don't match Civi

Before
Detailed steps to reproduce (embed screenshots)

* Create a form that contains a contact prefix field (Mr., Mrs., etc.).
* On the Build tab, change the prefix field to static options. Also modify the labels. I think it needs to be a Listbox.
* Enter a sample submission.
* Go to Results and edit the submission.
* Go to Edit. Note that the prefix you selected appears twice in the select box.
* Go to Notes and attempt to submit. Get a WSOD.

After
Fixed.

Comments
Drupal Ticket - https://www.drupal.org/project/webform_civicrm/issues/3363414

FYI - @MegaphoneJon

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2708,6 +2708,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       $data = $webform_submission->getData();
     }
     else {
+      $webform_submission = $this->submission;
       $data = $this->submission->getData();
     }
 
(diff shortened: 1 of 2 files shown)


jitendrapurohit:
Ah, the test failed which indicates the `webform_civicrm_webform_submission_load()` is actually used to display entity links on the submission page. I've restored that part.

The wsod is still fixed by the updated PR.

PR #881, opened 2023-06-05
D9 - Fill cc values after billing address

jitendrapurohit:
Overview
Fill cc values after billing address

Before
Test fails intermittently for Drupal 10.

After
Pass.

Technical Details
The test fails to enter number on the Credit card field. This field is displayed after user enters an amount. Maybe, the test is too fast and tries to enter the cc value before it is fully loaded.

Fix here is to fill billing address first so it gets enough time to render the cc field.

@KarinG


KarinG:
Ah clever - this past on first go. Will merge it and see if it is consistent.

KarinG:
Will wait for re-run on 5.47/D9 to be completed. That just looked like a one - off timeout, but will wait to see it completed.

KarinG:
Another time-out. Ok one more try!

KarinG:
Still similar behaviour - can you think of any other re-arranging - or perhaps adding a wait?

PR #884, opened 2023-06-12
fixed fatal error when editing submission with cs field

pradpnayak:
Overview
TypeError: Unsupported operand types: string * int in wf_crm_webform_postprocess->fillHiddenContactFields() (line 727 of /var/www/html/drupal7/sites/defaults/modules/contrib/webform_civicrm/includes/wf_crm_webform_postprocess.inc).


KarinG:
@pradpnayak - we're D9/D10 first, so please PR to 6.x - so that it will run the extensive test matrix on it. After we see that PR through -> then we can look at backporting to D7.

pradpnayak:
Hi @KarinG 

The field works fine on D9, but fails on d7 because it uses different code for submissionValue().

KarinG:
What's that section of code like in D9? Is this PR a direct backport?

PR #885, opened 2023-06-14
apply variants on postProcess

MegaphoneJon:
Don't merge yet - this is to see if it passes tests.

Overview
Webform-CiviCRM doesn't account for variants in the postProcess.

Before
Variants ignored

After
Variants applied.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -76,7 +76,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
     }
 
     $this->node = $webform_submission->getWebform();
-
+    $this->node->applyVariants($webform_submission);
     $handler_collection = $this->node->getHandlers('webform_civicrm');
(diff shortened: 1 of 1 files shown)


MegaphoneJon:
With tests passing, I'm going to deploy this into production.  I'll request a review/merge once I'm satisfied that it's not causing issues.

MegaphoneJon:
I never requested a review here - but this has been in production since June with no issues.

jitendrapurohit:
@MegaphoneJon I notice that variants are already being implemented at various levels in the core webform module. Do you have any specific usecase for me to test that wasn't functioning prior to this change?

MegaphoneJon:
@jitendrapurohit Because WFC wasn't variant-aware, we were hitting issues when the "settings" portion of a variant (which contains all the non-"User Select" options in WFC) changed.

So in my particular case, my primary page used Authorize.net, but my variants used Stripe.  When submitting a variant, we'd receive messages about required fields being missing that are only required with Authorize, not Stripe (see attached).  But I imagine that many other issues with variants are possible because of this.

image

PR #887, opened 2023-06-23
[NFC] Remove unit test line that installs from zip only

demeritcowboy:
Overview
I don't think this is needed anymore, and interferes with `git am` lower down.

Before
This line makes composer attempt to use the zip download from github first. It means that most of the time the `git am` call will fail. It's not called very often, but is useful.

After
Should be no change in terms of what gets tested.

Technical Details
Civi used to "guess" at what type of errors and logging to do, and one of the ways it did that was to see if there was a .git folder present. Civi doesn't guess anymore.

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,6 @@ jobs:
           composer config extra.compile-mode all
           composer config minimum-stability dev
           composer config prefer-stable true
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you Dave!

demeritcowboy:
Thanks!

PR #888, opened 2023-07-10
[NFC] Unit tests - allow installing from git

demeritcowboy:
Overview
Followup to https://github.com/colemanw/webform_civicrm/pull/887

Before
Always installs from zip file from github

After
Installs from zip when the package is a numbered release, from git clone when it's a `.x-dev` or `dev-master` branch.

Technical Details
PR 887 was correct about the part about .git folders, however there are two further pieces of information:
1. There's a specific override in the github workflow line that downloads civi, which I didn't notice.
2. In composer 2.1 they changed the default so that using zips is the default, so in PR 887 removing the line that set it to "dist" actually did nothing since that line has done nothing since composer 2.1.

Setting it to "auto" sets it to "dist" for numbered releases, and "source" for dev installs. This is a good compromise. (Dist is slightly faster, which is why it's the default, but source is needed since otherwise sometimes patches won't apply.)

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,6 +77,7 @@ jobs:
           composer config extra.compile-mode all
           composer config minimum-stability dev
           composer config prefer-stable true
+          composer config preferred-install auto
           composer config allow-plugins.civicrm/composer-compile-plugin true
           composer config allow-plugins.civicrm/composer-downloads-plugin true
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you Dave!

demeritcowboy:
Thanks!

PR #889, opened 2023-07-10
M is not m

demeritcowboy:
Overview
M is Mar, m is 03.

Before
Tests failing after core change yesterday: https://github.com/civicrm/civicrm-core/pull/26768

After

Technical Details
It's looking for a field that no longer has the same name, because for some reason the date format is part of the field name.

Comments
Coincidentally this is lexically really similar to my [other PR](https://github.com/civicrm/civicrm-core/pull/26780) today, but for completely different reasons/tech.


KarinG:
Very confident -> merging!

demeritcowboy:
Oh right. Before 5.65 it's going to still be M.

PR #894, opened 2023-07-18
Issue #3375337: only call civi in hook if handler enabled

herbdool:
Fixes https://www.drupal.org/project/webform_civicrm/issues/3375337


jitendrapurohit (changes requested):
Applying this patch returns a fatal error on results page (if civicrm handler is disabled). To replicate:

- Create a webform with civicrm enabled.
- Submit a contact using the form.
- Go to the results page.
- Disable the webform_civicrm handler from webform settings.
- Reload the results page.
- Error.

herbdool:
@jitendrapurohit What was the error message? What is the desired result? I don't think it makes sense for the results that were submitted with webform_civicrm handler enabled, *not* to have some errors. Maybe just handle the error and display a message that the webform_civicrm handler is disabled? Or don't display anything for the civicrm-connected fields?

I think the error you found is better as a new issue. There has to be a way to truly disable the handler, which is what this PR attempts to do. How the results display deals with the civicrm handler being disabled seems like a new issue to me.

jitendrapurohit:
@herbdool I mentioned the error because the PR aimed to fix "something" if the handler is disabled. I'm not sure what was the original issue this PR was trying to fix. Can you add more details pls? 

Note: The above mentioned error replicates only after this change is applied and works fine without it.

> There has to be a way to truly disable the handler, which is what this PR attempts to do.

Can you pls explain what is the exact use-case you're trying to fix? And the reason behind disabling the handler? what is the expected workflow after the handler is disabled? Do you still expect to see civicrm fields on the webform? If no, maybe a proper way to handle it would be to also disable all the civi fields from the webform?

herbdool:
@jitendrapurohit I'll get back to this soon. Basically, this supports migrating webform submissions. We can't have the migration triggering the handler. It should only be migrating and transforming via the migration script. If the handler is enabled during this migration (or if this hook is calling civi even if the handler is disabled) then webform_civicrm does all kinds of stuff assuming it's someone filling out a form.

I'll ask the reverse question: does it ever make sense to have handler that can't be truly disabled? If the handler is disabled on a form, nothing should trigger CiviCRM for that form. So the hook should respect that.

herbdool:
@jitendrapurohit I cannot replicate a fatal error. I do *not* get a fatal error when visiting the Results page (after submitting the webform), when I have applied this patch, and have the handler disabled.

*Without* the patch, I get this error:

> Call to undefined function Drupal\webform_civicrm\civicrm_api3() in Drupal\webform_civicrm\Utils->wf_civicrm_api() (line 657 of modules/contrib/webform_civicrm/src/Utils.php).

But it loads fine otherwise. Can you test this again?

jmcclelland:
@jitendrapurohit I am also  unable to replicate the errors you are getting. 

And  I also need this patch so I can migrate our webform submissions from Drupal 7 to Drupal 10. I've now [patched the webform_civicrm_migrate](https://github.com/fuzionnz/webform_civicrm_migrate/pull/12) so it disables the webform_civicrm handler. However, without this patch it doesn't do any good.

PR #896, opened 2023-07-23
D9 - Set default currency on the contribution tab

jitendrapurohit:
Overview
Save button does nothing is currency is not set to any value.

Before
Currency field is required. If user fails to set it and navigates to another tab (eg Membership) & saves the webform => nothing happens and console returns the following error:

image

This is because the Contribution tab is not active on the browser. Saving the form now has required validation error on the invisible field. 

After
Default currency is pre-selected.

image

Comments
Drupal Ticket - https://www.drupal.org/project/webform_civicrm/issues/3375985


petednz:
Tested - sets currency to default. thx

mattwire:
@KarinG good to merge

PR #898, opened 2023-07-24
Fix missing financial_account_id for recurring contributions

mattwire:
Overview
Both `price_field_id` and `price_field_value_id` need to be set when creating the lineitems.
For memberships we were retrieving that data but not using it. For contributions we were only retrieving the default `price_field_id` and not setting `price_field_value_id`.

This resulted in errors such as found in https://lab.civicrm.org/extensions/mjwshared/-/issues/20 where CiviCRM ends up not setting the financial_account_id.

Before
"Invalid" lineitems created.

After
Valid lineitems created.

Technical Details
CiviCRM has a default "Price field ID" and a default "Price field value ID" which should be used if nothing else is specified. For memberships we can map them to existing ones.

Comments

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2345,32 +2345,40 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       if (empty($item['contribution_id'])) {
         $item['contribution_id'] = $id;
       }
-      $priceSetId = 'default_contribution_amount';
+
       // Membership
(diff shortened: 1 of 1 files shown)


jitendrapurohit (approved):
Tested by different combinations of amounts on the webform and this works as expected. Thanks @mattwire

mattwire:
I think the fix actually needs to be in CiviCRM core as this won't work with multiple line items - see https://github.com/civicrm/civicrm-core/pull/26974

PR #753, opened 2022-06-15
Fix addresses not hiding when "Shared Address" is enabled.

puresyntax71:
Overview
----------------------------------------
The address fields are not hiding when you have enabled "Share address of" set to "User select" and check "Contact 1". This is an example export:

<details>
    <summary>Export</summary>

```yaml
uuid: 6ff09daf-5c9a-4c1d-9930-d70baa2f9372
langcode: en
status: open
dependencies:
  module:
    - webform_civicrm
weight: 0
open: null
close: null
uid: 1
template: false
archive: false
id: test
title: Test
description: ''
category: ''
elements: |-
  civicrm_1_contact_1_fieldset_fieldset:
    '#type': fieldset
    '#title': 'Contact 1'
    '#form_key': civicrm_1_contact_1_fieldset_fieldset
    civicrm_1_contact_1_contact_existing:
      '#type': civicrm_contact
      '#title': 'Existing Contact'
      '#widget': hidden
      '#none_prompt': '+ Create new +'
      '#results_display':
        display_name: display_name
      '#no_autofill':
        address: address
      '#default': user
      '#allow_create': 1
      '#contact_type': individual
      '#form_key': civicrm_1_contact_1_contact_existing
      '#parent': civicrm_1_contact_1_fieldset_fieldset
      '#extra': {  }
    civicrm_1_contact_1_contact_first_name:
      '#type': textfield
      '#counter_type': character
      '#counter_maximum': 64
      '#counter_maximum_message': ' '
      '#contact_type': individual
      '#form_key': civicrm_1_contact_1_contact_first_name
      '#extra':
        width: 20
      '#parent': civicrm_1_contact_1_fieldset_fieldset
      '#title': 'First Name'
    civicrm_1_contact_1_contact_last_name:
      '#type': textfield
      '#counter_type': character
      '#counter_maximum': 64
      '#counter_maximum_message': ' '
      '#contact_type': individual
      '#form_key': civicrm_1_contact_1_contact_last_name
      '#extra':
        width: 20
      '#parent': civicrm_1_contact_1_fieldset_fieldset
      '#title': 'Last Name'
  civicrm_2_contact_1_fieldset_fieldset:
    '#type': fieldset
    '#title': 'Contact 2'
    '#form_key': civicrm_2_contact_1_fieldset_fieldset
    civicrm_2_contact_1_contact_first_name:
      '#type': textfield
      '#counter_type': character
      '#counter_maximum': 64
      '#counter_maximum_message': ' '
[...]

PR #903, opened 2023-09-08
fix address sharing with user select

MegaphoneJon:
Overview
PR #753 introduced a syntax error by unintentionally deleting a close square bracket.

To trigger this, you must enable address sharing with user select.  When you select a user, you'll get a syntax error in your JS console.

Before
JS broken after selecting address sharing.

After
Fixed.

Comments
Just a one byte change!


jitendrapurohit:
Thanks @MegaphoneJon

PR #906, opened 2023-09-13
[NFC] hack for chromedriver no longer needed

demeritcowboy:
Overview
I don't think this is needed anymore.


KarinG:
Bedankt! :-)

PR #911, opened 2023-09-24
D9/10 - Add setting to disable primary participant for Contact 1

jitendrapurohit:
Overview
Add setting to disable storage of registered_by_id on multiple participant registration.

Before
If multiple contacts are registered to an event using webform, contact 1 pid is stored as registered_by_id for all other participant records.

After
Config introduced to make this optional.

image

If enabled, all participants created on webform submission will be stored as normal participant records instead of being secondary records with primary as Contact 1.

Comments
@KarinG


KarinG:
Thank you Jitendra for getting to this quickly. Tested and merging :-)

PR #914, opened 2023-10-08
(alternate to 913) Temp unit test update for IATS to handle differences in versions

demeritcowboy:
Overview
Alternate to #913 

The field label is different in the git version of IATS, which is installed during drupal 10 runs but not during drupal 9 runs.

Before
Bank Account Number

After
Account No.

Technical Details
The drupal 10 run installs git versions of extensions, whereas drupal 9 installs released versions, and this label changed recently in iats master.

Comments
The reason that drupal 10 and 9 do different things is because when drupal 10 was first added to the test matrix, the various extensions didn't support drupal 10 in their released versions, so an alternate would be to take another look at whether the extensions used in tests are all now supporting drupal 10 in their released versions, and then remove that difference. Although it's kind of nice testing with the additional variation to get advance warning of extension issues. 🤷 

Another alternate specifically for this label issue would be a more sophisticated `if` block which looks at which version of iats got installed but it's not straightforward, and would still be temporary.


KarinG:
Thanks Dave - I meant to follow up my 913 with either this or with poking Alan to issue a new iATS release.

PR #915, opened 2023-10-08
update test matrix

demeritcowboy:
Overview
Update test matrix

An alternate would be keep 5.65 since because of the security release that's what many people are using, and then add a 5th to get the new RC 5.67 in there.


KarinG:
Yes, I'd like to keep 5.65.x in the matrix.

KarinG:
Thank you Dave!

PR #917, opened 2023-10-13
Allow manually running unit tests against a drupal 8 civi PR

demeritcowboy:
Overview
There's already the ability to do this for a core PR. This extends it to allow doing it on a PR from the https://github.com/civicrm/civicrm-drupal-8 repo.

Before
can't (easily)

After
can (easily)

Technical Details

Comments
I've already run this on one in my own fork, which is always another way to do this kind of thing, but it's convenient to have it in the workflow dropdown here.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,9 @@ on:
       coreprurl:
         description: (optional) Core PR URL
         required: false
+      drupal8prurl:
+        description: (optional) Drupal 8 PR URL
+        required: false
 
 jobs:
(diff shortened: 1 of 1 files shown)


KarinG:
Awesome - thanks Dave. Will merge and check on Merge Actions later

demeritcowboy:
Thanks.
Yeah it might be worth liberally sprinkling session->wait calls around for these field not founds that keep coming up and see if that helps. Or replacing fillfield calls with a convenience func that waits first.

PR #918, opened 2023-10-13
remove WFC 4.x-era test mode workaround

MegaphoneJon:
Overview
This code dates to when WFC forms required a matching Civi-native contribution page.

Given a payment processor with ID 1 and the corresponding test processor with ID 2:
* `civicrm_contribution_page` always stores the ID of 1, because test mode is set in the URL.
* WFC stores the *actual* payment processor ID.

But in the old days, we inherited the ID from the native page, so WFC would get a 1, even in test mode.

This code looked up the test processor ID - but that's no longer necessary.  We now store the correct ID.

Before
Code to look up the correct ID, which we already have.

After
We use the correct ID.

Comments
If it makes the reviewer feel better, all this code is wrapped in an `if` so it only fires in test mode.  This can't affect live mode.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2245,27 +2245,6 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
         $params[$key] = $value;
       }
     }
-
-    // Fix bug for testing.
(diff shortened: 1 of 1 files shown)


MegaphoneJon:
test this please

PR #919, opened 2023-10-19
Address display of tags

demeritcowboy:
Overview
This should address the test fail but I'm not sure yet what api3 will return for `return` => `tag` lower down. Let's find out.

### DIFF
--- a/src/Plugin/WebformElement/CivicrmContact.php
+++ b/src/Plugin/WebformElement/CivicrmContact.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\webform_civicrm\Plugin\WebformElement;
 
-use CRM_Core_BAO_Tag;
 use Drupal\Component\Utility\Html;
 use Drupal\Component\Utility\Xss;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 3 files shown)


petednz:
I clicked the link to see what failings were happening with Addresses - lol.

demeritcowboy:
Ha made you look.

But good point I should update the title.

KarinG:
😂 Dave is looking at Major Donor now being Major_Donor in dev-master. Investing in automated tests has been the best thing ever.

demeritcowboy:
It's currently failing because `civicrm_api3('contact', 'get', 'return' => ['tag'])` will return the `name` for the tag. I'm going to argue that this is "correct", simply because there's no robust way to get from label to name if instead it returned label, but you can always get from name to label. In the context of this test, it's using it for comparison/validation, so name is appropriate. I can see an argument that some people might be using such a call for display purposes and so now that same call will display the internal name, but it has to be one or the other, so it's going to be "wrong" for half of the people either way.

KarinG:
Funny the branch is major-donor (perhaps another variation we need to account for) :-)

KarinG:
@jitendrapurohit - I've reviewed and tested. Since it's not trivial - could you please have a look this weekend as well? If you concur - then go ahead and hit the Merge button!

jitendrapurohit:
ah, maybe we missed the tagset. Follow up PR - https://github.com/colemanw/webform_civicrm/pull/921

PR #920, opened 2023-10-23
IATS now released so field is the same field in all versions

demeritcowboy:
Overview
Temp code needed to handle different IATS versions no longer needed

This will still fail because of the tag stuff which is still unresolved.


KarinG:
Merged - expecting dev-master to fail as there is lots to sort re: tags.

PR #921, opened 2023-10-29
Fix field name for tagset

jitendrapurohit:
Overview
Follow up from https://github.com/colemanw/webform_civicrm/pull/919

### DIFF
--- a/src/Fields.php
+++ b/src/Fields.php
@@ -597,8 +597,9 @@ class Fields implements FieldsInterface {
           }
         }
       }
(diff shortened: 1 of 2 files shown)


KarinG:
Tested! I think Name and Label is all sorted now for Tags.

demeritcowboy:
Good find.
This does bring up something in core though: https://lab.civicrm.org/dev/core/-/issues/4740

KarinG:
Oopsies

PR #923, opened 2023-11-12
D9 - Fix autocomplete search if contact id plus display fields are selected

jitendrapurohit:
Overview
Autocomplete search broken if Contact ID plus other Contact Display Fields are selected

Before
See drupal ticket - https://www.drupal.org/project/webform_civicrm/issues/3396214

After
Fixed.

Technical Details
Ignore contact id while searching for string values.

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -142,6 +142,9 @@ class ContactComponent implements ContactComponentInterface {
     if ($str) {
       $searchFields = [];
       foreach ($display_fields as $fld) {
+        if ($fld == 'id' && !is_numeric($str)) {
+          continue;
+        }
         $searchFields[] = [$fld, 'CONTAINS', $str];
       }
       $params['where'][] = ['OR', $searchFields];


jitendrapurohit:
Jude tested this on drupal ticket https://www.drupal.org/project/webform_civicrm/issues/3396214#comment-15316113. Can we merge this? @KarinG

jmcclelland:
I have also tested this patch and it works. @KarinG Do you think you could merge this one? Thank you!

jmcclelland:
FYI, I think this is mostly a core bug (see https://lab.civicrm.org/dev/core/-/issues/6397) but fixing it here is also a good move.

KarinG:
@jmcclelland - can you confirm this is still an issue after we merged (and released) this:
https://github.com/colemanw/webform_civicrm/pull/1113

jmcclelland:
Oh yes! Thanks @KarinG That is a better fix because it still allows you to search by the contact id. I just tested #1113 and it solves this problem so I think this merge request can be closed.

PR #932, opened 2023-12-19
Remove drupal 10 hack in unit tests

demeritcowboy:
Overview
Hack

Before
Hack

After
Still a little hack because being new mjwpaymentapi isn't in the "approved" list.

Technical Details
When initially trying to get running on drupal 10, various extensions needed the master version, not released. That's not true anymore, and, as of a few hours ago, the new stripe mjwpaymentapi doesn't need special version handling either.

Comments
Also remove the other hack for the parent webform which was from even earlier in the drupal 10 readiness attempts.


demeritcowboy:
Woohoo!

PR #925, opened 2023-11-21
Modify `fillOptions` to alphabetize dropdown list

brienneK:
Overview
----------------------------------------
This PR resolves the [issue](https://www.drupal.org/project/webform_civicrm/issues/3340709) of states not shown in alphabetical order.

Before
----------------------------------------
The browser re-sorts the JSON by id, so the options are not necessarily in alphabetical order. This is most apparent when the Country is set to United States, United Kingdom, India, or Bermuda.

Replication Steps
----------------------------------------
1. Create a webform with CiviCRM processing enabled
2. Include a set of address fields on Contact 1
3. Add *Country* and *State* to the form
4. Save and **View**
5. Note that when either United States, United Kingdom, India, or Bermuda are selected as the *Country*, the *State* drop down list is not in alphabetical order

After
----------------------------------------
Instead of sorting by id, the JavaScript now sorts the names alphabetically.

Technical Details
----------------------------------------
`fillOptions()` has been modified so that the data is an array instead of an object, which allows for sorting by the elements' values.

PR #934, opened 2023-12-28
Modify fillOptions to alphabetize dropdown list

MegaphoneJon:
Overview
This PR resolves the [issue](https://www.drupal.org/project/webform_civicrm/issues/3340709) of states not shown in alphabetical order.

Before
The browser re-sorts the JSON by id, so the options are not necessarily in alphabetical order. This is most apparent when the Country is set to United States, United Kingdom, India, or Bermuda.

To replicate:
*   Create a webform with CiviCRM processing enabled
*  Include a set of address fields on Contact 1
*  Add Country and State to the form
*  Save and View
* Note that when either United States, United Kingdom, India, or Bermuda are selected as the Country, the State drop down list is not in alphabetical order.

After
States/provinces are sorted alphabetically.

Technical Details
`fillOptions()` has been modified so that the data is an array instead of an object, which allows for sorting by the elements' values.

Comments
This is the fixed version of #925, which should be closed.  #925 doesn't keep a selected value when the function is called multiple times.

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -328,20 +328,21 @@ var wfCivi = (function (D, $, drupalSettings, once) {
   }
 
   function fillOptions(element, data) {
+    var sortedData = Object.entries(data).sort(([,a],[,b]) => a > b);
     var $el = $(element),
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you Jon and @briennekordis

PR #935, opened 2024-01-08
Return access denied if user is blocked from accessing the webform

jitendrapurohit:
Overview
Access Denied if user is blocked from accessing the webform.

Before
A hardcoded message is displayed if user is blocked on the webform. To test

- Create a webform.
- Enable `Block unknown user` from additional setting on civicrm tab.
- Access the webform as anonymous.
- A message is displayed on the top with few fields displayed too.

image

After
403 access denied is returned.

This also respects the access denied configuration on the webform. 

image

So admins can enter their custom message, etc for the blocked users.


KarinG:
Thank you @jitendrapurohit - test fail is being addressed in #936

PR #938, opened 2024-01-15
backport PR #830 to Drupal 7

MegaphoneJon:
Overview
Passing `NULL` to `strtolower()` generates a lot of log noise in PHP 8.  This occurs when you have relationship types that can use any contact type.  We fixed this in D8+ but not in D7.

Before
Log noise - `strtolower()` can't take `NULL` as argument 1.

After
No noise.


KarinG:
Backport of #830 👍

KarinG:
Thanks Jon!

PR #939, opened 2024-01-23
Set empty value for N/A.

puresyntax71:
Overview
When "Address" is enabled for a contact with _Country and State/Province_ element enabled, if user selects a country that has no state, this error is thrown upon form submission:

```
Drupal\Core\Entity\EntityStorageException: state_province_id is not a valid integer in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of /app/web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).
```

I think CiviCRM is expecting an integer although got "-" instead.

Before
Throws error when selecting "N/A" as the option for "State/Province" element.

After
No more error.

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -345,7 +345,7 @@ var wfCivi = (function (D, $, drupalSettings, once) {
       };
     }
     else {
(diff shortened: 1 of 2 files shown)


KarinG:
One of the fails is Stripe - re-running tests. 

image

PR #940, opened 2024-01-28
Skip Stripe tests

jitendrapurohit:
Overview
Skip stripe tests until keys are updated - https://lab.civicrm.org/extensions/stripe/-/issues/466


KarinG:
@mattwire - we've had to comment out the Stripe test for now. It can go right back in when you update the Stripe extension w/ new keys.

PR #941, opened 2024-01-29
Fix state display on preview

jitendrapurohit:
Overview
Fix state display on preview

Before
Existing Contact

After
Contact 1

Technical Details
The state has numerous options that we do not define as `#options` in the YAML configuration due to its daisy-chaining load based on country selection.

This PR adds a format function for label displays. I believe it should also fix the tokens @MegaphoneJon.

Comments
Drupal Ticket: https://www.drupal.org/project/webform_civicrm/issues/3410313

### DIFF
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -334,7 +334,7 @@ class CivicrmOptions extends OptionsBase {
       $state_id = $value;
     }
     else {
-      $state_id = $value['#plain_text'] ?? NULL;
+      $state_id = $value['#plain_text'] ?? $value['#markup'] ?? NULL;
     }
     if ($format === 'raw' || empty($state_id) || !is_numeric($state_id)) {
       return $value;
--- a/src/Plugin/WebformElement/CivicrmOptions.php
+++ b/src/Plugin/WebformElement/CivicrmOptions.php
@@ -321,4 +321,33 @@ class CivicrmOptions extends OptionsBase {
     }
   }
 
+  /**
+   * @inheritDoc
+   */
+  protected function format($type, array &$element, WebformSubmissionInterface $webform_submission, array $options = []) {
+    $value = parent::format($type, $element, $webform_submission, $options);
+    $format = $this->getItemFormat($element);
+    if (!str_ends_with($element['#form_key'], '_address_state_province_id')) {
+      return $value;
+    }
+    if ($type === 'Text') {
+      $state_id = $value;
+    }
+    else {
+      $state_id = $value['#plain_text'] ?? NULL;
+    }
+    if ($format === 'raw' || empty($state_id) || !is_numeric($state_id)) {
+      return $value;
+    }
+    $utils = \Drupal::service('webform_civicrm.utils');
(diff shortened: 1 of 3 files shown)


avallllovera:
Tested. It worked.

KarinG:
I kicked off the tests by making an edit to the main.yml file that I needed to make any way! :-)

MegaphoneJon:
My client is reporting that this doesn't fix the [webform_submission::values] token in emails.  I'll try to look into this myself first though, since I think this PR is most of the fix. I think we may just need to check if the value is numeric rather than whether the type is "Text".

src/Plugin/WebformElement/CivicrmOptions.php
```diff
@@ -321,4 +321,33 @@ public function validateConfigurationForm(array &$form, FormStateInterface $form
...
+    $value = parent::format($type, $element, $webform_submission, $options);
+    $format = $this->getItemFormat($element);
+    if (!str_ends_with($element['#form_key'], '_address_state_province_id')) {
+      return $value;
+    }
+    if ($type === 'Text') {
+      $state_id = $value;
+    }
+    else {
+      $state_id = $value['#plain_text'] ?? NULL;
```
MegaphoneJon: If you're sending an email with `state_province_id` fields, and the email is in HTML format, `$value['#plain_text']` doesn't exist.  I this we want `$state_id = $value['#plain_text'] ?? $value['#markup'] ?? NULL;`.

jitendrapurohit: @MegaphoneJon If i use the default body having`[webform_submission:values]` token to render all submission, it does work for me after the current PR. (Screenshot from email)

image

If i use a custom body with field tokens `[webform_submission:values: civicrm_1_contact_1_address_state_province_id]`, it still works and renders the state correctly in email

image

The debug value of $text is Html & $value does have `#plain_text` in this function -

image

I'm fine to include your change if `$value['#markup']` works for you. But curious to know your usecase since we have a [similar implementation](https://github.com/colemanw/webform_civicrm/blob/6.x/src/Plugin/WebformElement/CivicrmContact.php#L601) for Existing Contact, so i think `[webform_submission:values:civicrm_1_contact_1_contact_existing]` token might break for you too?

MegaphoneJon: Hi Jitendra,

That's not what I'm seeing.  Attached is a webform export that displays the state/province as an ID.

Selection_2143

[state id issue.yml.txt](https://github.com/colemanw/webform_civicrm/files/14240652/state.id.issue.yml.txt)

jitendrapurohit: Hmm, the attached webform still works fine on my end. Both Preview and email has state label instead of an id. Also, the debug values are same.

image

I believe I'll proceed with your suggestion since it appears to be a safe addition and works on your end :)

MegaphoneJon:
I've been running the current version of this patch in production for a few months now with no issues, it looks good.  I recommend removing the commented-out line in `WebformCivicrmPostProcess.php` but otherwise this LGTM.

jitendrapurohit:
ah thanks for pointing that out. Have removed the line.

KarinG:
Thank you all!

PR #943, opened 2024-01-31
Testing Irish names

KarinG:
Overview
A member of AWA reported not being able to make a payment using iATS.

Before
In the tests we're just using
```
    $billingValues = [
      'first_name' => 'Frederick',
      'last_name' => 'Pabst',
```

After
```
    $billingValues = [
      'first_name' => 'Frederick',
      'last_name' => "O'Pabst-Kelly",
```

I'm expecting this to fail. And when it does we know it needs to be addressed in the iATS Extension.


KarinG:
Ok that passed... I need a screenshot.

KarinG:
Ok that was the wrong test! Billing fields now updated for legacy and exporting a png.

PR #944, opened 2024-02-01
Support for participant count

MegaphoneJon:
Overview
In event price sets, a price option can have a "participant count".  E.g. purchasing a "table for 10" ticket will increment the participant count by 10 despite buying one ticket.  This PR adds support for participant counts in WFC.

Before
All tickets purchased count as one participant.

After
A "participant count" field exists This can be a numeric input, Computed Twig, etc.

The "Fee Amount" is *not* affected by the count.  If you want the fee amount to be "per participant", you'll use Twig.

Technical Details
Participant count requires a line item with a non-NULL `price_field_value_id`).  WFC has line items for paid events only.  We never populate `price_field_value_id`.

So:
* We introduce a new field "Participant Count" (in `AdminHelp.php` and `Fields.php`).
* We create a hidden price field exclusively for use by WFC events.  The `Utils.php` function returns that price field value, and will create it if it doesn't exist.
* When recording a contribution's line items, we now include the count.  We make changes to `qty` and `unit_price` to be count-aware (if used).  We include the required `line_total` and `price_field_value_id` so participant count is calculated correctly.
* When recording events, we check if it's a paid event.  If not, we create a line item tied to the participant record.

Comments
When testing, note that API4's "remaining participants" calculated field is [inaccurate for multiple reasons](https://lab.civicrm.org/dev/core/-/issues/4907).  To view an accurate count, call Api3 `event.get` and return underdocumented value `is_full` (see screenshot).

Selection_2130

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -199,6 +199,68 @@ class Utils implements UtilsInterface {
   }
 
   /**
+   *  Create a hidden Price Set that contains a Price Field Value with a `count` of one so that `qty` accurately reflects
+   * number of seats registered on submission
+   */
+  function wf_crm_get_participant_price_set() {
+    // Get the price_set_value id
+    $qtyPriceFieldValueId = civicrm_api4('PriceFieldValue', 'get', [
+      'checkPermissions' => FALSE,
+      'select' => ['id'],
+      'where' => [['name', '=', 'quantity_count_field_value']],
+    ])[0]['id'] ?? NULL;
+    // If the Price Set Value exsists, return it.
+    if ($qtyPriceFieldValueId) {
+      return $qtyPriceFieldValueId;
+    }
+    else {
+      // If it does not exist, create the price set, price field, and price field value (with a `count` of 1).
+      $activeFinancialtypeId = civicrm_api4('FinancialType', 'get', [
+        'checkPermissions' => FALSE,
+        'select' => ['id'],
+        'where' => [['is_active', '=', TRUE]],
+        'limit' => 1,
+      ])[0]['id'];
+      $qtyPriceSetId = civicrm_api4('PriceSet', 'create', [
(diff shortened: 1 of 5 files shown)


MegaphoneJon:
I'll address these failures tomorrow.

MegaphoneJon:
test this please

KarinG:
I just kicked off the test by pushing a trivial commit to your `participant-count-deux` branch.

MegaphoneJon:
I'll resolve the conflict here soon.  This should pass existing tests, but I have new tests that aren't all passing.  Event line items have always been a little inconsistent in WFC and I need to figure out how consistent they need to be made to support this.

KarinG:
Ah cool on the test! If needed we can bring this over the finish line in Montreal!

MegaphoneJon:
OK - this is now updated to create line items correctly in both "register all for same event" and "register each contact separately".

Previously - line items were created correctly only when "register each contact separately" was selected.  When "register all for same event" was selected, one line item was created for the entire transaction.  This threw off participant count calculations, which assume if there's a missing line item, the count is 1.  So registering 3 people for 4 seats each would take up 14 places, not 12 (since contact 1 would have a line item for 12 places, contacts 2 and 3 would have no line items).

My latest changes ensure that separate line items are created for each participant record.

src/Utils.php
```diff
@@ -198,6 +198,68 @@ function wf_crm_get_events($reg_options, $context) {
...
+          'count' => 1,
+          'amount' => 1,
+          'is_active' => TRUE,
+          'financial_type_id' => $activeFinancialtypeId,
+        ],
+      ])[0]['id'];
+      return $qtyPriceFieldValueId;
+    }
+  }
+
```
jitendrapurohit, line 262: Should this be moved to an upgrade hook?

PR #947, opened 2024-02-19
D9 - Ensure new events are loaded as default on the event field

jitendrapurohit:
Overview
URL Event is not loaded if event is added after configuring the webform.

Before
To replicate:

- Create a webform with events = `User Select`.
- Enable `Allow events to be autoloaded from URL` config from Events tab.
- Load the webform with query string `event1=<id>`. Works fine.
- Create a new event in civicrm.
- Load the webform with query string `event1=<id>` where <id> is the event id of the newly created event. The event is not loaded as default on the webform.

After
Event is loaded as default.

Technical Details
Ensure we have the latest options if we have  `civicrm_live_options` enabled for the webform element.

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -554,9 +554,13 @@ abstract class WebformCivicrmBase {
   protected function getExposedOptions($field_key, $exclude = []) {
     $field = $this->getComponent($field_key);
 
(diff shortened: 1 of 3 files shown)


KarinG:
Cool - tested! This is actually a feature (event autoload from URL) that I had never used :-)

jitendrapurohit:
Thanks for testing @KarinG :)

PR #949, opened 2024-02-20
Allow test mode in webform test

puresyntax71:
Overview
This is in relation to #373. I'm not sure if it's a good implementation or not though. What it does is switch to test mode once the "Test" feature is used in webform_civicrm. It does however remove the "Payment Processor Mode".

> Would you like to create a PR for D8/9WFC that would make Payment Processor Mode a -user select-? That would give you options to quickly switch LIVE to TEST;

Or would this be a better implementation?

A use case for this one is to be able to use test payment processor in dev instances while using live on production and that the webform configurations can be sync'ed.

Before
Settings are forced to use only either live or test transactions. Using test mode in dev creates configuration differences when live mode is used for prod.

After
Payment Process Mode selection is removed and end users can either use the normal webform to use live transactions or the "Test" mode in webform to use test transactions.

Comments
I think I missed to either update or remove a few more things although the change here seem to work just fine.


MegaphoneJon:
I have no comment on this approach, but wanted to say that Webform Variants are a good way to create a syncable test version of a page.

PR #952, opened 2024-02-25
D9 - Add Receive Date field to contribution tab

jitendrapurohit:
Overview
Add Date Received field on the Contribution tab

Before
No way to allow users to choose receive date for their pay later contributions.

**Use-case**: A member selects pay later and admin is wondering when? Next month, next quarter?
If you do a pay later in webform -> it lands on today:
Letting Admins add Contribution Date (Date Received) to a webform (in general) would allow users to not just say Pay Later, but also say When. 

After
Receive Date field added on the contribution tab.

Comments
@KarinG


KarinG:
I've tested this on one of my sites:

image

image

PR #956, opened 2024-03-17
Revert "Skip Stripe tests"

KarinG:
Reverts colemanw/webform_civicrm#940

### DIFF
--- a/tests/src/FunctionalJavascript/StripeTest.php
+++ b/tests/src/FunctionalJavascript/StripeTest.php
@@ -38,7 +38,6 @@ final class StripeTest extends WebformCivicrmTestBase {
    * Verifies the payment with 1 contribution and 2 line item amounts.
    */
   public function testSubmitContribution() {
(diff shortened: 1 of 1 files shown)


mattwire:
See https://lab.civicrm.org/extensions/stripe/-/commit/0753e8640db342d504eebfacf4eb3e81bba7cd18

KarinG:
Tests pending

KarinG:
Stripe test is still failing - WebDriver\Exception\UnexpectedAlertOpen: unexpected alert open: {Alert text :  Expired API Key provided: pk_test_k2******************RaYh}

mattwire:
So that's somehow trying to use the old code/keys before the above commit!

KarinG:
Is the above commit in a release yet?

KarinG:
Actually - (and possibly) also it's testing against the Drupal/CiviCRM versions at the time of this original revert. So instead of reverting we need a new PR to test against the current D10.1.*, D10.2.* and D10.3.* (i.e. D11).

PR #957, opened 2024-03-21
[NFC] Deprecated call to FinancialAccount::add

demeritcowboy:
Overview
See https://github.com/civicrm/civicrm-core/blob/61472fd6c819d2fe3d6b3311085fec0883119ef1/CRM/Financial/BAO/FinancialAccount.php#L55

It comes up at the bottom of every test run.


KarinG:
Thank you Dave!

PR #960, opened 2024-04-23
D10 - Fatal error on empty group submission if configured as public group in settings

jitendrapurohit:
Overview
Fatal error on empty group submission if configured as public group in settings

Before
To replicate

- Create a webform with group = Public Group
- Submit the form as anonymous user without enabling any options for the group field.
- Error
>Drupal\Core\Entity\EntityStorageException: 'public_groups' is not a valid option for field group_id in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).

After
Fixed.

Technical Details
Unsetting $val works fine for logged-in or authorized users because it assigns it to $this->data only if the contact is logged in - https://github.com/colemanw/webform_civicrm/blob/6.x/src/WebformCivicrmPostProcess.php#L2579-L2583

For anonymous users, the key remains in `$this->data`, leading to an error. Removing it immediately fixes the error.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2504,7 +2504,7 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
           if (substr($name, 0, 6) === 'custom' || ($table == 'other' && in_array($name, ['group', 'tag']))) {
             $val = array_filter($val);
             if ($name === 'group') {
-              unset($val['public_groups']);
+              unset($val['public_groups'], $this->data[$ent][$c][$table][$n][$name]['public_groups']);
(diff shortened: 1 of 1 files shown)


petednz:
Can you pls add a test

rikvrede:
Can someone please merge this? I've been bashing my head against this error for a while, and ultimately ended up creating a stupid new public group "No thanks" just so people could submit the form.
Only by reading this PR did I realize I could solve it by not selecting the "public groups" option.

KarinG:
It was waiting on a test. And is not mergeable now (conflict). Have Converted to Draft.

PR #961, opened 2024-05-05
D9/10 - Add test for ajax payment using iats

jitendrapurohit:
Overview
Test cover to assert iATS payment on AJAX webform.

Before
No test.

After
Test added.

Comments
@KarinG


KarinG:
Great - merged!

PR #962, opened 2024-05-07
D9/10 - Fix display of contact fields, if hidden in settings

jitendrapurohit:
Overview
Contact field is displayed even if hidden in settings.

Before
To replicate:
- Set the type of second contact to Organization and enable Org Name field.
- Open the Build form, edit the 2nd Existing Contact element:
-- Form Widget = Select
-- Field to Lock = Name
-- Locked Fields should be = Hidden
- Visit the webform page.
- Org Name element is still displayed.
image
- Choose an org name, the field is hidden correctly.
- After reverting the option back to `Choose Existing`, the field remains hidden correctly.
 image
- The issue appears to occur only when the page is visited or loaded.

After
Fixed. The field is hidden on page load too.

Technical Details
triggers the change event on page load.

### DIFF
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -43,7 +43,7 @@
             true,
             field.data('form-defaults'),
           );
-        });
+        }).change();
 
         //In case of error, highlight the token-input field.
         if (field.hasClass('error')) {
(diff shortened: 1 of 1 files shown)


jitendrapurohit:
fixed in https://github.com/colemanw/webform_civicrm/pull/1008.

PR #967, opened 2024-05-17
Support GoCardless

artfulrobot:
Overview

A hack to allow webforms to setup GoCardless subscriptions.

Before

Using webform for a recurring GoCardless payment doesn't work.

After

It does work.

Technical Details

This has long been filed as a bug against the GoCardless ext: https://lab.civicrm.org/extensions/gocardless/-/issues/32

I suspect that WebForm also won't support other processors that use billing mode 4/NOTIFY.

This is a hack because it looks like webform's payment processing needs a big rewrite, and this is not that. I note that it already contains custom code for another processor, so while this is an anti-pattern in the long term, this is not designed for the long term.

### DIFF
--- a/includes/wf_crm_webform_postprocess.inc
+++ b/includes/wf_crm_webform_postprocess.inc
@@ -17,6 +17,13 @@ class wf_crm_webform_postprocess extends wf_crm_webform_base {
   private $billing_params = [];
   private $totalContribution = 0;
   private $contributionIsIncomplete = FALSE;
+
+  /**
+   * Is GoCardless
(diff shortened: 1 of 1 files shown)


KarinG:
The D7 branch has been maintenance only since D7 EOL was announced. If you'd like us to consider this, please put up a PR for the 6.x branch and once tests are passing we can look at a possible backport. We have extensive automated functional testing on Github Actions running on 4 combinations D10 x CiviCRM (the 6.x branch). That needs to run first.

PR #969, opened 2024-05-20
drupal 11 readiness

demeritcowboy:
Overview
Small test script change that can be done now.

Before
* Silent fail where nothing seems to happen during phpunit run and it just skips the test.
* Ignoration (is that a word?) of BROWSER_OUTPUT_DIRECTORY fixed in core. Now it actually uses the variable so it needs to match where the artifact upload step is looking for the files.

After
* phpunit runs the test
* Unignoration of BROWSER_OUTPUT_DIRECTORY

Technical Details
In phpunit10 you have to give it the argument `--display-skipped` to tell you why it skipped tests. The error was `DevToolsActivePort file doesn't exist) (The process started from chrome location /opt/google/chrome/google-chrome is no longer running`. This turned out to be https://www.drupal.org/node/3240792 - see description API Changes where you now have to use goog:chromeOptions. (Note the deprecation notice does not appear in 11 since they removed the notice, but 10.3 just went beta on the 17th, so that notice wasn't showing up anywhere.)

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -165,15 +165,15 @@ jobs:
         run: php -S 127.0.0.1:8080 -t ~/drupal/web &
       - name: Run PHPUnit
         run: |
-          mkdir $BROWSERTEST_OUTPUT_DIRECTORY
+          mkdir -p $BROWSERTEST_OUTPUT_DIRECTORY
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you Dave!

PR #972, opened 2024-05-24
Don't search related tables when filtering a contact reference field.

jmcclelland:
Overview

If you have an "Existing Contact" field, and, in the Builder tab, in the field definition's "Form display" section you configure the "Contact Display Fields" to display the State/Province, you will not get accurate results when entering a string to filter the results by. The field will display all results.

The problem is that the code builds the WHERE clause using every field entered as a display field. However, perhaps because the contact tables are LEFT JOIN'ed and the WHERE clause is an OR clause, the query somehow fails to filter the given string at all.

Since it's unlikely and even unexpected to have your search term applied to the address, email or phone tables, this patch simply removes that clause.

Before
Contact reference fields that include the "State/Province" as a display field fail to filter properly.

After
Contact reference fields, regardless of the display fields configured, are properly filtered.

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -148,6 +148,12 @@ class ContactComponent implements ContactComponentInterface {
     if ($str) {
       $searchFields = [];
       foreach ($display_fields as $fld) {
+        // Don't search related tables - not intuitive and also when searching a
+        // :label version of a related record, the LEFT join causes CiviCRM to
+        // return all records.
+        if (preg_match('/^(address|phone|email)/', $fld)) {
+          continue;
+        }
         $searchFields[] = [$fld, 'CONTAINS', $str];
       }
       $params['where'][] = ['OR', $searchFields];
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -103,9 +103,9 @@ class ContactComponent implements ContactComponentInterface {
       'email' => ['email', 'email'],
       'phone' => ['phone', 'phone'],
       'city' => ['address', 'city'],
-      'state_province' => ['address', 'state_province_id:label'],
-      'country' => ['address', 'country_id:label'],
-      'county' => ['address', 'county_id:label'],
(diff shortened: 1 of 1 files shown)


KarinG:
Thank you Jamie - I've kicked off our tests.

jitendrapurohit:
@jmcclelland Perhaps, this requires a discussion. I think that if a text is part of a dropdown option, searching for any portion of that text should filter the results accordingly. For instance, searching for "Auck" should show Auckland contacts in the dropdown. Do you agree?

image

With this PR included, the above would return an empty result.

> the query somehow fails to filter the given string at all.

Yes, can replicate that too, and i think its due to the API4 not filtering on the `address.state_province_id:name` clause. If i add an extra join to add state table, it filters the result correctly. 

Something like https://github.com/colemanw/webform_civicrm/commit/e0ec394156ce4ae77d4fc86082e59f9bb10a58db would be able to fix the missing part. Could you please confirm?

jmcclelland:
Thanks for the tip @jitendrapurohit  - your change both fixes the problem and preserves the ability to search on address fields. Although it might be unexpected to have your search term applied to an address field - I think you are right that some people might want that and expect it to work. And since all fields that are searched are displayed, it should be quite obvious what is happening if you did not intent for it.

I just made two small changes from your fix:

1. Change the field names from "state_province_id.name" to "state_province.name" so they match the join table names.
2. I added an explicit join field when adding the tables. It doesn't seem necessary for state province or county, but when I combined state province and country I got an error - I think because of the country_id field in the state province table - CiviCRM is not sure which table to join on.

Thanks for your help with this one.

jmcclelland:
@KarinG Hi Karen - Any chance we could get this merged? I just upgraded but forgot this patch and ended up with a buggy existing contact search. Thank you!!

KarinG:
Yes - absolutely!

jmcclelland:
@KarinG Thank you so much!!

PR #974, opened 2024-05-30
D11 support

jackrabbithanna:
Overview
Changes to support Drupal 11

Technical Details
Drupal deprecation code analysis provides this output for deprecations needing to be changed in D11
```
 ------ --------------------------------------------------------- 
  Line   src/WebformCivicrmBase.php                               
 ------ --------------------------------------------------------- 
  825    Call to deprecated function file_icon_class():           
         in drupal:10.3.0 and is removed from drupal:11.0.0. Use  
           \Drupal\file\IconMimeTypes::getIconClass() instead.    
 ------ --------------------------------------------------------- 

```
Comments
Drupal change record: https://www.drupal.org/node/3411269

To support D9 and D11, we have to check if function exists, and then use full object path. 
When D9 support is dropped, should add a "use" statement and remove the conditional

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -821,11 +821,17 @@ abstract class WebformCivicrmBase {
     }
     $file = $this->utils->wf_crm_apivalues('Attachment', 'get', $val);
     if (!empty($file[$val])) {
+      if (function_exists('file_icon_class')) {
+        $icon_class = file_icon_class($file[$val]['mime_type']);
+      }
+      else {
+        $icon_class = \Drupal\file\IconMimeTypes::getIconClass($file[$val]['mime_type']);
+      }
       return [
         'data_type' => 'File',
         'name' => $file[$val]['name'],
         'file_url'=> $file[$val]['url'],
-        'icon' => file_icon_class($file[$val]['mime_type']),
+        'icon' => $icon_class,
       ];
     }
     return NULL;
--- a/webform_civicrm.info.yml
+++ b/webform_civicrm.info.yml
@@ -1,7 +1,7 @@
 name: 'Webform CiviCRM'
 type: module
 description: 'Webform CiviCRM Integration'
-core_version_requirement: ^9.4 || ^10
+core_version_requirement: ^9.4 || ^10 || ^11
 package: 'CiviCRM'
 dependencies:
   - drupal:webform


jackrabbithanna:
I've created this composer template to assist with development: https://github.com/jackrabbithanna/drupal11-civicrm-proto

Uses recently merged D11 support in CiviCRM Core
Retrieves the Webform module with Drupal lenient, and patches it.

demeritcowboy:
Thanks for the PR. It should get merged once the tests are running. They don't run yet because the parent webform even patched or with "lenient" doesn't install during tests. Something about ckeditor schema.

And looks like the function_exists is needed in 10.2 since the replacement doesn't even exist until 10.3.

stesi561:
Are we at a point we can drop 10.2 support now that it's eol?

eileenmcnaughton:
@demeritcowboy - @jackrabbithanna is using this in production with a bunch of sites with just this patch - it would be great to get it merged - it looks like you want your patch merged first??

demeritcowboy:
There's a timing issue:

* there's a desire to NOT advertise that this module supports 11 when the parent webform does not have an official release supporting 11 yet (you currently have to install a beta of the parent webform)
* there are some unexplained javascript errors preventing being able to run tests on 11 fully. There's some hope those will be resolved by the parent webform.
* I think there's a desire to do a release of this module soon for other reasons, so merging would require some extra branch administration.

Having said that, this PR is sufficient to get something usable running if you install a beta of the parent module and then apply this patch.

jackrabbithanna:
Its a tiny patch to make it compatible. People should know they can easily be doing new sites with D11 instead of D10 and avoid an upgrade later. Its more beneficial for the community to have the option than to not, beta or alpha.

KarinG:
My current thinking:

I'm going to issue one more WFC release without it. I've been trying to find time to perhaps add more test coverage to some new membership functionality I added a few weeks back.

Then I'll merge the D11 work.

Folks who edit composer.json to allow for beta modules will know how to pull in D11 WFC using git. 

When the parent module is no longer beta then I'll issue a release.

jackrabbithanna:
For people looking at this, its possible to use Drupal Lenient to pull in Webform CiviCRM with the patch.
Reference: https://www.drupal.org/docs/develop/using-composer/using-the-drupal-lenient-composer-plugin

```
composer require mglaman/composer-drupal-lenient
composer config --merge --json extra.drupal-lenient.allowed-list '["drupal/webform_civicrm"]'
```

Add to the "extra" section of your composer.json
```
"patches": {
            "drupal/webform_civicrm": {
                "d11 fix": "https://patch-diff.githubusercontent.com/raw/colemanw/webform_civicrm/pull/974.patch"
            }
```
Example: https://github.com/jackrabbithanna/drupal11-civicrm-proto/blob/master/composer.json#L120

francescbassas:
We haven't been able to install it with Drupal Lenient. We'll try to test it with a patch applied in a fork.
+1 to have the new version with the patch applied soon.

KarinG:
Closing as this was merged as part of
https://github.com/colemanw/webform_civicrm/pull/1052
Thank you Mark!

PR #975, opened 2024-06-04
Update for label change in core

demeritcowboy:
Overview
Needed after https://github.com/civicrm/civicrm-core/pull/30188


KarinG:
Awesome - thanks Dave

PR #976, opened 2024-06-11
d10 - Fix default load of yesno custom field

jitendrapurohit:
Overview
Fix default load of yesno custom field

Before
To replicate:
- Create a custom set and a field of data type yesno in CiviCRM.
- Add this field to a webform.
- Provide a default value to the webform element. Eg Yes.
- Load the form with the default cid
- Choose custom field value = No
image
- Submit the form. Notice the value is correctly saved to civicrm.
- Load the form again.
- The value `Yes` is loaded by default instead of `No` existing on the contact.

After
Correct default value from the contact is loaded on the form.


KarinG:
This too could use a follow up test to ensure it does not regress.

PR #978, opened 2024-06-15
Undeclared vars mostly

demeritcowboy:
Overview
Pulling out some test-only commits from the drupal 11 prep that applies generally, e.g. php8 undeclared vars.


KarinG:
Thanks Dave!

PR #980, opened 2024-06-22
D10: Update soft credit implementation

jitendrapurohit:
Overview
Revise soft credit implementation to make it in line with latest CiviCRM.

Before
Code still uses the 4.x civicrm format of `honor_type_id` & `honor_contact_id`

image

After
Soft Credit section added for admins to configure type and contact.

image

Comments
The PR includes an upgrade code to update all existing webforms using honor_type_id, assigning it to the soft_credit_type_id element as needed.

The only limitation is that it does not support assigning multiple soft_credit_type values to contacts. This can be added in a follow-up if required.

With this PR, multiple contact IDs can be selected for a single soft credit type.

### DIFF
--- a/js/webform_civicrm_admin.js
+++ b/js/webform_civicrm_admin.js
@@ -500,25 +500,6 @@ var wfCiviAdmin = (function (D, $, once) {
         changeContactLabel.call(this);
       });
 
-      // Contribution honoree fields
-      $(once('crm-contrib', 'select[name$=contribution_honor_contact_id]', context)).change(function() {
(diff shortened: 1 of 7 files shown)


KarinG:
One of the key reasons for doing this is to make all soft credit types available:

image

image

KarinG:
@MegaphoneJon - Jon: if/when you have capacity, would you please have a look at this PR?

KarinG:
Just a note that I have tested this PR and am using it on an actual project.

PR #982, opened 2024-06-24
update matrix 2

demeritcowboy:
Overview
* Some undeclared vars from iats that are merged in master but not released yet.
* Two more undeclared.
* Most people will be on 5.74 currently, and then next month the esr switches to 5.75, so have updated to test 5.74 and one less of 5.69.


KarinG:
👍

PR #985, opened 2024-07-23
Add is email receipt as doPayment parameter

francescbassas:
Overview
This little modification allows transfer if the Webform takes care of sending the receipt to doPayment function.

This can allow, for example, payment gateways like [Redsys](https://lab.civicrm.org/extensions/redsys) to decide whether to manage the sending of receipts or delegate it to the webform.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -321,12 +321,14 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
       ->fields($record)
       ->execute();
 
+    $isEmailReceipt = wf_crm_aval($this->data, "receipt:number_number_of_receipt", FALSE);
+    
     // Calling an IPN payment processor will result in a redirect so this happens after everything else
     if ($this->contributionIsIncomplete && !$this->contributionIsPayLater && !empty($this->ent['contribution'][1]['id']) && !$this->submission->isDraft()) {
 //      webform_submission_send_mail($this->node, $this->submission);
-      $this->submitIPNPayment();
+      $this->submitIPNPayment($isEmailReceipt);
     }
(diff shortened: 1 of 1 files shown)


src/WebformCivicrmPostProcess.php
```diff
@@ -2160,6 +2164,7 @@ private function submitIPNPayment() {
       $params['cancelURL'] = $this->getIpnRedirectUrl('cancel');
     }
 
+    $params['is_email_receipt'] = $isEmailReceipt;
```
jitendrapurohit, line 2167: Can we simply use 

    $params['is_email_receipt'] = wf_crm_aval($this->data, "receipt:number_number_of_receipt", FALSE);

instead of sending `$isEmailReceipt` as a param argument?

francescbassas: Sure! Like this? https://github.com/colemanw/webform_civicrm/pull/987

mattwire:
The problem with this is that it's completely unsupported to pass in a parameter for email receipt to `doPayment()`. I'm not actually aware of any payment processors that would support that?

I do understand that there is a problem to solve - we can set email_receipt on contribution_page or contribution_recur but not on the contribution itself but I think that needs to be solved in CiviCRM first.
Also, the receipt would not usually be sent from within `doPayment()` as it would be handled when recording the payment which would be done when the IPN code is invoked.

I think we'd need support for a receipt parameter in CiviCRM first (maybe an additional field in `civicrm_contribution` that could be set?).

francescbassas:
Thanks @mattwire

With this, I'm trying to resolve the [Redsys](https://lab.civicrm.org/extensions/redsys) payment processor casuistry when its IPN it's invoked. The idea with the change is that the payment processor can distinguish when the webform is already managing the sending of the receipt and, therefore, not send the one that is enabled by default. I don't know if there is any other payment processor that could also benefit.

cc @sluc23

PR #996, opened 2024-09-02
unit tests - see if yes-sandbox solves fails in latest chrome/chromedriver

demeritcowboy:
Overview
_A brief description of the pull request. Try to keep it non-technical._

Before
_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

After
_What has been changed. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

Technical Details

Comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -170,7 +170,7 @@ jobs:
           git apply 458.diff
       - uses: nanasess/setup-chromedriver@master
       - name: Run chromedriver
-        run: chromedriver &
+        run: chromedriver --port=9515 &
       - name: Run php built-in server
         run: php -S 127.0.0.1:8080 -t ~/drupal/web &
       - name: Run PHPUnit
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -182,7 +182,7 @@ jobs:
           SYMFONY_DEPRECATIONS_HELPER: 999999
           SIMPLETEST_DB: mysql://root:@127.0.0.1:${{ job.services.mysql.ports[3306] }}/db
           SIMPLETEST_BASE_URL: http://127.0.0.1:8080
-          MINK_DRIVER_ARGS_WEBDRIVER: '["chrome", {"browserName":"chrome","goog:chromeOptions":{"args":["--disable-gpu", "--no-sandbox", "--headless"]}}, "http://127.0.0.1:9515"]'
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Nope, still failing in the same way.

bsilvern:
@demeritcowboy Just wanted to give you a heads-up on a Chromedriver bug that was causing failures for me at `ContactSubmissionTest.php --filter testSubmitWebform#2` on the website = 'https://example.com' assertion. I'm using Chromedriver 127.0.6533.88. Haven't tested other versions.

The issue in a nutshell is that if it's possible to scroll the window up (current scroll pos is not at the top), then when SendKeys sends a slash character, it is interpreted as a PageUp key. See my full description [here](https://issuetracker.google.com/u/1/issues/42323689).

My workaround has been to patch vendor/behat/mink/src/Session.php:
```php
    public function resizeWindow($width, $height, $name = null)
    {
      //$this->driver->resizeWindow($width, $height, $name);
      // Avoid SendKeys slash bug: https://issuetracker.google.com/u/1/issues/42323689.
      // If not in headless mode, the dimensions below will be limited by actual screen resolution.
      $this->driver->resizeWindow(1600, 2000, $name); 
    }
```
By making the browser window larger, the page does not scroll down when the form is filled, and the slash is therefore interpreted correctly.

demeritcowboy:
Thanks that might be helpful for some of the other fails that happen. This one was about trying to track down what had changed in chrome/chromedriver 128. It's the same issue as https://www.drupal.org/project/drupal/issues/3471113, and their temporary solution was to [lock the version at 127](https://git.drupalcode.org/project/drupal/-/commit/9734ca5f39d1546198d94e7a80909b521f062b96), so we've done the same for now.

We could try such a patch and see if it cuts down on the other intermittent fails.

bsilvern:
Good to know about v128!

- When I run the webform_civicrm test suite without my patch I get only two errors, both in ContactSubmissionTest::testSubmitWebform (same error in data sets 2 and 4).

- See the last item in [this PR](https://github.com/colemanw/webform_civicrm/pull/991/files#diff-3db5493d1b66e3133832b40c91cdb209b59cdc20055daebf3e414b45f113fa88) `// Wait for ajax message...` which resolves a failure I was getting in ContactRelationshipTest::testSubTypeRelationship.

- Hack of the day: Inserting a return at the top of web/core/misc/drupal.js::Drupal.deprecationError() reduces the full webform_civicrm test execution time (single configuration) from 32 to 22 minutes on my machine!

PR #1000, opened 2024-09-06
[WIP] see if helps with intermittent fails

demeritcowboy:
Overview
See https://github.com/colemanw/webform_civicrm/pull/996#issuecomment-2334841619


KarinG:
Dave I think this can be closed?

demeritcowboy:
Yes it didn't seem to make a difference.

PR #1008, opened 2024-09-16
Correct issue: existing_contact select widget fails to lock fields

bsilvern:
Overview
Locked fields are not actually locked when an existing_contact element is configured as a Select widget.

Before
Locked fields with an existing_contact Select widget were not actually locked.

After
* Locked fields with an existing_contact Select widget are locked as demonstrated by `ExistingContactElementTest::testNextPrevSaveLoad()` line 640.
* The test introduced in the previous PR in this series completes without error.

Technical Details

Comments
The tests are unchanged from the previous PR in this series.

### DIFF
--- a/js/webform_civicrm_contact.js
+++ b/js/webform_civicrm_contact.js
@@ -17,15 +17,19 @@
             searchingText: "Searching...",
             enableHTML: true
           };
-          wfCivi.existingInit(
-            field,
-            field.data('civicrm-contact'),
-            field.data('form-id'),
-            autocompleteUrl,
-            toHide,
-            tokenValues
-          );
         }
+        else {
+          var tokenValues = false;
+        }
+        
+        wfCivi.existingInit(
+          field,
+          field.data('civicrm-contact'),
+          field.data('form-id'),
+          autocompleteUrl,
+          toHide,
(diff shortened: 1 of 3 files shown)


jitendrapurohit (approved):

KarinG:
Jitendra approved - going to merge this now so that they will be in the next release.

KarinG:
Bob -> note we ended up skipping this test for D10.3+ 

https://github.com/colemanw/webform_civicrm/blob/6.x/tests/src/FunctionalJavascript/ExistingContactElementTest.php#L532-L536

Because in D10.3 we get:

`$webform->get('elements') is {  }`

We tried a couple of things e.g.
`$elements = $webform->getElementsDecodedAndFlattened();`

But could not get it to work.

Also not directly related but relevant for future: 
`966x: The "yaml_parser_class" setting is deprecated in drupal:10.3.0. This setting should be removed from the settings file, since its usage has been removed. See https://www.drupal.org/node/3415489. which means Drupal\Core\Serialization\Yaml is deprecated.`

bsilvern:
@KarinG OK, thanks. I'll update the test if I can find a way to support D10.3+.

PR #1009, opened 2024-09-17
Rename uploaded files using webform filename pattern (and tokens)

sluc23:
Overview
Webform hast the feature of setting specific filenames for `File` type fields. This allows to rename in the server uploaded files with specific names or using tokens. Description [here](https://www.drupal.org/node/2943067)
This feature does not work in Webform CiviCRM, the uploaded files are stored in with the original filename.. why? because how processes are executed in this order:

1. On webform submission (or before if uploads are inline), the file is uploaded to the server, usually in the `private://` location with original filename
2. `webform_civicrm` handler is executed first and and copies the filename from the first private location to `files/civicrm/custom` with the original name.  
ref: https://github.com/colemanw/webform_civicrm/blob/6.2.5/src/WebformCivicrmBase.php#L782
3. After that the webform code is executed and moves the file from private to final destination and applies the renaming (and replacing tokens if needed).  
ref: https://git.drupalcode.org/project/webform/-/blob/6.2.2/src/Plugin/WebformElement/WebformManagedFileBase.php?ref_type=tags#L1217
This is where webform calculates the final filename based on this pattern:
ref: https://git.drupalcode.org/project/webform/-/blob/6.2.2/src/Plugin/WebformElement/WebformManagedFileBase.php?ref_type=tags#L1249

So we need this feature to work in `webform_civicrm`, and after some research I found this solution

Before
Uploaded files are always saved with the original local name

After
Uploaded files are saved, following the filename pattern defined in the field settings

Technical Details
This is a solution I came up with, it could be better alternatives, but basically I had to copy how to calculate the filename from filename pattern in webform module.
`CivicrmWebformHandler` needs to have a reference to the `webform.token_manager` service to do the calculation of filename.. this could be beneficial for using Drupal tokens in other sections of the module's handler

Comments
Open to suggestions

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -48,6 +48,11 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
   private $database;
 
   /**
+   * @var \Drupal\webform_civicrm\Plugin\WebformHandler
+   */
+  private $handler;
+
+  /**
    * @var \Drupal\webform\WebformSubmissionInterface
    */
   private $submission;
@@ -79,10 +84,10 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
 
     $handler_collection = $this->node->getHandlers('webform_civicrm');
     $instance_ids = $handler_collection->getInstanceIds();
-    $handler = $handler_collection->get(reset($instance_ids));
+    $this->handler = $handler_collection->get(reset($instance_ids));
     $this->database = \Drupal::database();
 
-    $this->settings = $handler->getConfiguration()['settings'];
+    $this->settings = $this->handler->getConfiguration()['settings'];
     $this->data = $this->settings['data'];
     $this->enabled = $this->utils->wf_crm_enabled_fields($this->node);
     $this->all_fields = $this->utils->wf_crm_get_fields();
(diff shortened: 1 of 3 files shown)


KarinG:
Test fails are unrelated. I'll pull in the PR and test it this week. Thank you @sluc23 👍

jitendrapurohit:
Yes i think saving files in the presave sounds fine to me borrowing few code from webform module. But i think the approach can be made better if we can do the following? Eg

- Add a custom service for civicrm file managerment, eg CivicrmFileManager.
- Inject file_system and token manager service to it.
- Move `saveDrupalFileToCivi()` to it?

Thoughts @sluc23 ?

sluc23:
@jitendrapurohit I think it is a good idea too.. but in that case I think that refactor is much complex that what I'm able to achieve.. that's a bigger architecture modification. Creating a new service in the module will require more tests, check that does not affect other sections of the module source code, taking important decisions, etc..

In the case you want to move to that direction i can offer my self to review/test the solution, but not to code it..

KarinG:
I'm ok with the simpler approach here (for now). Thank you @sluc23

@jitendrapurohit did you have a chance the functionality? If yes - then when can merge this.

jitendrapurohit:
yes have tested and it works as per the functionality is concerned. Merging. Thanks @sluc23 👍

src/WebformCivicrmPostProcess.php
```diff
@@ -2523,7 +2528,11 @@ private function fillDataFromSubmission() {
           }
         }
         elseif ($dataType == 'File') {
-          if (empty($val[0]) || !($val = $this->saveDrupalFileToCivi($val[0]))) {
+          // Replace filename (with tokens) if set.
+          if (isset($component['#file_name']) && $component['#file_name']) {
+            $newFilename = $this->handler->tokenManager->replace($component['#file_name'], $this->submission);
+          }
+          if (empty($val[0]) || !($val = $this->saveDrupalFileToCivi($val[0], $newFilename))) {
```
stesi561, line 2535: I've just hit an error here with a vanilla install and I think this is the problem. It assumes we set the new Filename above. But I'm getting a warning "Undefined variable $newFilename " which suggests we didn't.

PR #1017, opened 2024-10-26
Fix calculation of tax on line item amount

jitendrapurohit:
Overview
Fix calculation of tax on line item amount if the field is displayed on payment page.

Before
To replicate:
- Create a webform with contribution amount and line item amount.
- Ensure financial type set on both the field has 15% GST applied to it.
- Ensure both the field are set to display on second page (payment page).
- Submit the webform.
- On payment table, only contribution amount row displays 15% tax. 
- Line item amount is displayed without tax amount.

image

After
Fixed

image

Technical Details
Consider tax amount for line item if its included in the webform.


KarinG:
Thank you @jitendrapurohit 👍

KarinG:
It would be good if Fuzion could add a follow up test to ensure this does not regress?

PR #1019, opened 2024-10-29
Fix check permissions in civicrm_api call

peetCreative:
Overview

I found that with my setup searching for existing contacts by city did not work.
After fideling around I found that checkpermissions was not set `FALSE` before calling the civicrm_api4.

Technical Details
------------------------

This was due to the behaviour of `+=` for php arrays, which won't overwrite existing elements. 
As the element `$params['checkPermissions']` was set true earlier, the element was not overwritten by `+=` and so the explicit join did not work.

I added a quick fix, setting the element in the common way. Also one other function.

Comments
----------------

However I found, that the extension is using `+=` quit often. But I wonder if there is a need for it?

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -632,9 +632,7 @@ class Utils implements UtilsInterface {
     if (!$entity) {
       return [];
     }
-    $params += [
-      'checkPermissions' => FALSE,
-    ];
+    $params['checkPermissions'] = FALSE;
     $result = civicrm_api4($entity, $operation, $params, $index);
     return $result;
   }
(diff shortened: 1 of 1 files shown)


KarinG:
5.79 RC build crashed but not due to this PR. Merging (tests will run again on merge).

KarinG:
Thanks for this too 👍

MegaphoneJon:
This changed the intent and behavior of the function.  Karin, I'll contact you separately re: this.

peetCreative:
@MegaphoneJon 
I did not have time to look at your update. But the revert broke the functionality of the existing contacts feature as I described initially.
So what is the intent and behavior of the function?

My problem seems to be that the Drupal Role which should use the webform has not the correct permission.
I set the permission: _CiviCRM: view all contacts._
However as soon as address table is joined then the permission checking fails.

MegaphoneJon:
Hi @peetCreative - I didn't post more details because it introduced a security vulnerability and I didn't want to post the details publicly before there was an opportunity to get a patched version available.

This removes the "checkPermissions" from arrays where it was previously set to TRUE.  I'm sure this has farther-reaching effects, but at a minimum, if you allow URL autofill for existing contacts (`cid=1231`), anyone can view any contact, and the checksum doesn't matter.

peetCreative:
Thanks for the explenation.

PR #1021, opened 2024-11-13
Fix test fails on drupal 10.3.7

demeritcowboy:
Overview
Clean version of https://github.com/colemanw/webform_civicrm/pull/1020

Before
Something changed in 10.3.7 but not sure exactly what, just the symptom is the popout edit dialog doesn't open to edit webform components.

After

Technical Details
The first commit fixes something that seems like it should have never worked, where it seems like it's clicking ALL the `li`'s in the `ul` instead of just the edit one. Maybe it just happened to work out.

The second commit changes from waitForField on a different target (which doesn't actually do anything except wait for up to 10 seconds regardless, because the return value should be checked to see if it actually appeared) to doing the same thing drupal core does in its tests of "off-canvas" popouts.

Comments
One difficulty with troubleshooting this is that the test runs fine before with the same environment locally, so it's a bit of guesswork.

Running with the whole suite now - let's see if we need to adjust slightly for some tests.

### DIFF
--- a/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
+++ b/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php
@@ -351,9 +351,11 @@ abstract class WebformCivicrmTestBase extends CiviCrmTestBase {
    *  TRUE if only one option is enabled on the element.
    * @param string $asList
    *  TRUE if element need to be rendered as select element.
+   * @param string $secondarySelector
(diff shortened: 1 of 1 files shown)


KarinG:
Jippie!

Drupal 10.1.x stopped receiving security coverage in June 2024 - so I'm merging this and will follow up to remove D10.1 from the test matrix.

PR #1023, opened 2024-11-14
Delete skipped test

demeritcowboy:
Overview
Pretty sure this was just something Matt G whipped up while initially getting the system set up. All it seems intended to do is check that an email field on the form _doesn't_ have any live option choices, which doesn't really make sense. It's being skipped because it doesn't run properly here anyway.


KarinG:
Beat me to it :-)

PR #1024, opened 2024-11-17
[do not merge] See what's up with a single test fail

demeritcowboy:
Overview

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,15 +23,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - drupal: '10.2.*'
-            civicrm: '5.78.*'
-            php: '8.2'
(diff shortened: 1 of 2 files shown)


demeritcowboy:
It says $elements = array();

And `$webform->get('elements')` is `{  }`

The parent webform is the same version as other runs (6.2.7). Maybe `get()` is a generic drupal core function that's changed.

Also not directly related but relevant for future: `966x: The "yaml_parser_class" setting is deprecated in drupal:10.3.0. This setting should be removed from the settings file, since its usage has been removed. See https://www.drupal.org/node/3415489.` which means `Drupal\Core\Serialization\Yaml` is deprecated.

KarinG:
PS - after we solve this - let's leave this PR open as that will save us time to amend it to check any future test fail.

PR #1031, opened 2025-01-18
Fix activity notifications to respect civi setting

jitendrapurohit:
Overview
Fix activity notifications to respect civi setting

Before
Despite settings to prevent them, Activity notifications are being sent to Assignee

After
Activity notifications are not sent to Assignee if configured in civi settings.

Technical Details
Check if activity type is configured in do not notify setting in civicrm.


KarinG:
Ext dir was MIA but is back now -> merging. Thank you Jitendra!

PR #1033, opened 2025-01-21
Fix Duplicate Contact Detection

shahrukh-compuco:
Overview
This pr fixes the detection of duplicate contact. Prior to this pr if a dedupe rule used email field for detecting duplicates then it did not work if the duplicates were between the contacts other than the first contact as the email for first contact is saved at the time of validation but for other contacts the emails are saved at the end after all new contacts have been created and for this scenario two different contacts were getting created through webform submission even if they have same dedupe rule fields. This pr fixes the issue and now the duplicate contacts are detected successfully.


KarinG:
Drupal 7 has reached EOL. This branch will no longer receive bug fixes - security/vulnerability issues will still be looked at.

PR #1035, opened 2025-02-11
[Test only] See if this helps with lab extdir overload

demeritcowboy:
Overview
Extension lookup times out a lot since the last few months

Before
Fail

After
Maybe not fail?

Technical Details
Just bypass the extdir lookups and fix the version of the extensions downloaded. Downside is need to update the yml if want a newer version of the extension, but this is kind of like how real sites work anyway.

Comments


KarinG:
"pretend we're a real site and have old versions of the extensions installed" 🥳

demeritcowboy:
Now I think we overloaded github too. No tests running.

KarinG:
image

KarinG:
Too many spaces? :-)

KarinG:
Merging this - it does solve the "not able to download extensions via cv" issue. It surfaced a new issue in RC/dmaster ->   `public function testDynamicCustomFields() `fails, but that's for later.

PR #1044, opened 2025-03-17
Fix tiny typo in Utils.php

niels-heinemann:
Overview
When using specific date title and display in event settings, this false reference to ```wf_crm_get_civi_setting()``` leads to  

Call to undefined function Drupal\webform_civicrm\wf_crm_get_civi_setting()


KarinG:
Hi Niels, I ran the tests - all green [as expected]. Thank you for catching this!

PR #1045, opened 2025-03-18
Fixing civicrm date format

vikascividesk:
Overview:
Webform CiviCRM was not submitting the date values in the correct format to CiviCRM records. This patch ensures proper handling of date values in Webform CiviCRM, preventing incorrect date storage and formatting issues.

**Before** 
The date values were being saved in an incorrect format, such as "June 1st, 400", instead of the expected "June 1986" format. This was caused by improper parsing of the date input before submission to CiviCRM. **Please see date-format-before-apply-patch.png**
date-format-before-apply-patch

**After** 
After applying the patch, Webform CiviCRM now submits the date in the correct format (Y-m-d), ensuring that CiviCRM correctly stores the selected year and month. **Please see date-format-after-apply-patch.png.png**
date-format-after-apply-patch

Technical Details:

Concatenates $val with $time to create a timestamp.
Uses preg_replace('/T.*/', '', $timestamp) to remove the time component and retain only the date part.
Ensures the extracted date has exactly 8 characters (YYYYMMDD).
Converts the valid date into a DateTime object and reformats it to Y-m-d.
If the extracted value is invalid, assigns an empty string ('') to $val.

### DIFF
--- a/src/WebformCivicrmPostProcess.php
+++ b/src/WebformCivicrmPostProcess.php
@@ -2582,21 +2582,20 @@ class WebformCivicrmPostProcess extends WebformCivicrmBase implements WebformCiv
           $timestamp = $val . $time;
           // Check if the $timestamp is not empty
           if (!empty($timestamp)) {
-            // Remove the time portion if present (everything after 'T')
+            // Remove the time portion (everything after 'T').
             $timestamp = preg_replace('/T.*/', '', $timestamp);
-
-            // Check if the length of the remaining timestamp is exactly 8 characters (YYYYMMDD format)
+            // Validate if $timestamp length is exactly 8 characters.
             if (strlen($timestamp) === 8) {
-                // Create a DateTime object from the timestamp using 'Ymd' format (YYYYMMDD)
-                $formatted_date = \DateTime::createFromFormat('Ymd', $timestamp);
+              // Convert $timestamp to DateTime object using 'Ymd' format.
+              $formatted_date = \DateTime::createFromFormat('Ymd', $timestamp);
             } else {
-                // If the length is not 8, set $formatted_date to false
-                $formatted_date = false;
+              // Set $formatted_date to false if length is not 8.
+              $formatted_date = false;
             }
-
-            // If the $formatted_date is valid, format it as 'Y-m-d' (YYYY-MM-DD), otherwise set $val to an empty string
+            // Format valid date as 'Y-m-d'.
+            // otherwise set $val to an empty string.
             $val = $formatted_date ? $formatted_date->format('Y-m-d') : '';            
-        }
+          }
         // The admin can change a number field to use checkbox/radio/select/grid widget and we'll sum the result
         elseif ($field['type'] === 'number' || $field['type'] === 'civicrm_number') {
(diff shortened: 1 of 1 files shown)


KarinG:
Hi, is there a drupal.org issue that describes the specific context? Many of our clients have webforms that have date of birth field on their webforms and they all work as expected. I'll add to a Functional JS test.

vikascividesk:
Hi @KarinG 
This is not a Drupal.org issue. I encountered the same issue on two websites, so I investigated and fixed it using this code. The problem is not limited to the **"date of birth"** field; it affects all **"Date List"** fields.

KarinG:
I'll need you to create an issue on drupal.org -> webform-civicrm project -> https://www.drupal.org/project/webform_civicrm
There is a template that will ask/guide you for details that we will need to reproduce the issue. We have dates working for all sorts of fields (date of birth, activities date-time, membership dates, etc) across dozens of clients so there is something in your setup/config that is different. I started on locking in dob as it works right now into FunctionalJS tests, but got distracted!

KarinG:
If you add a CiviCRM date field to a webform - it's added as a Date element. Are you retyping it to Date List?

nganivet:
@KarinG Thank you for adding the test and confirming that it works, @vikascividesk can you please investigate further and get back to @KarinG?

vikascividesk:
Hi @nganivet 
Sure! I'll look into it and update @KarinG shortly
Thank you

vikascividesk:
Hi @KarinG @nganivet ,

I’ve checked the issue again, but I’m still encountering the same problem with the Date of Birth field. I’ve documented all the details in the attached file for your reference.

Could you please review it and share your feedback when you get a chance?

[Fixing civicrm date format.docx](https://github.com/user-attachments/files/19605746/Fixing.civicrm.date.format.docx)


Thank you!

KarinG:
Key are steps to reproduce - and it looks like your steps are:

1. Add a CiviCRM Date field [for example Birth Date]: 
civicrm_1_contact_1_contact_birth_date
2. Observe (on Build tab) it is added of type = Date
3. Submit Form -> Birth Date is stored properly

Next:
4. On the webform Build tab -> edit element with form key civicrm_1_contact_1_contact_birth_date -> change the Type of the field from Date -> Date list
5. Submit Form -> Birth Date is not stored properly

So this PR is not about fixing CiviCRM Date fields, but about adding support for being able to retype Date fields to fields of type Date list. 

I'll close this PR. If supporting retyping Date fields -> into Date list fields is important to you - the steps are:
a) add a functional JS test to demonstrate reytping to Date list field type is not stored properly
b) once the test fails -> add a commit with a proposed fix

nganivet:
@KarinG we do not have the toolset in place to do functional JS testing. Would you accept a PR with a demonstrated steps to reproduce if it also passes the full existing test suite?

KarinG:
We can write the test if you can sponsor a couple of hours on that. That really needs to be the starting point to bring new functionality into the module.

nganivet:
@KarinG I understand, we will get back to our client about this. Thanks for your time reviewing this issue.

PR #1048, opened 2025-04-03
Always put the element for "Other" at last.

puresyntax71:
Overview
When switching elements to "Radios Other", the "Other" element is rendered on top of the radios.

Before
image

After
image


KarinG:
I've been using some js on webforms to move that field into place. This is nice and easy. Thank you!

PR #1050, opened 2025-04-09
don't count canceled etc. registrations toward participant cap

MegaphoneJon:
Overview
Details are here:
https://www.drupal.org/project/webform_civicrm/issues/3090133

Before
Canceled/rejected/other statuses whose `is_counted` is false are still counted.

After
Only `is_counted` statuses count toward the registration cap.


KarinG:
Tested - thank you Jon!

PR #1051, opened 2025-04-09
prevent type error when user can't access existing contact element

MegaphoneJon:
Overview
See https://www.drupal.org/project/webform_civicrm/issues/3518323.

Before
Visibility controls that affect the presence of this field sometimes have a type error.

After
No type error.

Comments
Sorry for not having a handle on creating a test case - but given the simplicity of the fix hopefully it's acceptable nonetheless.


MegaphoneJon:
This test fail seems spurious?

KarinG:
I think that's an intermittent fail - will re-run (something crashed after 5min)

KarinG:
Will circle back to make sure the merge commit passes.

KarinG:
All green on merge:

image

PR #1052, opened 2025-04-15
Drupal11 prep

demeritcowboy:
Overview
Checking if there's anything in here that doesn't also work in drupal 10 and so it could be merged ahead of time.

[Almost test-only]


KarinG:
🎉 Merging!

jackrabbithanna:
Whoop!! Thanks yall :)

PR #1053, opened 2025-04-21
[TEST ONLY] check if we still need labs workaround

demeritcowboy:
Overview
_A brief description of the pull request. Try to keep it non-technical._

Before
_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

After
_What has been changed. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

Technical Details

Comments


demeritcowboy:
I think the issue here is that Mathieu had put in something so that if an extension has no release in however long, it no longer appears in the ext feed. So the uscounties download is getting tripped up by that. But, since the other ones are working now, and the uscounties one obviously doesn't need an update regularly, could put the other ones back and just directly download uscounties.

KarinG:
Great - thanks Dave!

PR #1060, opened 2025-06-20
Add compatibility for searching contacts on iPhone/Safari

rubofvil:
Overview
Add compatibility for searching contacts on iPhone/Safari

Before
Search not functional on iPhone/Safari

After
Enable contact search functionality as in other browsers

### DIFF
--- a/js/jquery.tokeninput.js
+++ b/js/jquery.tokeninput.js
@@ -379,6 +379,14 @@
                     }
                     break;
               }
+          })
+          // Add input event for iPhone/iOS compatibility
+          .on("input", function(event) {
+              resize_input();
+              // Only trigger do_search if not a blur event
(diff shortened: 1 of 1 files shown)


KarinG:
What are the steps to reproduce the issue?

rubofvil:
> What are the steps to reproduce the issue?

Configure the 'Existing Contact' field with the autocomplete widget

image

Add a group of contacts to filter

image

Disable permission enforcement.

image

---

Use an iPhone to search for the contacts
image

jitendrapurohit (approved):
was able to replicate this on ios device and confirm the patch resolves the problem.

rubofvil:
> was able to replicate this on ios device and confirm the patch resolves the problem.

Thanks for the review, @jitendrapurohit!

PR #1064, opened 2025-07-09
Use civi's filenaming style for uploaded files

demeritcowboy:
Overview
It was desired to make it use the same filenaming style as core civi.

Before
It either uses drupal's style, the style used by the "sanitize" checkbox that webform file field config has, or the filename as is, depending on what/where.

After
The file stored on the civi side will always use the civi style, e.g. my_file_38d7c8d66aa6a88f3426.jpg

Technical Details

Comments
I'm not sure if it's possible to write a mink test when it requires using a filepicker dialog.


KarinG:
I don't think a test is needed here.

KarinG:
Is this ready for merge Dave?

PR #1066, opened 2025-07-11
test card features seems to have changed at stripe?

demeritcowboy:
Overview
I'm not sure 4111111111111111 was ever a valid test card for stripe but maybe always worked. Now it seems that it doesn't trigger the postal code to appear (which is normal for some countries's cards but it used to show up for this one).

Before
Fails on trying to fill out non-existent postal code field.

After
Use a card that triggers the widget to make postal code appear.

Technical Details

Comments


KarinG:
Thanks Dave!

PR #1070, opened 2025-09-15
Remove temporary chromedriver override

KarinG:


demeritcowboy:
This got solved eventually.

PR #1073, opened 2025-10-02
WIP Test Only - confirm bug with setting test mode

stesi561:
Overview
It would appear that setting the Payment Processor Mode to test and then reloading the contribution tab causes the Payment Processor to be set to user select.

This is a WIP test only to see if we can replicate this in automated testing.

### DIFF
--- a/tests/src/FunctionalJavascript/SaveSettingsTest.php
+++ b/tests/src/FunctionalJavascript/SaveSettingsTest.php
@@ -27,6 +27,58 @@ final class SaveSettingsTest extends WebformCivicrmTestBase {
   }
 
   /**
+   * Ensure payment process id setting persists when setting test mode.
+   */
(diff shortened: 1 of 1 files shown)


stesi561:
Yes seems to be failing as expected based on my manual testing.

KarinG:
Thanks! If either of you @stesi561 @jitendrapurohit can have a pass at this test to clean it up a bit - then the tests will automatically re-run and should pass. Then we can merge this in too.

stesi561:
Hmm. Wondering if a better solution here is to test to confirm setting a test payment process results in having a test payment processor.

KarinG:
I do really like end to end testing - as if an admin would be using a test payment processor -> all the way through to verifying that an is_test contribution has been created in CiviCRM.

PR #1076, opened 2025-10-06
avoid duplicate file upload to civicrm on multi-step form

sunilpawar:
Overview
When we having a file upload field in Multi step form many duplicate files being uploaded to the server.

Every time when user click next button then duplicate files being uploaded to the server.

saveDrupalFileToCivi function being called on form validation, this is creating the duplicate uploads to the server.

when we having 1 activity attachment fields in "4 steps form" then 4 files being uploaded to the server for a submission.

- For a single-step webform, no issues were found.
- For a multi-step webform:
- 1. Each file gets duplicated.
- 2. An entry is created in the civicrm_file table at every step.
- 3. An entry in the civicrm_entity_file table is only created after the final submission of the form.
- 4. Finally, three entries present in the `civicrm_file` table do not have any corresponding references in the `civicrm_entity_file` table.
- 5. These abandoned files lead to a large accumulation and potential storage issues.
- Checked the backtrace log file.
- Observed that saveDrupalFileToCivi is being called multiple times.
- This occurs during form validation on each step of a multi-step webform.
- As a result, duplicate files are being created.

Before
Multiple duplicate files created based on number of steps in webform.

After
Avoid creating duplicate files.

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -799,21 +799,105 @@ abstract class WebformCivicrmBase {
   public static function saveDrupalFileToCivi($id, $filename = NULL) {
     $file = File::load($id);
     if ($file) {
+      // Get CiviCRM configuration singleton instance.
       $config = \CRM_Core_Config::singleton();
-      $copyTo = $config->customFileUploadDir;
-      if(!isset($filename)) {
+
+      // Get the custom file upload directory and ensure it ends with a slash.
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
PR for https://www.drupal.org/project/webform_civicrm/issues/3550084
Seems to have existed for a long time https://www.drupal.org/project/webform_civicrm/issues/3196163

Came to our attention on a site which heavily uses activity attachment uploads in webforms.

Normal multi-step form submissions can duplicate a file 6-8 times sometimes, resulting in large use of storage.

Please let us know your thoughts! Thank you!

PR #1077, opened 2025-10-07
[TEST ONLY] Upstream patch no longer needed

demeritcowboy:
Upstream has merged the real PR
https://www.drupal.org/project/webform/issues/3547627


demeritcowboy:
There must be some delay between when drupal PRs are merged and what composer decides you should get - the log indicates it's downloading commit hash 9cab04f but the commit we need is a few hours later. Will try again tomorrow.

KarinG:
You're too fast!

PR #1078, opened 2025-10-11
Enable is_test if payment is made using test mode processor

jitendrapurohit:
Overview
Enable is_test if payment is made using test mode processor

Before
If a payment is made using a test mode processor, it currently results in a existing membership renewal

imo it should make a separate new 'test membership'

i just tested via a civi form on smaster and my membership was not extended when the contribution page was in test mode.

and it did create a Test Membership for the contact.

The issue is only with webforms where it 'knows' the contribution is a test because the payment processor is in test mode, but doesn't know that this should make the Membership a test too

After
If Test Mode is selected in civicrm settings of the webform, the contribution and membership is created with is_test = 1.

Technical Details
Add is_test flag is test mode is enabled in civi settings.

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -595,6 +595,17 @@ abstract class WebformCivicrmBase {
   }
 
   /**
+   * Enable is_test if payment is made using a test processor.
+   *
+   * @param array $params
+   */
+  protected function setTestMode(&$params) {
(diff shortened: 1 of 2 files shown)


KarinG:
No objections! :-)

Toby-fz:
Note, have tested this and works as expected

PR #1079, opened 2025-10-18
Fix admin group removal on submitting public group option

jitendrapurohit:
Overview
Fix admin group removal on submitting public group option

Before
- Enable the Groups field for an Existing Contact in the webform config, choosing "- User Select - (public groups)"
- Have a contact who is in Civi groups that are "User and User Admin only" and so won't show in a listing of public groups
- When the webform is completed after a URL is emailed to them with contact hash etc, the contact is removed from the "User and User Admin Only" groups that aren't listed on the webform.

After
Admin group remain unchanged on submitting public group from the webform.

Technical Details
The values in $this->data is normally changed during the webform submission, but $this->settings['data'] holds the original data values configured in the webform. The changes uses it instead to ensure the group options are fetched correctly.

### DIFF
--- a/src/WebformCivicrmBase.php
+++ b/src/WebformCivicrmBase.php
@@ -567,7 +567,7 @@ abstract class WebformCivicrmBase {
         'extra' => wf_crm_aval($field, 'extra', []) + wf_crm_aval($field, '#extra', []),
         'form_key' => $field['#form_key'],
       ];
-      $exposed = $this->utils->wf_crm_field_options($params, 'civicrm_live_options', $this->data);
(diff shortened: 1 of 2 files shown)


chrisgaraffa:
This fix worked for us on two sites

Before: If the contact had "User and User Admin only" groups, exposing just some public groups via "- User Select - (public groups)" would unsubscribe them from the groups that weren't displayed.

Now: They aren't removed from those groups.

KarinG:
Thank you @jitendrapurohit and @chrisgaraffa 

Tests fails are real - could you please review:

image

PR #1080, opened 2025-10-30
try to debug consistent test fail

demeritcowboy:
Overview

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,18 +23,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - drupal: '10.4.*'
-            civicrm: '6.4.*'
-            php: '8.2'
-          - drupal: '10.5.*'
-            civicrm: '6.5.x-dev'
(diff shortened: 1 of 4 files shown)


demeritcowboy:
According to the screenshots and debug output, everything looks ok until it tries to retrieve the submission id, even though screenshot 9 shows the successful "New submission added to CiviCRM" message. Debug output shows it has the right webform id, but it returns no submission ids:

webform id: civicrm_webform_test
submission ids: Array
(
)

Since this ran ok 3 days ago, my thinking is some change in drupal affecting the below query, but having trouble getting a specific hash-set of drupal code to install in the test to confirm that.
```php
\Drupal::entityQuery('webform_submission')
      ->accessCheck(TRUE)
      ->condition('webform_id', $webform->id())
      ->sort('created', 'DESC')
      ->range(0, 1)
      ->execute();
```
It's not some change in the parent webform since the only change there is code comments.

Will see if can reproduce manually on a 11.x-dev install.

PR #1081, opened 2025-11-03
ignore

demeritcowboy:
Overview

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,7 +79,7 @@ jobs:
           composer config repositories.0 composer https://packages.drupal.org/8
           composer config repositories.1 path $GITHUB_WORKSPACE
           composer install --no-interaction
-          COMPOSER_MEMORY_LIMIT=-1 composer require drupal/core-dev-pinned:11.x-dev#3eaf671 drupal/core-dev:11.x-dev#74b93b4 drupal/core:11.x-dev#aa12cc60d44f1d78 drupal/core-recommended:11.x-dev#2ce7bc1 doctrine/annotations:2.0.2 psr/cache -w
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Since it's passing here, with these specific versions `drupal/core-dev-pinned:11.x-dev#3eaf671 drupal/core-dev:11.x-dev#74b93b4 drupal/core:11.x-dev#75b81cd drupal/core-recommended:11.x-dev#2ce7bc1 doctrine/annotations:2.0.2`, it suggests the problem is something in drupal core or related libraries. Somewhere between core 75b81cd and da502d9.

* It doesn't appear to be anything in the tooling, and mink/behat and related are all the same versions.
* The upgrade to phpunit doesn't seem relevant. 
* It seems unlikely the removal of doctrine/annotations would matter, but maybe.
* What's clear from #1080 is that there's now a mismatch between the logged in user in the browser, and the logged in user in code that mink runs. I don't understand how that low-level mechanism works, but I don't see anything obvious related to that area except https://www.drupal.org/node/3548485.

Possible next step, try a poor-man's git bisect here.

demeritcowboy:
The bisect says https://github.com/drupal/core/commit/aa12cc60d44f1. It's possible it's wrong because it's difficult to do a proper bisect. Attempting to revert it fails.

demeritcowboy:
The bisect seems right and it does seem theme-related. The issue is either [this type of theme switching](https://github.com/colemanw/webform_civicrm/blob/61eb177d92043a6169bf14a22d9c4d26cfd88a81/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php#L93-L97) is not supported or the above-mentioned commit broke it, depending on how you want to phrase it.

PR #1084, opened 2025-11-03
[Tests only] Workaround for drupalLogin

demeritcowboy:
Overview

Before
Setting the theme procedurally now causes a problem where drupalLogin doesn't carry over the logged-in user to the backend.

After
Work around it for now.

Technical Details
See https://www.drupal.org/project/drupal/issues/3555844

Comments


demeritcowboy:
If #1085 passes I think this should be closed since 1085 is a simpler workaround.

PR #1085, opened 2025-11-03
[Tests only] Now need to reset the container after switching the theme

demeritcowboy:
Overview

Before
Setting the theme procedurally now causes a problem where drupalLogin doesn't carry over the logged-in user to the backend.

After
Work around it for now.

Technical Details
See https://www.drupal.org/project/drupal/issues/3555844

Comments


demeritcowboy:
The test fail is one of the usual and 11.x-dev is fine, so I think this is the best option.

KarinG:
Thank you Dave!

KarinG:
Hopefully the 10.5 / CiviCRM-dev combo passes during the (on merge) test runs.

PR #1086, opened 2025-11-15
Test Only: Target key instead of label for is_active field

jitendrapurohit:
Overview
Target key instead of label for is_active field

Before
Tests failed because the label changed from “Active” to “Enabled"

After
The test now targets the stable is_active field id, working for both checkbox and toggle variants.

fyi @KarinG


KarinG:
Thank you @jitendrapurohit

PR #1092, opened 2025-12-08
See if 3555844 is fixed upstream

demeritcowboy:
Overview


demeritcowboy:
This PR is just removing workarounds in tests for bugs in 3rd party dependencies that are now fixed, so merging.

PR #1094, opened 2026-01-05
[Test only] drupal core-dev-pinned is currently broken upstream

demeritcowboy:
Overview
Upstream needs updating, but don't know when that will be.

Before
Upstream pins to library versions that composer won't install anymore.

After
Make it more like a real install where you are subject to the randomness of library releases, but less likely to get blocked on a specific dependency.

Technical Details
Yes, it is technical.

Comments
I'm personally 50/50 on the strategy.

It's also unclear to me if this hasn't been caught upstream because (a) nobody is using pinned anywhere, (b) the holidays, (c) it's not the biggest priority upstream. But if it's (a) or (c) it suggests leaning towards not depending on it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,13 +23,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - drupal: '10.4.*'
+          - drupal: '10.5.*'
             civicrm: '6.4.*'
             php: '8.2'
           - drupal: '10.5.*'
             civicrm: '6.9.x-dev'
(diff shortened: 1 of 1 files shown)


demeritcowboy:
The 11.2 fail might just be one of those intermittent ones.
For 10.4, technically drupal 10.4 is no longer supported. Since 10.6 is out I'll update this PR to switch up the matrix a bit to include 10.6.

An alternative to all this is tell composer to allow insecure libraries, but that feels wrong.

PR #897, opened 2023-07-23
Multi-page webform invalid data with image

jitendrapurohit:
Overview
----------------------------------------
Data not submitted when upload image element is present in the multi-page webform setup.

Before
----------------------------------------
When having a multi-step form (with or without preview) and on one of the steps/pages we got a contact image field (file upload field), the submitted data do not keep the other changes but just the contact's image.

See drupal ticket for details - https://www.drupal.org/project/webform_civicrm/issues/3372150

After
----------------------------------------
Data submitted without any issues.

Technical Details
----------------------------------------
When you upload a file using a managed_file element, it involves AJAX requests behind the scenes, which trigger the form alteration hooks.

This patch prevents alterForm method from being called when you upload a file.
 
Comments
----------------------------------------
Drupal Ticket - https://www.drupal.org/project/webform_civicrm/issues/3372150

PR #1095, opened 2026-01-10
Issue#3533734: Using Exiting Contact + Select list + file upload fails

jitendrapurohit:
Overview
Using Exiting Contact + Select list + file upload fails to submit the form due to validation error.

Before
Validation error on submitting the form with existing select contact element + file upload field.

After
No error. 

Comments
fyi @KarinG

### DIFF
--- a/src/WebformCivicrmPreProcess.php
+++ b/src/WebformCivicrmPreProcess.php
@@ -109,10 +109,12 @@ class WebformCivicrmPreProcess extends WebformCivicrmBase implements WebformCivi
     // Early return if the form (or page) was already submitted
     $triggering_element = $this->form_state->getTriggeringElement();
 
-    // When user uploads a file using a managed_file element, avoid making any change to $this->form.
+    // When user uploads a file using a managed_file element, populate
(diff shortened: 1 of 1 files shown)


KarinG:
Thanks @jitendrapurohit :-)

Tests: looks like a (temp) upstream Drupal issue.

jmcclelland:
@jitendrapurohit @KarinG It works for me! Thanks - such a simple one line fix that eluded me.

KarinG:
Thank you both!

stesi561:
Just circling back here as have found this same problem via another angle. I'm a bit worried that this will result in the original behaviour failing -> https://github.com/colemanw/webform_civicrm/pull/897

When I was looking at this I don't think there is much testing coverage on attachments? But I could be wrong so I'm wondering if this might have reintroduced the above?

stesi561:
I'd worked up a slightly different approach here -> https://github.com/stesi561/webform_civicrm/tree/6.x-fixRequiredFiles 
this just addresses the required flag on the fields.

stesi561:
Note: I've not been able to replicate yet. So am deploying this fix.

PR #1098, opened 2026-01-20
Block unknown users: do not 403 embedding pages

yakim-ua:
## Overview
Fixes anonymous access regression when Block unknown users is enabled.
Instead of returning a 403 for the entire page that embeds the form, only the webform output is hidden.

Issue: https://www.drupal.org/project/webform_civicrm/issues/3485026
Reference MR: https://git.drupalcode.org/project/webform_civicrm/-/merge_requests/21

## Before
With Block unknown users enabled, visiting a page that embeds a Webform CiviCRM form as an anonymous user results in a 403 for the whole page, even if the page itself should be publicly viewable.

## After
The page remains accessible. The embedded webform output is hidden for anonymous users.

## Technical Details
- Scope access control to the webform render output instead of denying the route.
- Replace throwing AccessDeniedHttpException with setting the form render array access flag:
  - $this->form['#access'] = FALSE;

## Comments
Integration Tests should run on GitHub Actions for this PR.

### DIFF
--- a/src/WebformCivicrmPreProcess.php
+++ b/src/WebformCivicrmPreProcess.php
@@ -166,8 +166,7 @@ class WebformCivicrmPreProcess extends WebformCivicrmBase implements WebformCivi
         $this->form['#prefix'] = wf_crm_aval($this->form, '#prefix', '') . '<div class="webform-civicrm-prefix contact-unknown">' . nl2br($this->settings['prefix_unknown']) . '</div>';
       }
       if ($this->settings['block_unknown_users']) {
-        $this->form['submitted']['#access'] = $this->form['actions']['#access'] = FALSE;
-        throw new AccessDeniedHttpException();
+        $this->form['#access'] = FALSE;
       }
     }
     if (!empty($this->data['participant_reg_type'])) {


yakim-ua:
This PR is ready for CI.
GitHub Actions shows 1 workflow awaiting approval for this fork PR.
Could a maintainer please approve and run the Integration Tests workflow?
Thanks.

cc @colemanw @KarinG

KarinG:
I don't think this is a regression - it has always worked like this. I'm happy to run the tests - but am worried if such a change in behaviour would result in unintended and unexpected consequences for some admin/users.

petednz:
just to add my voice - i think this is a good fix and can't think of any scenarios where we would get bitten by this change.

KarinG:
Ok that's helpful. Thanks Pete.

Tests are passing (two fails are b/c civicrm.org was down for a bit so the tests could not download and install the extensions it needs).

KarinG:
Thank you @yakim-ua - I'll circle back to drupal.org and award you some credits.

yakim-ua:
Thanks, @KarinG  - appreciate you running the tests and confirming the external cause of the two failures.
Additionally, thank you, @petednz , for the endorsement and for helping de-risk the change with an extra reviewer perspective.

I’m happy to keep contributing and help resolve issues going forward, if that would be useful.

petednz:
Your welcome - just read the backstory about DevBranch. Fascinating. well done all of you

PR #1099, opened 2026-01-22
Support Toggle html type

pradpnayak:
Overview
A new custom field type, Toggle, was introduced in CiviCRM 6.10 for Yes/No data. Custom fields of this type do not currently appear in the CiviCRM Webform field listing.

This PR adds support for the new Toggle field type.


KarinG:
Ah - thank you @pradpnayak - I will merge on pass.

PR #1101, opened 2026-01-29
[Test only] Add test to check pay later is sent for Membership enabled webforms.

stesi561:
Overview

Test to hopefully demonstrate [2593357](https://www.drupal.org/project/webform_civicrm/issues/2593357)

Before

Pay later is only tested for a Contribution.

After

Adds a test to testSubmitMembershipQueryParams to check what happens on a membership enabled page.

Comments

Wasn't sure where best to put this as involves copying logic from two different tests. Ideally we would avoid copying test code around and abstract some of the logic into a utility function.

### DIFF
--- a/tests/src/FunctionalJavascript/MembershipSubmissionTest.php
+++ b/tests/src/FunctionalJavascript/MembershipSubmissionTest.php
@@ -283,6 +283,62 @@ final class MembershipSubmissionTest extends WebformCivicrmTestBase {
 
     $this->assertEquals('Basic Plus', $membership['membership_name']);
     $this->assertEquals('1', $membership['status_id']);
+
+    // Now lets check pay later notifications
+    // Possibly should be in it's own test but would require
+    // Duplicating a lot of above logic.
+    $this->drupalLogin($this->rootUser);
+    $this->redirectEmailsToDB();
+
+    $this->drupalGet(Url::fromRoute('entity.webform.civicrm', [
+      'webform' => $this->webform->id(),
+    ]));
+    $params = [
+      'payment_processor_id' => 'Pay Later',
+      'receipt' => [
+        'receipt_from_name' => 'Admin',
+        'receipt_from_email' => 'admin@example.com',
+        'pay_later_receipt' => 'Payment by Direct Credit to: ABC. Please quote invoice number and name.',
+        'receipt_text' => 'Thank you for your contribution.',
+      ],
+    ];
+    $this->configureContributionTab($params);
+    $this->saveCiviCRMSettings();
+
+    $this->drupalLogout();
+
+    $this->drupalGet($this->webform->toUrl('canonical', ['query' => ['membership' => 2]]));
+    $this->htmlOutput();
+    $this->assertPageNoErrorMessages();
+
+    $this->assertSession()->waitForField('First Name');
+    $this->getSession()->getPage()->fillField('First Name', 'Sarah');
+    $this->getSession()->getPage()->fillField('Last Name', 'Sinclar');
+    $this->getSession()->getPage()->fillField('Email', 'sarah@example.com');
+    $this->assertSession()->pageTextContains('Basic Plus');
+    $this->pressButtonOverride('Next >');
+    $this->htmlOutput();
+    $this->assertPageNoErrorMessages();
+
+    $this->pressButtonOverride('Submit');
+    $this->htmlOutput();
+    $this->assertPageNoErrorMessages();
+
+    $this->assertSession()->pageTextContains('New submission added to CiviCRM Webform Test.');
+
+    $api_result = $this->utils->wf_civicrm_api('membership', 'get', [
+      'sequential' => 1,
+      'status_id' => "Pending",
+    ]);
+    $this->assertEquals(1, $api_result['count']);
+    $membership = reset($api_result['values']);
+    $this->assertEquals('Basic Plus', $membership['membership_name']);
+
+    $sent_email = $this->getMostRecentEmail();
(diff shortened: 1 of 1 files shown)


KarinG:
I don't mind copied test code - it keeps things easy to read.

stesi561:
Note: after some more digging it looks like this might be a CiviCRM core issue.

KarinG:
Oh - then we'll merge this test as soon that's fixed in Core [to catch a future regression].

stesi561:
Just noting that I've put up #1103. 

Further digging suggests that Core has changed how they are including pay later text for memberships and receipts. I think the solution is we update webform_civicrm to account for new handling. 

I don't think the test failures above are from what I was wanting to test. Note sure if you want the tests - once failing and working properly wrapped into the same PR or not? I assume probably best to demonstrate a failure here, the PR passing - then cherrypick this commit onto the other PR and demostrate it then passes to confirm that we are good.

tests/src/FunctionalJavascript/MembershipSubmissionTest.php
```diff
@@ -287,6 +287,58 @@ public function testSubmitMembershipQueryParams() {
...
+    $this->drupalGet($this->webform->toUrl('canonical', ['query' => ['membership' => 2]]));
+    $this->htmlOutput();
+    $this->assertPageNoErrorMessages();
+
+    $this->assertSession()->waitForField('First Name');
+    $this->getSession()->getPage()->fillField('First Name', 'Sarah');
+    $this->getSession()->getPage()->fillField('Last Name', 'Sinclar');
+    $this->getSession()->getPage()->fillField('Email', 'sarah@example.com');
+    $this->assertSession()->pageTextContains('Basic Plus');
+    $this->pressButtonOverride('Submit');
```
demeritcowboy, line 323: Looking at the screenshots, by configuring the contribution tab above it's now a multistep form with a payment step, so the button is now called Next and then there's another page. But I'm not sure if that's just how pay later with webform works or whether this isn't the way to set that up for memberships.

stesi561: Where do I see the screenshots?

Aha. I was curious as to why the other test didn't end up with a multi step form.

I think I just changevthisvto click next then submit.

stesi561: Where do I see the screenshots?

Aha. I was curious as to why the other test didn't end up with a multi step form.

I think I just changevthisvto click next then submit.

stesi561: Thanks!

demeritcowboy: Yes I think so. I was looking into it but then got sidetracked by the original test being confusing (why does it create a paid membership but then offer it without payment?) So I think setting up the contribution tab is correct and yes then next and then just submit.

The screenshots are hidden on the summary page for the tests, at the bottom: https://github.com/colemanw/webform_civicrm/actions/runs/21460733706?pr=1101

stesi561:
Thanks - sorry I couldn't spot then got distracted setting up tests to run on my local - which lead to other rabbit holes and I didn't have time to push everything back off the stack!

demeritcowboy:
No worries

demeritcowboy:
Success! By which I mean failure.
That feels like there's a life lesson there: If you try to fail, you will always succeed no matter what.

```
Failed asserting that 'MIME-Version: 1.0\n
From: Admin <admin@example.com>\n
... [lots of stuff] ...
' contains "Payment by Direct Credit to: ABC. Please quote invoice number and name.".
```

So now let's cherry-pick the fix and see.

stesi561:
Hmm. After the test run I'll fix up that merge as we don't want any changes to that file.

demeritcowboy:
@stesi561 Does this need any more updates other than remove that last commit and then squash? It looks like 1103 was merged and then that last commit was just to trigger tests, and then this passed, so seems mergeable?

stesi561:
Yeap looks good.

demeritcowboy:
Hmm, consistently failing now:

```
Drupal\Tests\webform_civicrm\FunctionalJavascript\MembershipSubmissionTest::testSubmitMembershipQueryParams
Failed asserting that 'MIME-Version: 1.0\n
From: Admin <admin@example.com>\n
To: Sarah Sinclar <sarah@example.com>\n
Subject: Invoice -  - Sarah Sinclar\n
Return-Path: admin@example.com\n
Date: Fri, 03 Jul 2026 08:37:17 +1000\n
Reply-To: Admin <admin@example.com>\n
Content-Type: multipart/alternative;\n
 boundary="=_b00ad47dcb1acd49a45cc56e0b8c5cc7"\n
\n
--=_b00ad47dcb1acd49a45cc56e0b8c5cc7\n
Content-Transfer-Encoding: 8bit\n
Content-Type: text/plain; charset=utf-8\n
\n
Dear Sarah,\n
\n
Thank you for your contribution.\n
\n
Membership Information\n
Membership Type	Basic Plus\n
Membership Start Date	July 3rd, 2026\n
Membership Expiration Date	July 2nd, 2027\n
Membership Fee\n
Contribution Amount	$1.00\n
Amount	$1.00\n
Date	July 3rd, 2026 8:37 AM\n
Registered Email\n
sarah@example.com\n
--=_b00ad47dcb1acd49a45cc56e0b8c5cc7\n

... [omitted] ...

' [ASCII](length: 4171) contains "Payment by Direct Credit to: ABC. Please quote invoice number and name." [ASCII](length: 71).
```

PR #1102, opened 2026-01-29
Removes unneeded unset

stesi561:
Overview
We create an array and populate it - but do not populate the key 'payment_processor'. Then we call unset.

Before
This seems unnecessary?

After
Removes  an unneeded unset.

Technical Details

Comments
I didn't actually check the origin of this I should and will note below.


stesi561:
Seems was added when this code was added back in 3b0e54965a559e97929aeb8277b697939c75564e

PR #1105, opened 2026-02-03
Fixed State province ordering on Mac

pradpnayak:
Overview
Sate is not sorted in alphabetical order

Before
Screenshot 2026-02-03 at 22 12 40

After
Screenshot 2026-02-03 at 22 12 59

### DIFF
--- a/js/webform_civicrm_forms.js
+++ b/js/webform_civicrm_forms.js
@@ -328,7 +328,10 @@ var wfCivi = (function (D, $, drupalSettings, once) {
   }
 
   function fillOptions(element, data) {
-    var sortedData = Object.entries(data).sort(([,a],[,b]) => a > b);
+    var sortedData = Object.entries(data)
+      .sort(function(a, b) {
+        return a[1].localeCompare(b[1]);
+      });
     var $el = $(element),
       value = $el.attr('data-val') ? $el.attr('data-val') : $el.val();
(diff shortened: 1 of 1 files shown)


KarinG:
How come yours are not ordered alphabetically to start with?

image

pradpnayak:
I am on Drupal 11, may be that's why? I can see no issue with some website which are on d10 but definitely on few of them which are on d11

pradpnayak:
Canada looks fine because they are stored in alphabetical order in database, but not for some countries like US, UK, India, China

See: https://demo.circle-interactive.co.uk/form/civicrm-contribution-webform

KarinG:
The form you referenced above has the County field (not State Province). For US - those are separate fields. This is what I see for UK -> on D10.

image

pradpnayak:
Hi @KarinG 

The state province is renamed as county, if you inspect the element it would say state_province_id

Also the screenshot you shared about UK states is still not sorted if see closely and also scroll down

KarinG:
Ah right - I thought perhaps you had counties similar to https://civicrm.org/extensions/us-county-loader.

KarinG:
Hi @pradpnayak - Dave and I looked at this and on not-a-Mac - there is no issue with the current sorting that is already there. But I can see the sorting issue on a Mac. I take it you're working on a Mac :-)

PR #1109, opened 2026-03-09
Updating Civi versions in the Matrix in prep for MARCH security update#1107

KarinG:
Overview

Upping versions

6.10 is ESR
6.12 is latest stable
6.13.x-dev is RC
dev-master is exactly that


KarinG:
Ok this should work now!

PR #1110, opened 2026-03-09
[Test only] work around whatever brick is up to

demeritcowboy:
Overview
Testing ~~https://lab.civicrm.org/extensions/stripe/-/merge_requests/294~~ https://lab.civicrm.org/extensions/stripe/-/merge_requests/295


demeritcowboy:
stripe master has been updated

PR #1113, opened 2026-03-20
autocomplete not working properly when display fields contains contact id

demeritcowboy:
Overview
autocomplete not working properly when display fields contains contact id.

Before
1. Have more than 12 contacts in your database (this just makes it easier to see)
1. Create a form with Existing Contact and make it an autocomplete.
1. For the display fields include contact id, e.g. display name and contact id.
1. Visit the form and try to use the autocomplete. It will match on everything.

After
Better

Technical Details
Something about CONTAINS has changed when the field is non-string. Sometime between 6.4 and 6.12.

Comments
First run includes just the failing test, then second commit includes a fix.

### DIFF
--- a/src/ContactComponent.php
+++ b/src/ContactComponent.php
@@ -160,7 +160,7 @@ class ContactComponent implements ContactComponentInterface {
     if ($str) {
       $searchFields = [];
       foreach ($display_fields as $fld) {
-        $searchFields[] = [$fld, 'CONTAINS', $str];
+        $searchFields[] = [$fld, 'LIKE', '%' . $str . '%'];
       }
       $params['where'][] = ['OR', $searchFields];
     }
--- a/tests/src/FunctionalJavascript/ExistingContactElementTest.php
+++ b/tests/src/FunctionalJavascript/ExistingContactElementTest.php
@@ -177,6 +177,15 @@ final class ExistingContactElementTest extends WebformCivicrmTestBase {
    * searchable with all display field values.
    */
   public function testDisplayFields() {
+    // Fill with more than the widget displays, and that would come first
+    // before our desired contact
+    for ($i = 1; $i < 21; $i++) {
+      $this->createIndividual([
+        'first_name' => 'Aaron',
+        'last_name' => 'Aardvark' . $i,
+        'source' => '',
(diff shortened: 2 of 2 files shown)


demeritcowboy:
Yes it fails. Now adding fix and seeing if that breaks anything.

```
1) Drupal\Tests\webform_civicrm\FunctionalJavascript\ExistingContactElementTest::testDisplayFields
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Aaron'
+'James'

Error: /home/runner/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Equality/IsEqual.php:95
/home/runner/work/webform_civicrm/webform_civicrm/tests/src/FunctionalJavascript/WebformCivicrmTestBase.php:354
/home/runner/work/webform_civicrm/webform_civicrm/tests/src/FunctionalJavascript/ExistingContactElementTest.php:237
```

KarinG:
Thanks Dave! Is this ready to be merged? Shall I issue a new release - I'm thinking yes.

jmcclelland:
Thanks for making this fix. Before seeing it I opened a core issue here: https://lab.civicrm.org/dev/core/-/issues/6397

colemanw:
I'm confused by this issue because the query `WHERE id LIKE "%Bob%"` is nonsensical. But I went ahead and added a core fix anyway: https://github.com/civicrm/civicrm-core/pull/35194

I'm still confused tho. Shoudn't we avoid nonsensical queries?

demeritcowboy:
It's a fair point but I suppose there's:
* Semantic
* Technical
* Other considerations

For semantic, `WHERE id LIKE '%Bob%'` maybe doesn't make sense, but `WHERE id LIKE '%233%'` might (although you'd probably not be looking in the middle of the string it would be more likely you meant `WHERE id LIKE '233%'`, but that's fine.)

For technical, mysql seems ok with it, and numbers can be treated as a subset of strings.

For other considerations, with the core patch then people have a choice - they can upgrade webform if they don't want to upgrade core, or they can upgrade core if they don't want to upgrade webform. Or they can patch either - both are small patches.

PR #1116, opened 2026-05-02
Fixes #3588116 - Contact field submits empty if populated via AJAX

jitendrapurohit:
Overview
Fixes hidden contact fields (populated via relationship AJAX autofill) not being saved on form submission.

Before
When a contact field is configured with "Default value = Relationship" and a hidden/static widget, the AJAX correctly fills the field in the browser, but the submission always stores a blank value.

After
The relationship-autofilled contact value is correctly saved on submission.

Technical Details
The fix adds a fallback in `getExistingContactIds()` to read from `$form_state->getUserInput()` when `submissionValue()` returns 0, still passing through `verifyExistingContact()` for permission checks.

Comments
Drupal Issue: https://www.drupal.org/project/webform_civicrm/issues/3588116


KarinG:
Hi @jitendrapurohit - can you please add a (separate) test to show this is currently failing?

PR #1119, opened 2026-05-15
Issue #3590183: Check field key instead of field name to avoid accidentally deleting other hidden fields

sdoijad:
reference the Drupal.org issue https://www.drupal.org/project/webform_civicrm/issues/3590183

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -2233,7 +2233,7 @@ class AdminForm implements AdminFormInterface {
       }
       else {
         $field = $this->utils->wf_crm_get_field($key);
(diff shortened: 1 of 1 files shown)


sdoijad:
@colemanw : Can you please look at it

demeritcowboy:
When you change between test and live or vice-versa on the civicrm tab and save, if you look at the `value` of the field on the build tab (click to edit - see the value), it switches between 1 and 0 and vice versa. This seems correct to me. It doesn't seem like it needs to be removed as long as it has the correct value.

Have you seen any problems with test vs live not working correctly at run time?

PR #1121, opened 2026-06-03
Add option to hide existing membership status message

francescbassas:
When using webform_civicrm with membership forms, the existing membership status message is always shown to contacts with an active/expired membership.

Some use cases require hiding this message (e.g. renewal forms where the status info is redundant or confusing for the user).

## Proposed changes
- Adds a checkbox "Show existing membership status message" in the Memberships tab of the CiviCRM webform settings.
- Defaults to enabled (TRUE) to preserve existing behavior for all current installations.
- Passes the option as a parameter to `loadMemberships()`, which conditionally calls `$this->setMessage()`.

## Backward compatibility
Fully backward compatible. Default value is TRUE (checkbox checked), so no existing configuration is affected.

### DIFF
--- a/src/WebformCivicrmPreProcess.php
+++ b/src/WebformCivicrmPreProcess.php
@@ -457,7 +457,7 @@ class WebformCivicrmPreProcess extends WebformCivicrmBase implements WebformCivi
    * @param int $c
    * @param int $cid
    */
-  private function loadMemberships($c, $cid) {
+  private function loadMemberships($c, $cid, $showMessage = TRUE) {
     $today = date('Y-m-d');
     foreach ($this->findMemberships($cid) as $num => $membership) {
       // Only show 1 expired membership, and only if there are no active ones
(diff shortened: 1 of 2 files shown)


francescbassas:
@KarinG Are the test errors correct? Could you guide me if so?

src/WebformCivicrmPreProcess.php
```diff
@@ -186,7 +186,8 @@ public function alterForm() {
         }
         // Membership
         if (!empty($this->data['membership'][$c]['number_of_membership'])) {
-          $this->loadMemberships($c, $contact['id']);
+          $showMessage = (bool) wf_crm_aval($this->data, 'membership_options:show_status_message', 1, TRUE);
+          $this->loadMemberships($c, $contact['id'], $showMessage);
```
demeritcowboy, line 190: For the test fails, this is passing a new parameter to loadMemberships() but down on line 460 the function hasn't been updated to accept it as a parameter.

francescbassas: 🤦 thanks @demeritcowboy

francescbassas:
Moved here https://github.com/colemanw/webform_civicrm/pull/1125

PR #1122, opened 2026-06-03
Refresh Existing Contact #allow_create flag when CiviCRM settings are saved

peetCreative:
### Problem

An "Existing Contact" element's `#allow_create` flag never updates once the element is on the form. If you later remove the last name/email field for that contact, `#allow_create` stays stuck at `1`, so the form keeps trying to create contacts (and can error out during processing) even though there are no fields to create one from. The reverse is also true — adding name/email fields to a contact that already has an Existing Contact element won't flip the flag to `1`.

### Root cause

`#allow_create` is derived from the presence of name/email fields via `wf_crm_name_field_exists()` and is recomputed in `AdminForm::insertComponent()`. But `insertComponent()` only writes the value back when **adding a new** element — its final guard skips elements that already exist:

```php
// AdminForm::insertComponent()
if (empty($enabled[$field['form_key']])) {   // false for an existing field
  $enabled[$field['form_key']] = $field;
}
```

`addEnabledElements()` likewise only rewrites array (new) entries, not the string (existing) ones. So for any Existing Contact field already on the form, the recomputed `allow_create` is discarded and the stored value goes stale.

There was even a long-empty stub left in `postProcess()` marking where this was meant to be handled:

```php
// Update existing contact fields
foreach ($existing as $fid => $id) {
  if (substr($fid, -8) === 'existing') {
    $stop = null;
  }
}
```

### Fix

Implement that stub: on save, for each existing `*_contact_existing` field, recompute `allow_create` (same calculation `insertComponent()` uses) and write it back via `setElementProperties()` when it differs. The trailing `$webform->save()` in the settings-form submit handler persists the change.

### Test

`SaveSettingsTest::testExistingContactAllowCreateRefresh()` — enables First/Last Name (asserts stored `#allow_create == 1`), then removes both name fields via the settings form + delete confirmation (asserts it refreshes to `0`).

### Notes

- Behaviour change is limited to the moment CiviCRM settings are saved; it does not touch front-end submission handling.
- Existing forms pick up the corrected value the next time their CiviCRM settings are saved.

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -2040,10 +2040,25 @@ class AdminForm implements AdminFormInterface {
 
     $this->addEnabledElements($enabled);
     $this->setPaging();
-    // Update existing contact fields
+    // Refresh the allow_create flag on existing "Existing Contact" fields.
+    // It is derived from the presence of name/email fields on the form and may
+    // have changed, but insertComponent() only writes the recomputed value when
+    // adding a *new* element (see its guard) - so existing elements keep a stale
+    // value. Bring them back in sync here.
     foreach ($existing as $fid => $id) {
-      if (substr($fid, -8) === 'existing') {
-        $stop = null;
+      if (!str_ends_with($fid, '_contact_existing')) {
+        continue;
+      }
+      $element = $this->webform->getElementDecoded($fid);
+      if ($element === NULL) {
+        continue;
+      }
+      [, $c] = explode('_', $fid);
+      $contact_type = wf_crm_aval($this->settings['data']['contact'], "$c:contact:1:contact_type");
+      $allow_create = $this->utils->wf_crm_name_field_exists($enabled + $this->settings, $c, $contact_type);
+      if ((int) wf_crm_aval($element, '#allow_create', 0) !== (int) $allow_create) {
+        $element['#allow_create'] = $allow_create;
+        $this->webform->setElementProperties($fid, $element);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


peetCreative:
@eileenmcnaughton @colemanw @demeritcowboy 
Can you have a look at this PR? The CI is failing on StripeTest and ContributionPayLaterTest .

demeritcowboy:
I don't think either of the fails are "real", but re-running them now.

peetCreative:
@demeritcowboy OK cool now all tests are green.

demeritcowboy:
Hi @peetCreative I don't usually handle anything except the tests here but will see if I can review.

demeritcowboy:
I spent some time trying to see what's going on here:

* I can only reproduce a crash if there is something else on the civicrm tab that requires a contact, e.g. the activity section is enabled so that it creates an activity.
* With the patch, I still get the error even if I resave the civicrm tab and it updates allow_create, because allow_create doesn't seem to be the actual cause of the crash, it's that it's trying to create an activity (or whatever) and there is no contact created or existing contact. That isn't controlled by allow_create - it happens here regardless of allow_create: https://github.com/colemanw/webform_civicrm/blob/4a7ec023a579e896dd036840d1b360e311f80d41/src/WebformCivicrmPostProcess.php#L211-L215
* Even then, I'm not clear on what _should_ happen if you've configured a form to have an existing contact and also something like activity creation but not the right contact fields and no contact selected. Should it just skip doing everything and silently do nothing? That seems wrong too. It seems like it should warn that there is no logged in user or contact passed in/selected.
* There's nothing in here for upgrading existing sites with a bad config. It's not a blocker, and might be out-of-scope, especially if the problem is not really allow_create.
* Minor note: For the unit test, ideally it should be testing that submitting the form works, not just checking the value of allow_create. Except see earlier note: it's not really clear what the result should be.

peetCreative:
@demeritcowboy 
This was a bug I was running and fixing, before I could do the actual feature PR. So I split it.
So this is  mainly a previous step for: https://github.com/colemanw/webform_civicrm/pull/1123

PR #1124, opened 2026-06-04
Fix for missing class "CRM_Core_DAO" not found

shaneonabike:
Overview
While doing a drush update I kept getting

```
  [notice] Update started: webform_civicrm_update_8008
>  [error]  Class "CRM_Core_DAO" not found 
>  [error]  Update failed: webform_civicrm_update_8008
```

Technical Details

I don't fully understand, but my guess is that when loading the Webforms that might trigger some CiviCRM related items, and since it's not loaded it borks. Adding the line above makes the upgrade work properly

### DIFF
--- a/webform_civicrm.install
+++ b/webform_civicrm.install
@@ -467,6 +467,7 @@ function _recursiveKeyReplace(array &$elements, $searchKey, $replacementKey) {
  * Update usage of group to crmgroup
  */
 function webform_civicrm_update_8008() {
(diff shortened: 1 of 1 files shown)


shaneonabike:
@colemanw @KarinG I don't think that these are failures because of my code change but I could be wrong. I can't really tell

KarinG:
They are not but I've hit re-run on the D10.5/6.12 one and will re-run the tests again on Merge.

PR #1125, opened 2026-06-12
Add option to hide existing membership status message

francescbassas:
When using webform_civicrm with membership forms, the existing membership status message is always shown to contacts with an active/expired membership.

Some use cases require hiding this message (e.g. renewal forms where the status info is redundant or confusing for the user).

## Proposed changes
- Adds a checkbox "Show existing membership status message" in the Memberships tab of the CiviCRM webform settings.
- Defaults to enabled (TRUE) to preserve existing behavior for all current installations.
- Passes the option as a parameter to `loadMemberships()`, which conditionally calls `$this->setMessage()`.

## Backward compatibility
Fully backward compatible. Default value is TRUE (checkbox checked), so no existing configuration is affected.

### DIFF
--- a/src/AdminForm.php
+++ b/src/AdminForm.php
@@ -1052,7 +1052,7 @@ class AdminForm implements AdminFormInterface {
       '#type' => 'checkbox',
       '#title' => t('Show existing membership status message'),
       '#description' => t('When a contact with an existing membership fills out the form, display a message showing their current membership status and expiry date.'),
-      '#default_value' => (bool) wf_crm_aval($this->data, 'membership_options:show_status_message', 1, TRUE),
+      '#default_value' => (bool) wf_crm_aval($this->settings, 'membership_options:show_status_message', 1, TRUE),
       '#parents' => ['membership_options', 'show_status_message'],
     ];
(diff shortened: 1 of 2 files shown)


src/AdminForm.php
```diff
@@ -1048,6 +1048,13 @@ private function buildMembershipTab() {
       '#group' => 'webform_civicrm',
       '#attributes' => ['class' => ['civi_icon_fa-id-badge']],
     ];
+    $this->form['membership']['membership_show_status_message'] = [
+      '#type' => 'checkbox',
```
demeritcowboy, line 1052: I don't know if special handling is needed because it's a checkbox, but if I uncheck it and then save, it still stays checked when I reload, and then also lower down when it's actually used on the form `wf_crm_aval($this->data, 'membership_options:show_status_message', 1, TRUE);` returns "1" because it's not in `$this->data`.

It does seem to save it in the db properly as 1 or 0, it's just not getting into $this->data. I'm not sure how that gets populated or what is needed to make that happen.

demeritcowboy: I see what it is: It should be using `$this->settings` not `$this->data`. And then also down below.

francescbassas: Thanks @demeritcowboy changes made!

PR #1133, opened 2026-08-22
[test only] makework

demeritcowboy:
Overview
See https://www.drupal.org/node/3548485


demeritcowboy:
Given the intermittent infra issues and that this is just for tests and NFC since all it does is duplicate what drupal core already does in the base class, I'm going to merge.

PR #1137, opened 2026-09-21
Block unknown users: hide the form children, not the form root

magnolia61:
## Overview

Since #1098 a form with "Block unknown users" enabled renders as an empty page for anonymous visitors: no form, no message, no explanation. With verbose error reporting on, Drupal also logs:

```
Warning: Undefined array key "#markup" in Drupal\Core\Render\RenderCache->getCacheableRenderArray() (line 130 of core/lib/Drupal/Core/Render/RenderCache.php)
```

#1098 replaced the 403 with `$this->form['#access'] = FALSE`, set on the form root. `Renderer::doRender()` returns an empty string for an inaccessible element without adding a `#markup` key, and `getCacheableRenderArray()` reads that key unconditionally while `#attached` and `#cache` next to it use null coalescing.

This keeps the intent of #1098, no 403 for the page that embeds the form, and hides the form's children instead of the root. The status message and the admin-defined `prefix_unknown` text are visible again.

## Steps to reproduce

1. Create a webform with the CiviCRM handler, enable "Block unknown users"
2. Open the form as an anonymous visitor, without a checksum in the URL
3. The page shows the title only. With `error_level: verbose` the warning above appears in the log

## Before

Empty page plus the core warning.

## After

The visitor gets the "no permission" message and the unknown-contact text, no form fields, no warning.

## Implementation note

The children are walked rather than listed by name, so a webform footer, a wizard progress bar and anything a later Webform version adds are covered too. Only `form_build_id`, `form_id` and `form_token` are left in place, because Drupal's form processing expects them and they render as hidden inputs.

Elements added by form_alter implementations that run after this handler are not covered. On my site that is Honeypot, which leaves a hidden `honeypot_time` and its decoy field in the markup. Both are infrastructure rather than form input, so I left this as is, but please say if you would rather see this moved to an `#after_build` callback.

`setMessage()` is part of this change because the message that #1098 removed is restored here, and that method ignored its `$type` argument: the two existing callers that pass 'warning' (event full, event ended) rendered as a green success message.

## Testing

Drupal 11.4, Webform 6.x, webform_civicrm 6.3.3, patched against 6.x at 945 lines.

| check | result |
|---|---|
| forms with the setting that rendered an empty page before | 42 |
| same forms after the change, rendering a form plus message | 42 |
| visible input fields on a blocked form | 0 |
| webform footer on a blocked form | hidden |
| forged POST to a blocked form | no submission created |
| form with a valid checksum | renders in full, all wizard pages |

Not covered by my testing: the event-full and event-ended paths, since this site does not use event registration on these forms. The `submitted` key in the event-full branch does not exist in Webform 6.x either, so that branch hides only the actions, but I left it alone here and will report it separately.

Issue: https://www.drupal.org/project/webform_civicrm/issues/3485026


magnolia61:
For reference: that warning is a known core symptom, reported as [#3405539](https://www.drupal.org/project/drupal/issues/3405539) in the render system and closed as outdated. `getCacheableRenderArray()` reads `#markup` unconditionally while `#attached` and `#cache` next to it use null coalescing, so any render array that never got that key surfaces this way. Since core is not changing that, a module that denies access to a form root will keep producing it.

